PR #8, opened 2019-01-16
add windows compatibility

Atrox:
There is no `/tmp` on windows. Go will use the OS specific temporary directory if we pass no specific directory to `ioutil.TempDir`

### DIFF
--- a/actions/parser.go
+++ b/actions/parser.go
@@ -38,7 +38,7 @@ func ParseWorkflows(workingDir string, workflowPath string) (Workflows, error) {
 	}
 	workflows.WorkingDir = workingDir
 	workflows.WorkflowPath = workflowPath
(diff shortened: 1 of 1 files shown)


cplee:
@Atrox makes sense, thanks!  Problem is, this change will break macOS.  I had to explicitly add `/tmp` because on macOS my `TMPDIR` is something like `/var/folders/xh/7yly4s4d2psd5wfztxrcps6r0000gn/T/` which is not allowed in Docker for shared volumes.  Any opinions on how to work around that?

Atrox:
Ouh, did you fix the macOS related issue? This is new too me...

cplee:
your fix worked fine...i was mistaken.  thank you!

PR #11, opened 2019-01-17
Revert "Allow non-symbolic-ref HEAD"

Code0x58:
Sorry, I didn't mark it was a WIP in time - I didn't read properly and the change breaks (although won't be any more broken than existing code)

Reverts nektos/act#10


cplee:
@Code0x58 my bad...sorry!  i did see the same problem and fixed it as well as implemented a fix for your issue with 20f20eddf71aaad0bb852254d0d8bdcea94e2870

thanks!

PR #23, opened 2019-01-25
fix #22: require prefix './' for local actions

### LINKED ISSUE #22: workflow parser too permissive with local path references to actions in uses clauses
act version: 0.0.5
platform: OSX 10.14.2

When referencing actions that are contained in sub directories of the same repo. i.e.:

action "Lint" {
  **uses = "./.github/actions/go"**
  args = "lint"
}

the act parser allows you to write the uses clause without "./" at the start like this:

action "Lint" {
  **uses = ".github/actions/go"**
  args = "lint"
}

Although this is very correct and will run locally, the GitHub action parser won't be as permissive and will emit a parser error when you push your code

Behaviors should be aligned...

cplee:


michaKFromParis:
LGTM

PR #29, opened 2019-02-10
--watch

jnewland:
This PR adds support for a `--watch` or `-w` flag to `act`. When provided, `act` doesn't exit after the first run, but instead runs a goroutine that watches the local directory for changes (ignoring files in `.gitignore`). When changes are detected, the workflow file is parsed and run again.


jnewland:
Oh dang! Rookie mistake! Looks like I submitted this against a pretty old version of act. I'll close and revisit after getting back up to speed. Sorry about that!

PR #30, opened 2019-02-10
--watch

jnewland:
This PR adds support for a `--watch` or `-w` flag to `act`. When provided, `act` doesn't exit after the first run, but instead runs a goroutine that watches the local directory for changes (ignoring files in `.gitignore`). When changes are detected, the workflow file is parsed and run again.

A redux of #29, this time without conflicts.


cplee:
@jnewland this is awesome!  I'm very excited to add this...i'll review this week.

jnewland:
@cplee right on no rush at all! 👍I'm happy to iterate on this PR in any way too: don't be shy with your review. 😄

cmd/root.go
```diff
@@ -52,34 +56,91 @@ func newRunCommand(runnerConfig *actions.RunnerConfig) func(*cobra.Command, []st
...
+	)
+
+	folderWatcher.Start()
+	log.Debugf("Watching %s for changes", dir)
+
+	go func() {
+		for folderWatcher.IsRunning() {
+			for changes := range folderWatcher.ChangeDetails() {
+				log.Debugf("%s", changes.String())
+				fn()
```
cplee: the `fn` parameter for this `watchAndRun` function should be `fn func() error` and then use the return of that function to be the return for `watchAndRun` to avoid swallowing the error.

cmd/root.go
```diff
@@ -52,34 +56,91 @@ func newRunCommand(runnerConfig *actions.RunnerConfig) func(*cobra.Command, []st
 			runnerConfig.EventName = args[0]
 		}
 
-		// create the runner
-		runner, err := actions.NewRunner(runnerConfig)
+		err := parseAndRun(cmd, runnerConfig)
```
cplee: how about moving this below as an `else { }` after the `if watch {}`?

PR #45, opened 2019-02-27
Scaffold a new action with --init (-i) flag

sosedoff:
I think `act` could be improved with a feature to help setting up a new workflow.

For any existing project running `act --init [push|pull-request|other event type]` will create a new workflow file and an example action template using dummy Dockerfile. It's not much but definitely enough to get rolling with github actions without all the manual work.

New command will only execute if user does not have a `.github` directory (or whatever is set with workspace dir). In order to get started with a new action using docker, one has to type 2 commands:

```
cd ~/projects/foobar
act --init
act
```

And then take it from there to develop and test whatever action they need.

`act --init` creates the following files in the current directory:

```
./.github
./.github/main.workflow
./.github/example-action
./.github/example-action/Dockerfile
./.github/example-action/entrypoint.sh
```

@nektos any thoughts?

### DIFF
--- a/actions/api.go
+++ b/actions/api.go
@@ -44,6 +44,8 @@ type RunnerConfig struct {
 	EventPath       string          // path to JSON file to use for event.json in containers, relative to WorkingDir
 	ReuseContainers bool            // reuse containers to maintain state
 	ForcePull       bool            // force pulling of the image, if already present
+	Init            bool            // init a new action
+	InitRepo        string          // git repo to pull template from
 }
 
 type environmentApplier interface {
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -3,6 +3,7 @@ package cmd
 import (
 	"context"
 	"fmt"
+	"io/ioutil"
 	"os"
 	"path/filepath"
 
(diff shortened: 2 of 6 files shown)


cplee:
@sosedoff i love the idea!  I have some thoughts that i'd like to discuss.  I don't like that all the templating is hard coded.  How about this idea...what if you could pass a URL of a git repo (even just a github slug `sosedoff/example-action`) and it would pull down the template files from that repo and apply them.  

This would allow you to have all sorts of sample workflows that could be applied.  Thoughts?

sosedoff:
Yeah, the hardcoded template is just a POC to get started with, but i would prefer not to do any upfront work to begin with, like creating a template repository.

Maybe we can setup a public repo (or an org) with templates for act so users don't have to create them and when `act --init [push]` runs it'll use the `github.com/act-templates/push` as a source. If you prefer to use a git repo for templates - i'd suggest we keep them in this repo for now until there's a real need to extract into a separate one. 

Overall, i see `act --init` to be similar to `rails new myapp` generator which runs the default scenario and offers an option to specify the template. Basically, i just want a straightforward init feature that does not require me to think about how to use it, that's all.

cplee:
I'm good with using the existing repo for the templates, but i would definitely prefer keeping the templates out of string variables in the code.  Love the feature and think it is valuable

PR #84, opened 2020-02-18
Support GA of GitHub Actions

### LINKED ISSUE #74: Support Actions v2
Just to keep track of the work here :)

Thanks for you awesome tool !

cplee:
Closes #74 

support the new v2 of GitHub Actions (e.g. yaml files in `.github/workflows/`).  This has been a much larger effort than i first thought 😰

Here's a summary of features:

- [x] Ability to read workflow files from `.github/workflows`
- [x] Ability to run commands via `run:` syntax
- [x] Ability to run commands in container via `jobs.<id>.container`
- [x] Ability to `use` docker urls
- [x] Ability to `use` github repo actions of type `node12`
- [x] Ability to `use` github repo actions of type `docker`
- [x] Ability to `use` local actions (e.g. `use: ./.github/...`)
- [x] Support for logging commands (e.g. `::set-outputs::..`)
- [x] Support for `matrix` strategies
- [x] Expressions (e.g. `run: echo ${{secret.foo}}`


marcosnils:
Hey @cplee. Congrats on landing this PR. I've been trying to find the time to work in a local way to run actions for a specific repo since I'm a big fan of `act` which I used in actions v1. 

Seems like you did an amazing work and effort to implements all this features. The way what I thought about implementing this local runner is to actually use the open source runner (https://github.com/actions/runner) that already has all the logic on how to run and process the workflow yaml files. 
I'd love to know if you thought about that approach and if so, the main reasons why you chose to implement it yourself. If you believe we can use the open source runner instead, I already did some work around that that I'd like to share.

cplee:
@marcosnils I did look at it as my first approach.   Ran it locally and captured the API traffic back to GitHub from the runner.   The API is very chatty, verbose....and undocumented. Lots of references to VSTS in the API payloads 🤯

Looked much more difficult to simulate the API between the runner and GH so I went the direction I did.  Would love to hear more about your learnings and I'm happy to reconsider!

marcosnils:
I haven't checked in detail, but I only saw that the listener can be invoked in something that's called `local run` (https://github.com/actions/runner/blob/6c70d53eead402ba5d53676d6ed649a04e219c9b/src/Misc/layoutroot/run.sh#L23-L27). I was hoping that option could actually help running stuff locally without having to interact with the VSTS or GH API. Seems like you have better understanding about the runner internals and it might be a dead end?

cplee:
Looks like it has potential...that's the right idea.  Can't find any details on how that works though... 🤔

marcosnils:
Maybe @TingluoHuang can help us here?

cplee:
Moving conversation above over to issue #88

PR #135, opened 2020-03-06
Bump golang to v1.14

chenrui333:
go 1.14 release notes, https://golang.org/doc/go1.14


chenrui333:
You are lightening fast! :)

PR #142, opened 2020-03-10
fix #141 - use host network to match GitHub Actions runners

### LINKED ISSUE #141: [Question] is act container with hostNetwork supported?
Hi, I'm trying to use an [action](https://github.com/engineerd/setup-kind) that creates a [kind](https://kind.sigs.k8s.io/) cluster, which works just fine but this action opens a port in 127.0.0.1, which is not possible to reach from any step. Is it possible to use hostNetwork or what is the suggestion for this scenario?
Thanks

cplee:


aledbf:
@cplee thank you for implementing this feature.

PR #155, opened 2020-03-14
Add support for runs-on array form (closes #146)

### LINKED ISSUE #146: Array of runs-on runners fails
Using `act` 0.2.6. An array of runners (even if only one element) fails like so:

```
Error: yaml: unmarshal errors:
  line 9: cannot unmarshal !!seq into string
  line 70: cannot unmarshal !!seq into string
  line 100: cannot unmarshal !!seq into string
  line 184: cannot unmarshal !!seq into string
```

Submitting a PR to accept an array and use only the first runner would be trivial, but probably not super helpful. A PR to actually run all `runs-on` runners would probably be more substantial and worth discussion first. What do you think?

aidansteele:
Implemented as per discussion in #146. I guess the "interesting" thing here is the tests. It was easy enough to see that the parser no longer crashes on the array form. I wasn't able to think of a test for the "no matching labels" case - right now it just logs a message and skips over it silently. This was the existing behaviour, so I guess that's not a problem?

### DIFF
--- a/pkg/model/workflow.go
+++ b/pkg/model/workflow.go
@@ -56,7 +56,7 @@ func (w *Workflow) On() []string {
 type Job struct {
 	Name           string                    `yaml:"name"`
 	RawNeeds       yaml.Node                 `yaml:"needs"`
-	RunsOn         string                    `yaml:"runs-on"`
+	RawRunsOn      yaml.Node                 `yaml:"runs-on"`
 	Env            map[string]string         `yaml:"env"`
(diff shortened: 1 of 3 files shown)


aidansteele:
It also occurred to me that since this was a literal copy and paste of the `Needs()` method, it might be desirable to factor the code out into a `StringSliceish` type that implements `yaml.Unmarshaler`. I get the feeling this is going to come up again

PR #164, opened 2020-03-18
Use go-git for looking up repository details

rsotnychenko:
**What was changed?**
This PR replaces all custom logic used to fetch Git metadata with `go-git` library.

**Why do we need the changes?**
While working on a project I noticed `act` is setting an incorrect repository name & reference in the event.
This is happening due to the repositories layout I'm using: a single repo containing a number of submodules, each of those define their own workflows. Since git submodules reside in `.git` folder of the parent repo, `act`'s logic of parsing `.git/config` wasn't working as it should.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -12,7 +12,6 @@ require (
 	github.com/docker/docker v0.0.0-20200229013735-71373c6105e3
 	github.com/docker/go-connections v0.4.0 // indirect
 	github.com/emirpasic/gods v1.12.0 // indirect
-	github.com/go-ini/ini v1.41.0
 	github.com/gopherjs/gopherjs v0.0.0-20181103185306-d547d1d9531e // indirect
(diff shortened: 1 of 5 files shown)


cplee:
@rsotnychenko - like the idea, there are some conflicts though due to upgrade of go-git...you able to take a peek?

jsoref:
Some git parsers (I can't remember which? go-git? rust-git?) scream about unexpected metadata, esp. [hg-git](https://hg-git.github.io/)'s. https://github.com/elastic/elasticsearch/pull/37035#issuecomment-450674993

It would suck if suddenly a majority of the things I touched exploded (OTOH, if someone fixed this, I'd be overjoyed).

Ichhabejk:
# ``__

Ichhabejk:
@

Issue #145, opened 2020-03-11
Seemingly invalid path separator on Windows using actions/setup-node


Swiftwork:
Running command
```powershell
act -P ubuntu-latest=nektos/act-environments-ubuntu:18.04
```
on `Microsoft Windows 10 Pro Build 10.0.18363` using `release v0.2.6`
```yaml
- name: Set up Node.js version 13.x
  uses: actions/setup-node@v1
  with:
    node-version: 13.x
```
results in

```powershell
[Continuous Delivery/Node 13.x]   �🐳  docker cp src=act/actions-setup-node@v1 dst=/actions\
| internal/modules/cjs/loader.js:985
|   throw err;
|   ^
|
| Error: Cannot find module '/github/workspace/\actions\actions-setup-node@v1\dist\index.js'
|     at Function.Module._resolveFilename (internal/modules/cjs/loader.js:982:15)
|     at Function.Module._load (internal/modules/cjs/loader.js:864:27)
|     at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:74:12)
|     at internal/main/run_main_module.js:18:47 {
|   code: 'MODULE_NOT_FOUND',
|   requireStack: []
| }
[Continuous Delivery/Node 13.x]   ❌  Failure - Set up Node.js version 13.x
Error: exit with `FAILURE`: 1
```
This seems to be an issue with Windows and POSIX not using the same path separators. However I don't know where this issue lies, if its with [nektos/act](https://github.com/nektos/act) or [actions/setup-node](https://github.com/actions/setup-node). If you can't reproduce the issue it occurs currently with [trutoo/event-bus](https://github.com/trutoo/event-bus). It does however work in in my WSL.

cplee:
This looks like a bug with `act`.  In an effort to support windows, i overdid the use of `filepath.Join()` and `filepath.Separator`.  These create paths that are windows specific and are then fed in to the linux container 🤦‍♂️

All the places that use those need to be reviewed to determine if the path is used on the host or within the container.  The ones in the container should all use `/` explicitly.

aslafy-z:
This behavior has to be specific to Linux containers. In windows containers, path separator has to match the windows path separator (`\`). I've seen that windows runners aren't supported in act from now so it may not be a priority but it guess it can be implemented in a adaptable way for the future.

ArwynFr:
Workspace volume mount is impacted as well.
Instead of creating a hierarchic directory, all files are flattened in the same directory :

```
[...]

PR #180, opened 2020-04-01
Windows+Docker path reference fixes, mostly around the not relying on filepath.Separator

syntaqx:
Updates path concatenation/usage to leverage `filepath.Separator` (often through `filepath.Join`) when referencing the host OS, but explicitly using `/` (often through `path.Join`) when referencing paths in docker containers/mounts.

__Update:__ In addition to the above, there is also a healthy amount of path normalization so that Windows paths can be compared to their Docker counterparts. This should still be cross-os friendly as well.

In theory, this should work bi-directional including when using Windows containers, as Windows is very lax with this, but I don't really have the time to test it.

Addresses #145 and possibly a chain of others related to it.

> Additionally adds a `go-install` to make as `/usr/local/bin` not available, or accessible in Git Bash. There's plenty of other ways of doing this, but this met my needs.

Please let me know if I need to update anything to meet contribution guidelines I didn't follow, or changes you'd like before merging.

### DIFF
--- a/pkg/container/docker_run.go
+++ b/pkg/container/docker_run.go
@@ -5,14 +5,15 @@ import (
 	"bytes"
 	"context"
 	"fmt"
-	"gopkg.in/src-d/go-billy.v4/helper/polyfill"
-	"gopkg.in/src-d/go-billy.v4/osfs"
-	"gopkg.in/src-d/go-git.v4/plumbing/format/gitignore"
 	"io"
 	"io/ioutil"
 	"os"
 	"path/filepath"
 	"strings"
+
+	"gopkg.in/src-d/go-billy.v4/helper/polyfill"
+	"gopkg.in/src-d/go-billy.v4/osfs"
+	"gopkg.in/src-d/go-git.v4/plumbing/format/gitignore"
 
 	"github.com/docker/docker/api/types"
 	"github.com/docker/docker/api/types/container"
--- a/pkg/container/docker_run.go
+++ b/pkg/container/docker_run.go
@@ -322,10 +322,8 @@ func (cr *containerReference) copyDir(dstPath string, srcPath string) common.Exe
 		defer os.Remove(tarFile.Name())
 		tw := tar.NewWriter(tarFile)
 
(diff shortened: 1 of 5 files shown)


syntaqx:
So, after further digging into issue, this doesn't seem to be functional yet. I don't seem to be able to get the `node` command to run correctly, as it's having issues removing the `act` TempDir prefix and matching files to the mounted volume. Would be curious if anyone has thoughts.

syntaqx:
After just about pulling my hair out, it now runs! Hopefully it'll work for everyone else too 🤣

lonelam:
I also encountered this problem, your solution is really inspiring, but I think it failed on test action. Trying to figure it out.

cplee:
Very interested in how this progresses 👀

pkg/runner/step_context.go
```diff
@@ -245,14 +246,16 @@ func (sc *StepContext) runAction(actionDir string, actionPath string) common.Exe
 
 		actionName := ""
 		containerActionDir := "."
+
 		if step.Type() == model.StepTypeUsesActionLocal {
```
haithaison-cpu, line 250: t

pkg/container/docker_run.go
```diff
@@ -5,15 +5,16 @@ import (
...
-	"gopkg.in/src-d/go-git.v4/plumbing/format/gitignore"
 	"io"
 	"io/ioutil"
 	"os"
 	"path/filepath"
 	"strings"
 
+	"gopkg.in/src-d/go-billy.v4/helper/polyfill"
+	"gopkg.in/src-d/go-billy.v4/osfs"
+	"gopkg.in/src-d/go-git.v4/plumbing/format/gitignore"
```
xmesaj2, line 16: these are at v5 as of f6e37a8d67feb8c3b0fa0f0263f2e4732b1206a0

syntaqx:
Unfortunately I don't have the availability to finish this PR, closing :'(

tyler274:
What still needed to be done here?

Issue #179, opened 2020-04-01
how to trigger a tag event locally 


jonyhy96:
act version 0.2.7

workflow:
```yml
name: postgres action 

on:
  push:
    branches:
      - master
    tags:
      - v*

jobs:
  build-and-push:
    runs-on: ubuntu-16.04
    timeout-minutes: 3
    steps:
    - name: Checkout
      uses: actions/checkout@v2

    - name: build and push image
      uses: docker/build-push-action@v1
      with:
        username: ${{ secrets.DOCKER_USERNAME }}
        password: ${{ secrets.DOCKER_PASSWORD }}
        repository: byzanteam/nitrogen-postgres
        registry: ${{ secrets.DOCKER_REGISTRY }}
        tag_with_ref: true
```

tag.json
```json
{
    "ref": "refs/tags/v0.0.1"
}

```

log:
```
DEBU[0000] Loading workflows from '/PATH/.github/workflows' 
DEBU[0000] Reading workflow '/PATH/.github/workflows/main.yml' 
DEBU[0000] Using detected workflow event: push          
DEBU[0000] Planning event: push                         
DEBU[0000] Reading event.json from /PATH/tag.json 
DEBU[0000] Loading slug from git directory '/PATH/.git' 
DEBU[0000] Found revision: 7c6d9e0bd6b4fcd2dfe32296e8298ae4c32ac2d3 
DEBU[0000] Loading revision from git directory '/PATH/.git' 
DEBU[0000] Found revision: 7c6d9e0bd6b4fcd2dfe32296e8298ae4c32ac2d3 
DEBU[0000] HEAD points to '7c6d9e0bd6b4fcd2dfe32296e8298ae4c32ac2d3' 
DEBU[0000] HEAD matches refs/heads/master               
DEBU[0000] using github ref: refs/heads/master          
[postgres action/build-and-push] 🚀  Start image=node:12.6-stretch-slim
DEBU[0000] Loading slug from git directory '/PATH/.git' 
DEBU[0000] Found revision: 7c6d9e0bd6b4fcd2dfe32296e8298ae4c32ac2d3 
DEBU[0000] Loading revision from git directory '/PATH/.git' 
DEBU[0000] Found revision: 7c6d9e0bd6b4fcd2dfe32296e8298ae4c32ac2d3 
DEBU[0000] HEAD points to '7c6d9e0bd6b4fcd2dfe32296e8298ae4c32ac2d3' 
DEBU[0000] HEAD matches refs/heads/master               
[...]

Issue #124, opened 2020-03-05
Support dotenv


muuvmuuv:
I'm storing all my secrets passed to GitHub in a dotenv and it would be sooo helpful if I could do `act -f .env pull_request`.

cplee:
Closing to keep discussion going on #123 

@muuvmuuv - can you review the suggestion there and see if that would help or if `.env` is the better way to go

cplee:
implemented in #133

muuvmuuv:
Oh, that was fast! Sorry for the delay and thank you! :)

muuvmuuv:
Doesn't seem to work.

The `.env.build` is in the root.

```
❯ act -b -r -v --env-file .env.build fork
DEBU[0000] Loading environment from /Users/marvinheilemann/Development/Portfolio/.env.build
DEBU[0000] Loading workflows from '/Users/marvinheilemann/Development/Portfolio/.github/workflows'
DEBU[0000] Reading workflow '/Users/marvinheilemann/Development/Portfolio/.github/workflows/build.yml'
DEBU[0000] Reading workflow '/Users/marvinheilemann/Development/Portfolio/.github/workflows/fork.yml'
DEBU[0000] Reading workflow '/Users/marvinheilemann/Development/Portfolio/.github/workflows/lint.yml'
DEBU[0000] Reading workflow '/Users/marvinheilemann/Development/Portfolio/.github/workflows/notify.yml'
DEBU[0000] Reading workflow '/Users/marvinheilemann/Development/Portfolio/.github/workflows/stale.yml'
DEBU[0000] Planning event: fork
DEBU[0000] Loading slug from git directory '/Users/marvinheilemann/Development/Portfolio/.git'
DEBU[0000] Found revision: 078042c0e1c7955e49f19816f18a0a0c939bac50
DEBU[0000] Loading revision from git directory '/Users/marvinheilemann/Development/Portfolio/.git'
DEBU[0000] Found revision: 078042c0e1c7955e49f19816f18a0a0c939bac50
DEBU[0000] HEAD points to '078042c0e1c7955e49f19816f18a0a0c939bac50'
DEBU[0000] HEAD matches refs/heads/development
DEBU[0000] using github ref: refs/heads/development
[Fork created/Notify by email] 🚀  Start image=node:12.6-buster-slim
[Fork created/Notify by email]   🐳  docker pull node:12.6-buster-slim
DEBU[0000] Image exists? true
[...]

PR #184, opened 2020-04-03
feat: load every environment from --env-file to workflow

jonyhy96:
As mentioned in #124 and #179, users are confused about the flag `--env-file` behavior.

So i think it is better to make it easier for user to use the .env file to set up some environments they want to use in the further workflow.


cplee:
@jonyhy96 thanks for this PR.  i have some hesitation about assuming ALL environment variables should be accessible as GitHub secrets...seems like an assumption that could get folks in trouble, so i avoided that  in #124 

The idea was only those environment variables that you explicitly name (with `-s`) can be passed along.

Now this PR is trying to address a different problem from #179, about being able to override the environment variables that are used within the containers, specifically the GITHUB environment variables.

Couple thoughts...first, should this be a different flag to avoid overloading the current `--env-file` flag since these env vars are used differently...or if you like the name, add a new `--secrets-file` and change the current behavior to read secrets from that file?

Also, i don't really like how the path to env-file is passed through a new env variable and reread.  I'd rather those vars are read in from the CLI and passed in the runner context.

jonyhy96:
Thanks for reply. 

> a new `--secrets-file` and change the current behavior to read secrets from that file

I prefer this way which is more clear to the user to use with `-s` flag.However, i think we could remove the `-s` flag and keep only `--secrets-file` flag which can be more easier for user to use secrets.(without supply two flags)

> Also, i don't really like how the path to env-file is passed through a new env variable and reread. I'd rather those vars are read in from the CLI and passed in the runner context.

And i totally agree with this, i use a new environment rather than context here because of i'm not sure if this pull request is my personally misunderstanding of the behavior. i will try to figure out a way to pass it through by a context.

jonyhy96:
/cc @cplee  
Please review and let me know if anything wrong with the way i use the context.Value.
Thanks!

cplee:
sorry @jonyhy96 - i misspoke.  I meant the run config, not context.  my bad!

I have a commit with what i had in mind and also has the `--secret-file` implemented.  I'd like to keep the existing `-s` as many folks like to pull secrets from their existing env (e.g. AWS_ACCESS_TOKEN) or pass a one off secret.

https://github.com/jonyhy96/act/pull/1

Thanks again!

jonyhy96:
> [jonyhy96#1](https://github.com/jonyhy96/act/pull/1)

This looks good to me.I notice that [mergeMaps](https://github.com/jonyhy96/act/pull/1/files#diff-2c17edb7e58fb6d5704993cc8f5dc583L48) may overwrite environments in env-file by some built-in environments.Maybe we should mention that.
By the way, should i merge the pr to my fork repo?

cplee:
@jonyhy96 yes, please merge to your branch

jonyhy96:
> @jonyhy96 yes, please merge to your branch

Done

PR #223, opened 2020-05-04
Implement UnmarshalYAML for ActionRunsUsing

mheap:
In #222 I added case normalisation to ReadAction() to ensure that Docker and docker are interpreted the same way.

I realised that this was being done at the wrong level and required multiple type conversions. By implementing `func (a ActionRunsUsing) UnmarshalYAML` we can lowercase the string as it's being unserialized

This has an added benefit that any time this type is hydrated the `runs.using` key will be lowercased, rather than relying on `ReadAction()`

### DIFF
--- a/pkg/model/action.go
+++ b/pkg/model/action.go
@@ -1,6 +1,7 @@
 package model
 
 import (
+	"fmt"
 	"io"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


ethan797979:
Please help me to unsubscribe it.




於 星期一, 5月 4, 2020, 20:40，Casey Lee <notifications@github.com> 寫道：




Merged #223 into master.

ethan797979:
Please help me that if anything was by me to Make anything was subscribed,
All of everything ,I want they become unsubscribe .
Thank you


於 星期一, 5月 4, 2020, 20:40，Casey Lee <notifications@github.com> 寫道：




Merged #223 into master.

Issue #186, opened 2020-04-04
Folder is persisted across runs


Sygmei:
If I create a simple folder inside the cloned repository with the `mkdir some_folder` command and I try to re-run the script, it will fail with the following error :

`mkdir: cannot create directory 'some_folder': File exists`

Is this the expected behaviour ? Is there a way to clean the "cache" ?

chrislow:
Got the same problem. The should be flag for clearing the cache / rebuild the image on very run?

cplee:
there is a volume created for each job.  it _should_ get cleared after each job, but seems there may be situations in which the volume isn't being removed

PR #283, opened 2020-06-18
Container volumes sometimes reused even if configured not to

N2D4:
This can occur even without the `-r` flag, but it's easier to reproduce like this.
`.github/workflows/workflow.yaml`:
```yaml
name: Example workflow

on: push

jobs:
  example:
    name: Example job
    runs-on: ubuntu-latest

    steps:
      - run: echo "Hello" >> file.txt
      - run: echo "World!" >> file.txt
      - run: cat file.txt
```
Now we run:
```sh
$ act -j example -r
...
| Hello
| World!
...
```
This is expected.
```sh
$ act -j example -r
...
| Hello
| World!
| Hello
| World!
...
```
This is also expected, as the `-r` flag means that the containers should be re-used; in other words, file.txt contains the original content from the last run, plus the new content from this run.
```sh
$ act -j example
...
| Hello
| World!
| Hello
| World!
| Hello
| World!
...
```
This is unexpected; the `-r` flag is no longer used, but the file system wasn't reset.
```sh
$ act -j example
...
| Hello
| World!
...
```
This is expected; after the run without `-r`, the volume was deleted from disk and this time we have a clean file system.

The same thing happens even when not using `-r` if the container clean-up doesn't run (eg. when `act` is terminated in the middle of a build); the file system will persist. In the code, there's a safeguard which performs an additional clean-up before every build, but this safeguard only removes the container itself, not the volume associated with it (which is not anonymous). This PR fixes that by calling `runContext.stopJobContainer(...)` instead of `container.Remove()` before the build, which removes both the container and the volume.

### DIFF
--- a/pkg/container/docker_volume.go
+++ b/pkg/container/docker_volume.go
@@ -3,11 +3,34 @@ package container
 import (
 	"context"
 
+	"github.com/docker/docker/api/types/filters"
 	"github.com/nektos/act/pkg/common"
 )
 
-// NewDockerVolumeRemoveExecutor function
 func NewDockerVolumeRemoveExecutor(volume string, force bool) common.Executor {
+	return func(ctx context.Context) error {
+		cli, err := GetDockerClient(ctx)
+		if err != nil {
+			return err
+		}
+
+		list, err := cli.VolumeList(ctx, filters.NewArgs())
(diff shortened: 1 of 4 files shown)


cplee:
fix #186

PR #309, opened 2020-07-18
use t.Cleanup and prefer assert.NoError over .Nil

trashhalo:
testify has a method [NoError](https://godoc.org/github.com/stretchr/testify/assert#NoError) that is specifically for asserting that the err you have is nil. If failed it emits a message about the unexpected error. https://github.com/stretchr/testify/blob/master/assert/assertions.go#L1332

go 1.14 introduced [t.Cleanup](https://golang.org/pkg/testing/#B.Cleanup) its useful when you need to do something at the end of a test. Like cleaning up files.

### DIFF
--- a/pkg/common/git_test.go
+++ b/pkg/common/git_test.go
@@ -35,39 +35,39 @@ func TestFindGitSlug(t *testing.T) {
 	for _, tt := range slugTests {
 		provider, slug, err := findGitSlug(tt.url)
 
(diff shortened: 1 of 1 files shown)


trashhalo:
Im not sure whats up with the test failure. It seems to be in a package I didnt touch. Another pr is failing in the same way and didnt change any go code. https://github.com/nektos/act/pull/308

Issue #279, opened 2020-06-14
Allow loading secrets from a .env file


patoroco:
Hi,

currently `act` supports passing the secrets using the shell in two ways:

```bash
act -s SECRET=value

or

act -s SECRET
```
In the second option, if the `SECRET` environment variable exists, it will take that value, if not, it will prompt the user each time.

It would be really nice load these values directly from a `.env` file, what do you suggest?

MAHDTech:
You can try this, works for both secrets and standard environment variables.

* Create a file to store the environment variable secrets

```bash
vim ~/.config/act.vault

# Example secret
MY_SECRET=12345
```

* Then run act and load the files with the variables
```bash
act \
    --env-file "${HOME}/.config/act.env" \
    --secret-file "${HOME}/.config/act.vault"
```

👍

cplee:
thanks @MAHDTech!

@patoroco - can we close this?

patoroco:
Oh, actually the @MAHDTech explanation is not a workaround, it's just what I was looking for! Than @MAHDTech!

BTW, the only reference that I've found related to that is this file: https://github.com/nektos/act/blob/2f395475b0ac4f5c0fb3d7c8b97a23d9214ea77b/pkg/runner/testdata/secrets/.actrc#L2

Maybe we could add a small mention to it in the README, what do you think?

patoroco:
> thanks @MAHDTech!
> 
> 
> 
> @patoroco - can we close this?

Done! Thanks

PR #314, opened 2020-07-21
Document `secret-file` in README and add example

Strajk:
Hey 👋 thx for amazing work!

I needed to load secrets from a file and found the solution in issues, would be nice to add it to README :)

https://github.com/nektos/act/issues/279


Strajk:
I'm not very familiar with go ecosystem so I'm not sure why tests are failing :confused-parrot:

PR #327, opened 2020-08-01
add privileged mode

whywaita:
Subject Says It All!


whywaita:
@cplee Thank you for your merge commit! 

What's do you think about this change?

whywaita:
Don't worry! :)
Thank you!

PR #347, opened 2020-08-26
feat: docker 19.03.12

greut:
bump docker/docker to its latest release.


greut:
@cplee right windows build are a bit more tricky, I'll try again.

Issue #359, opened 2020-09-07
Copying directories from Windows to a Linux container flattens path.


barweiss:
When copying a directory to the container from a Windows host, the directory structure is flattened.

For example, consider the following directory structure:
```
|   file.txt
\---dir
        file.txt
```

Copying that to a Linux container will produce the following directory structure:

```
'dir\file.txt'
file.txt
```

Poselsky:
Solution wasn't uploaded to the realease yet and had the same issue when running commands with cmd or git bash (git bash makes even some weird artifact to the console).

I found the solution by using Windows 10 (v 2004) **wsl** with ubuntu subsystem and running github action from **wsl**.

PR #360, opened 2020-09-07
Fix copyDir to normalize file paths in tar archive

barweiss:
#359


cplee:
@barweiss - seems like a simple change. can you help confirm the intention of this change? Are we looking to replace `\` with `/` on windows hosts?

barweiss:
> @barweiss - seems like a simple change. can you help confirm the intention of this change? Are we looking to replace `\` with `/` on windows hosts?

Exactly, the files in the TAR archive end up having a backward slash in their names.

cplee (approved):

PR #367, opened 2020-09-15
Add macOS install instructions

yallen011:
On a mac, the Docker Daemon installed, you have to run a few extra commands to get docker to work with act.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,22 @@ Alternatively, you can use the following:
 
 `curl https://raw.githubusercontent.com/nektos/act/master/install.sh | sudo bash`
 
+If you are running on a Mac OS, run the following commands
+`brew install docker docker-machine`
+
+`brew cask install virtualbox`
+
+Accept the software installation from `Oracle America Inc.`
+
+Return to the console and run the following:
+
+`docker-machine create --driver virtualbox default`
+
+Run this command to configure your shell:
+
+`eval "$(docker-machine env default)"` 
+
+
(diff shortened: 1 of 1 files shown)


cplee:
@yallen011 - thanks for the contribution! I don't believe `virtualbox` or `docker-machine` are required to use `act`...can you help me understand why those are needed?

yallen011:
Yes, after following the instruction in the README for my mac, I got an error when trying to run the `act -n` command. Saying something like the docker daemon needed to be started. After looking it up, I came across the following articles that helped me get things up and running. https://dev.to/icanhazstring/local-testing-for-github-actions-on-macos-4lob and https://apple.stackexchange.com/questions/373888/how-do-i-start-the-docker-daemon-on-macos . After following these instructions, I was about to run the `act -n`command successfully. Sorry I don't have the exact error anymore. I can try to find it

seanknox:
@yallen011 I think is due to some confusion with docker packages in homebrew. I think most people use the cask to install Docker Desktop. You don't need to install anything else if you're doing `brew cask install docker` (e.g. I don't use virtualbox at all and act runs fine).

rodo36 (approved):

cplee:
@yallen011 - what are your thoughts given @seanknox comments: https://github.com/nektos/act/pull/367#issuecomment-714178889

Thoughts on providing the simpler installation of `brew cask install docker` instead?

yallen011:
Ok, I that will work. Just to be clear I understand correctly.  The change should be:
Install docker desktop if not already installed
`brew cask install docker`

cplee:
@yallen011 - correct

PR #371, opened 2020-09-25
Fix path for local node12 actions

### LINKED ISSUE #185: local js action fails (uses: ./)
When developing a new JS action, I like to create a workflow within the same repo so I can test various use cases.  In such a case, I reference the local action with `- uses: ./` as defined in [the documentation](https://help.github.com/en/actions/building-actions/creating-a-javascript-action#example-using-a-private-action)

```yaml
on: [push]
jobs:
  hello_world_job:
    runs-on: ubuntu-latest
    name: A job to say hello
    steps:
      - name: Checkout
        uses: actions/checkout@v2
      - uses: ./
        with:
          who-to-greet: 'Mona the Octocat'
``` 

Unfortunately, it fails with the following error: 

```
internal/modules/cjs/loader.js:628
|     throw err;
|     ^
| 
| Error: Cannot find module '/github/workspace/aws-instance-ripper/index.js'
|     at Function.Module._resolveFilename (internal/modules/cjs/loader.js:625:15)
|     at Function.Module._load (internal/modules/cjs/loader.js:527:27)
|     at Function.Module.runMain (internal/modules/cjs/loader.js:839:10)
|     at internal/main/run_main_module.js:17:11 {
|   code: 'MODULE_NOT_FOUND',
|   requireStack: []
| }
``` 

It does work if I move the action code in an `action` subfolder and use it with `- uses: ./action`.

This can be reproduced with [the sample hello-world-javascript-action](https://github.com/actions/hello-world-javascript-action.git), you just need to add a basic workflow.

jromero:
This changes makes it so that the proper path is used when executing a local javascript (`node12`) action.

Additional minor changes:
- Updated Makefile
    - Updated `make test` to be functional by running expected `go test` task based on [README.md](https://github.com/nektos/act#building-from-source)
    - Added `make format` to format based on [CONTRIBUTING.md](https://github.com/nektos/act/blob/master/CONTRIBUTING.md#-styleguide)
    - Made `GITHUB_TOKEN` be set only once instead of re-executing `cat` per target.
- Update `expression_tests` to retain previous behaviour (when no file matched) now that a `package-lock.json` file exists in `testdata`.

Fixes #185

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,7 +11,7 @@
 # Output of the go coverage tool, specifically when used with LiteIDE
 *.out
 
-dist/
+/dist/
 .todo
 
 *.nupkg
--- a/Makefile
+++ b/Makefile
@@ -7,15 +7,19 @@ PATCH_VERSION = $(word 3, $(subst ., ,$(word 1,$(subst -, , $(VERSION)))))
 NEW_VERSION ?= $(MAJOR_VERSION).$(MINOR_VERSION).$(shell echo $$(( $(PATCH_VERSION) + 1)) )
 
 ACT ?= go run main.go
-export GITHUB_TOKEN = $(shell cat ~/.config/github/token)
+export GITHUB_TOKEN := $(shell cat ~/.config/github/token)
 
 .PHONY: build
 build:
 	go build -ldflags "-X main.version=$(VERSION)" -o dist/local/act main.go
 
+.PHONY: format
+format:
+	go fmt ./...
+
 .PHONY: test
 test:
-	$(ACT) -W examples/workflows --verbose
+	go test ./...
 
 .PHONY: install
 install: build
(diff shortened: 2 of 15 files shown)


jromero:
The changes here are minimal but the PR looks huge because of the node_modules. I can try making the testdata action a compiled version via ncc if that's desired.

cplee:
@jromero - i'd rather not have the `node_modules` in the repo. can we come up with another option for this?

jromero:
> @jromero - i'd rather not have the `node_modules` in the repo. can we come up with another option for this?

Yup I can compile it as mentioned above. Given that this is more or less a static javascript app the concern of having to update it via compilation seems very minimal.

I'll update the PR and squash it.

jromero:
@cplee updated. Let me know what you think.

cplee (approved):
awesome @jromero 🎉

Issue #32, opened 2019-02-13
Error: unsupported event: push


sosedoff:
I'm playing around with an action that runs when a PR is created, however with `act` command i'm getting this error (running `act`):

```
Error: unsupported event: push
```

Example workflow file:

```
workflow "test" {
  on = "pull_request"
  resolves = ["build"]
}

action "build" {
  uses = "./pull-request-action"
}
```

Im not sure why the error refers to `push` event

sosedoff:
Just realized that `act` uses `push` as a default event. When i execute `act pull_request` everything works as expected.

Perhaps `act` could check the workflow file for the event type and run it instead of using a default `push` event.

cplee:
@sosedoff this was intentional, based on the fact that GH actions uses the `push` event as the default as per the [docs](https://developer.github.com/actions/creating-workflows/workflow-configuration-options/#events-supported-in-workflow-files)

![image](https://user-images.githubusercontent.com/2239057/52761947-7ba4f280-2fe3-11e9-93b6-c85af6f4dd1c.png)

cplee:
I'd be open for a PR if you have something else in mind!

sosedoff:
Yeah, i'll experiment a bit on this front. Also, i think i had an older version of `act` installed, once i updated it no longer printed any errors and terminated silently.

mcolyer:
> @sosedoff this was intentional, based on the fact that GH actions uses the `push` event as the default as per the [docs](https://developer.github.com/actions/creating-workflows/workflow-configuration-options/#events-supported-in-workflow-files)

This might have been unintentional on our part. I believe we used the word default here to mean that it's the default event that is selected within the workflow editor. It doesn't really have any significance beyond that. Having support for other events in `act` is definitely encouraged.

cplee:
So `act` does support running other events, but I do like @sosedoff suggestion.  Let's keep this open and implement.

sosedoff:
Ok so i played with workflows a bit more and realized that you can actually have more than one workflow in the same file triggered on the same event (`push`). Im thinking `act` could do the following (to eliminate extra typing):

- Read the workflow file
- If the workflow file contains only a single workflow - detect the event type
- If used did not provide an event we could use the value defined with `on` instead of default `push` 

So if i had a workflow like this:

```
workflow "dummy" {
  on = "pull_request"
  resolves = ["task"]
}
// ...
```

[...]

PR #382, opened 2020-10-06
Simpler list view

lshamis:
Issue https://github.com/nektos/act/issues/260

Simplifies the output of `-l` into a more standard table view.
Also allows for ID discovery, to go with the `-j` flag.

For this repo:
===========================================
Before:
```
# act -l
    ╭──────╮ ╭──────╮
    │ lint │ │ test │
    ╰──────╯ ╰──────╯
            ⬇
 ╭─────────╮ ╭──────────╮
 │ release │ │ snapshot │
 ╰─────────╯ ╰──────────╯
```

After:
```
# act -l
ID        Stage  Name      
lint      0      Lint      
test      0      Test      
snapshot  1      Snapshot  
release   1      Release   
```

For alephzero repo:
===========================================

Before:
```
# act -l
 ╭──────────────────────────────╮ ╭───────────────────────────╮ ╭───────────────────╮ ╭──────────────────────╮ ╭──────────────────────────╮ ╭────────────╮ ╭──────────────╮ ╭──────────╮ ╭────────────────────────────╮ ╭───────────────╮ ╭─────────────╮ ╭────────────╮ ╭─────────────────────────────╮ ╭──────────────────╮ ╭────────╮
 │ Undefined Behavior Sanitizer │ │ Ubuntu Bionic Default GCC │ │ Address Sanitizer │ │ Include What You Use │ │ Ubuntu Focal Default GCC │ │ Clang-Tidy │ │ Alpine Clang │ │ Valgrind │ │ Ubuntu Focal Default Clang │ │ Code Coverage │ │ Alpine Musl │ │ Alpine GCC │ │ Ubuntu Bionic Default Clang │ │ Thread Sanitizer │ │ Format │
 ╰──────────────────────────────╯ ╰───────────────────────────╯ ╰───────────────────╯ ╰──────────────────────╯ ╰──────────────────────────╯ ╰────────────╯ ╰──────────────╯ ╰──────────╯ ╰────────────────────────────╯ ╰───────────────╯ ╰─────────────╯ ╰────────────╯ ╰─────────────────────────────╯ ╰──────────────────╯ ╰────────╯
```

After:
```
# act -l
ID            Stage  Name                          
ubsan         0      Undefined Behavior Sanitizer  
iwyu          0      Include What You Use          
tsan          0      Thread Sanitizer              
clang_tidy    0      Clang-Tidy                    
bionic_gcc    0      Ubuntu Bionic Default GCC     
alpine_musl   0      Alpine Musl                   
[...]

PR #383, opened 2020-10-07
[Add] More logging for default push action

TechIsCool:
When we originally downloaded this tool and ran it. It appeared to do nothing. What actually happened was we had no `push` events in our repo. We turned up logging and didn't get any more helpful information. This adds enough to be helpful.

Adds more Logs for #32


TechIsCool:
Since unit test is failing on another PR as well at the same point I will assume that its accurate but not my code.
`TestRunEvent/issue-228` #382

cplee (approved):

PR #406, opened 2020-11-02
Change handling of ExtraPath

winksaville:
To make the handling of ExtraPath portable append it to the front of the
PATH when it exists, instead of always changing the script.


cplee (approved):
lgtm

winksaville:
Thanks, what do I need to do so `make test` works for me locally?

PR #408, opened 2020-11-04
Use PathListSeparator

winksaville:
In PR #406 I used PathSeparator and I should have used PathListSeparator.
 See: https://golang.org/pkg/os/#pkg-constants

### DIFF
--- a/pkg/runner/step_context.go
+++ b/pkg/runner/step_context.go
@@ -94,7 +94,7 @@ func (sc *StepContext) setupEnv() common.Executor {
 
 		if (rc.ExtraPath != nil) && (len(rc.ExtraPath) > 0) {
 			s := append(rc.ExtraPath, os.Getenv("PATH"))
-			env["PATH"] = strings.Join(s, string(os.PathSeparator))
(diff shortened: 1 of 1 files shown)


winksaville:
I would like to make a test which tests the ExtraPath handling with something like I used to manually test this change:
```
name: test add path
on: push

jobs:
  job1:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v2
      - name: set-add-path
        run: |
          echo PATH=$PATH
          echo "::add-path::new-path"
      - name: test-add-path
        run: |
          echo PATH=$PATH
          [[ $PATH == new-path:* ]] || exit 1
```

cplee (approved):

winksaville:
@cplee, sorry for my mistake!

I'd sure like to create a test for this, is there a way to do an "end-to-end" test
so that I could execute "act" passing a "string" with code like [above](https://github.com/nektos/act/pull/408#issuecomment-721927833)?

cplee:
@winksaville - yes, there is. check out https://github.com/nektos/act/blob/master/pkg/runner/runner_test.go

PR #411, opened 2020-11-09
Add variadic version of hashFiles

### LINKED ISSUE #324: hashFiles documentation says you can supply multiple paths, but act doesn't let you
I'm not sure how new this is, because I know `hashFiles` used to only allow one path, but the github actions documentation now says you can supply multiple paths: https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#hashfiles

When I try this in `act` though, I get: `[RangeError: expected 1 argument(s); got 3]`

My expression is this: `${{ hashFiles('lib/**/*.ex', 'test/**/*.ex', 'mix.lock') }}`

EDIT: Another interesting point: in `act` you can do `${{ hashFiles('{lib/**/*.ex,test/**/*.ex,mix.lock}') }}`, but github actions will not recognize the multi-glob pattern.

Bigpet:
fixes #324

I don't know any Go, so double check this.

No idea if this needs some file dedup or if the hashes match the GitHub Actions version, this is just to make sure the same *.yaml code doesn't fail in act while it works in the GH Actions runner

### DIFF
--- a/pkg/runner/expression.go
+++ b/pkg/runner/expression.go
@@ -236,11 +236,15 @@ func vmFromJSON(vm *otto.Otto) {
 
 func (rc *RunContext) vmHashFiles() func(*otto.Otto) {
 	return func(vm *otto.Otto) {
-		_ = vm.Set("hashFiles", func(path string) string {
-			files, _, err := glob.Glob([]string{filepath.Join(rc.Config.Workdir, path)})
-			if err != nil {
-				logrus.Errorf("Unable to glob.Glob: %v", err)
-				return ""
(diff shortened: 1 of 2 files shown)


cplee (approved):
looks great!

CODY-RAY:
@Bigpet it seems to break when there is a `.` in the file name

`DEBU[0015] Evaluating 'hashFiles('**/package-lock['json']')' instead of 'hashFiles('**/package-lock.json')' 
ERRO[0015] Unable to interpolate string '${{ hashFiles('**/package-lock.json') }}' - [(anonymous): Line 1:29 Unexpected identifier (and 2 more errors)] `

PR #412, opened 2020-11-09
envs in if: - take 2

### LINKED ISSUE #387: Align "if: ${{ always() }}" step behaviour to GH
Good day,

I'm using act 0.2.15 here.

I have a job that creates per-job objects in the cloud so it needs to clean up them regardless whether it was successful or not.
I solved it by defining the last clean up step as:

```yaml
    - name: Cleanup
      if: ${{ always() }}
      run: make cleanup
```

Indeed this step is unconditionally executed when it runs in GitHub actions but not by act. I.e. if one of the previous step fails, act will not run my Cleanup step.

Can you please align this behaviour with the GH?

torbjornvatn:
I discovered that table based tests must have define `assert := a.New(t)` inside the `t.Run` internal func. 
If not, the assert wasn't actually asserting correctly 🙈 

This lead to the discovery of some test cases that wasn't actually working. 
I had to make several changes to the EvalBool function support all the permutations of negations, envs and interpolations

Fixes #387


pkg/runner/expression.go
```diff
@@ -5,22 +5,22 @@ import (
 	"encoding/hex"
 	"encoding/json"
 	"fmt"
+	"github.com/opencontainers/runc/Godeps/_workspace/src/github.com/Sirupsen/logrus"
```
cplee, line 8: can this be removed?

cplee (changes requested):
looks great! one change requested...

torbjornvatn:
@cplee the stray logrus import is removed now.

cplee (approved):

PR #415, opened 2020-11-11
Fixes include when using matrix and strategy build.

SteffenSeckler:
The `include` of the matrix strategy was incorrectly handled.
This simply appends the provided includes to the matrix product, s.t. this works (taken from [the official doc](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#example-including-additional-values-into-combinations))
```yaml
runs-on: ${{ matrix.os }}
strategy:
  matrix:
    os: [macos-latest, windows-latest, ubuntu-18.04]
    node: [4, 6, 8, 10]
    include:
      # includes a new variable of npm with a value of 2
      # for the matrix leg matching the os and version
      - os: windows-latest
        node: 4
        npm: 2
```

Also replaces the comparison of the `commonKeysMatch()` with a `reflect.DeepEqual` to be able to compare maps.
Potentially the entire loop could be replaced that way`

PS:
I do not know how testing works in this project, so I cannot easily add one. But I have seen, that neither `include` nor `exclude` appears to be tested.

### DIFF
--- a/pkg/model/workflow.go
+++ b/pkg/model/workflow.go
@@ -3,6 +3,7 @@ package model
 import (
 	"fmt"
 	"io"
+	"reflect"
 	"regexp"
 	"strings"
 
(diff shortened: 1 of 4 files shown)


pkg/model/workflow.go
```diff
@@ -195,7 +192,7 @@ func (j *Job) GetMatrixes() []map[string]interface{} {
 
 func commonKeysMatch(a map[string]interface{}, b map[string]interface{}) bool {
```
SteffenSeckler, line 193: Potentially, this entire function could be replaced with a `reflect.DeepEqual(a, b)`.
What do you think?

cplee:
@SteffenSeckler  - can you include some test cases for this?  For examples of how to test sample workflow files, see: https://github.com/nektos/act/blob/master/pkg/runner/runner_test.go#L34

cplee (changes requested):
can you include some test cases for this? For examples of how to test sample workflow files, see: https://github.com/nektos/act/blob/master/pkg/runner/runner_test.go#L34

cplee (approved):

Issue #416, opened 2020-11-11
Handling of "env" variables is not correct sometimes


winksaville:
I've created [act-failing-python-cache](https://github.com/winksaville/act-failing-python-cache) with additional details.

This was tested with act master branch as of 11/10/2020:
```
$ git log -1
commit 695c4966843d319933d110316032863d138a721b (HEAD -> master, upstream/master, origin/master, origin/HEAD)
Author: Steffen Seckler <38119442+SteffenSeckler@users.noreply.github.com>
Date:   Tue Nov 10 23:55:27 2020 +0100

    Adds ability to use container images from matrices. (#413)
    
    Uses rc.ExprEval.Interpolate on container image.
```
On a linux computer:
```
$ inxi -Fxz
System:    Kernel: 5.9.2-arch1-1 x86_64 bits: 64 compiler: N/A Desktop: Cinnamon 4.6.7 Distro: Arch Linux 
Machine:   Type: Desktop Mobo: ASRock model: X570 Creator serial: <filter> UEFI: American Megatrends v: P2.10 date: 12/16/2019 
CPU:       Topology: 12-Core model: AMD Ryzen 9 3900X bits: 64 type: MT MCP arch: Zen 2 L2 cache: 6144 KiB 
           flags: avx avx2 lm nx pae sse sse2 sse3 sse4_1 sse4_2 sse4a ssse3 svm bogomips: 182458 
           Speed: 2196 MHz min/max: 2200/3800 MHz Core speeds (MHz): 1: 2200 2: 2199 3: 3707 4: 2010 5: 2197 6: 2200 7: 2200 
           8: 2195 9: 2201 10: 2198 11: 2198 12: 2199 13: 2200 14: 2196 15: 2109 16: 4080 17: 2098 18: 2199 19: 2199 20: 2200 
           21: 2199 22: 2196 23: 2200 24: 2148 
Graphics:  Device-1: NVIDIA TU116 [GeForce GTX 1660] vendor: Micro-Star MSI driver: nvidia v: 455.38 bus ID: 33:00.0 
           Display: x11 server: X.org 1.20.9 driver: nvidia unloaded: modesetting resolution: <xdpyinfo missing> 
           Message: Unable to show advanced data. Required tool glxinfo missing. 
Audio:     Device-1: NVIDIA TU116 High Definition Audio vendor: Micro-Star MSI driver: snd_hda_intel v: kernel bus ID: 33:00.1 
[...]

Issue #384, opened 2020-10-07
Support for $GITHUB_ENV


haizaar:
Good day,
First, thank you for writing this tool - AFAIK it's the only local simulator for any CI out there.

To the topic - GH seem to have moved to a new way to define env vars [1]:
```shell
echo "action_state=yellow" >> $GITHUB_ENV
```
However it doesn't seem to be supported by act. Specifying the following in the workflow file:
```yaml
- name: Set CI instance ID                                                                                               
  run: |                                                                                                                 
    echo GITHUB_ENV=$GITHUB_ENV                                                                                          
    echo 'CI_INSTANCE_ID="-ci${{ github.run_id }}"' >> $GITHUB_ENV  
```
fails with:
```
| GITHUB_ENV=
| /github/workflow/1: line 3: $GITHUB_ENV: ambiguous redirect
```

Can you please support it?

[1] https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#setting-an-environment-variable

haizaar:
Using the `set-env` command also results in a deprecation warning:
```
The `set-env` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/
```

fd0:
Thank you very much for this great tool! This issue also applies to the `add-path` command (e.g. via `echo ::add-path`).

Jaesin:
Another common use is writing to `$GITHUB_PATH`. i.e. `echo "$(composer config -g home)/vendor/bin" >> $GITHUB_PATH` so the path persists in subsequent run steps.  `$GITHUB_PATH` would be a file that would get parsed for paths on every run command. I remember doing something like this manually with travisci a while back so it would be really nice to have this working.

Risto-Stevcev:
Any updates on this?

torbjornvatn:
I've gave this a go yesterday, not completely done yet though.
I'll report back when I get it working.

haizaar:
`set-env` & friends commands are now disabled by default, i.e. warnings became errors.
Now need to set `ACTIONS_ALLOW_UNSECURE_COMMANDS` to make it still work.

torbjornvatn:
@haizaar I'll see if I manage to wrap up my PR on this today

torbjornvatn:
I have something working now. Just have to do some clean up and testing tomorrow

[...]

PR #417, opened 2020-11-12
Act env

torbjornvatn:
This can be used to skip certain steps that you don't want to run locally
when testing. E.g. steps that sends messages to Slack channels on successful
builds etc.


winksaville:
@torbjornvatn I fetched this change hoping it might resolve #416 as its related to handling of "env". Sepcifically in one step there is a `[t1/Linux build and test]   ⚙  ::set-env:: pythonLocation=/opt/hostedtoolcache/Python/3.6.12/x64` and in the next step a `path: ${{ env.pythonLocation }}` operation fails because env.pythonLocation is empty. But if I insert a step between the those steps that displays `env.pythonLocation` then the `path: ${{ env.pythonLocation }}` operation works.

I'd appreciate it if you could take a few minutes to look at #416 and potentially advise me where the problem might exist so as my search for a "real" solution might be faster.

Thanks in advance.

torbjornvatn:
Hi @winksaville. 
Unfortunately I don't think that this particular PR will fix your problem as it's just a small addition to enable skipping of certain steps when running locally with Act.

But hopefully the fix I'm working on for #384 will solve your problem. I'll make sure to add a test case that reads the env in two successive steps.

As a workaround for now you might want to try using the `::set-output` syntax instead https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idoutputs

cplee:
@torbjornvatn - this PR needs some attention now that #424 is merged

torbjornvatn:
@cplee I've merged with master and fixed the conflicts now

cplee (approved):

PR #426, opened 2020-11-20
GitHub env file support

### LINKED ISSUE #384: Support for $GITHUB_ENV

torbjornvatn:
This PR attempts to add support for the new [Environment Files](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#environment-files) since the `::set-env` command has been deprecated.

I found a way to download a tar of the `envs.txt` file that I've assigned to `$GITHUB_ENV` from a previous step's container and merge the contents into the env map that is passed to the next step.
It's possible that there are more elegant ways of doing this, but I couldn't find any other way of getting access to that file between steps.

UPDATE: I did an new attempt at solving this using the existing command handler, but it didn't work as the new way of setting the envs uses the >> to file syntax. That output can be generated from any command that's called within the `run:` and as far as I understand that can't be intercepted in the command handler.

DEPRECATION? We should probably add a warning or an error if you try to use the old and deprecated `::set-env` syntax, but I'm not sure how to best handle that @cplee 

Fixes #384


cplee:
@torbjornvatn - i really appreciate you working on this! 

I'm going to need more time to review...lots of changes and it is a bit hard to discern which changes are related to #384 and which were just general "clean up" changes.

dkravetz:
@bixu and I are also running into this issue. I'd be happy to test a release candidate :smiley:

cplee:
@torbjornvatn  - can you fix the conflict?

torbjornvatn:
@cplee sorry for the delay!
Fixed the conflict with master now

torbjornvatn:
@cplee I've also fixed the linting issue now.

cplee (approved):

2m:
Thank you for this very much! I noticed that writing to `GITHUB_PATH` to [add to system path](https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#adding-a-system-path) does not currently work. However it is possible to work around it with:

```
echo "PATH=$PATH:$(python3 -m site --user-base)/bin" >> ${GITHUB_ENV}
```

PR #429, opened 2020-11-22
Extract runTestJobFile from TestRunEvent

winksaville:
This will allow other tests to be created in runner_test.go that can
be tested individually.


winksaville:
I'd like to git this merged before we take #423 which I'll rebase on top of this PR.

cplee (approved):

PR #466, opened 2021-01-11
Use correct syntax for Powershell Scripts as file

### LINKED ISSUE #467: Doesn't correctly handle worker-level Powershell script runs
### Issue
Powershell 7 scripts do not run in native workers via `run:` step, and report a false positive of succeeding

### Steps to Reproduce
With the following simple config:
```yaml
name: Test
on:
  push:
  pull_request:
    branches:
      - main

jobs:
  test:
    runs-on: ubuntu-latest
    steps:
      - shell: pwsh
        run: |
          "this should work!"
          "Evidence" > /github/workspace/worked
```
and swapping in the official powershell worker (and using -r so the container doesn't disappear):
`act -rvP ubuntu-latest=mcr.microsoft.com/powershell:lts-ubuntu-18.04`
The result says success:
image

However upon inspecting the image via attaching a shell to the docker container the "evidence" file was not created:
image

And running the command manually in the container from shell shows an exec problem
image

Running the command using the `-f` format works fine however
image
image

JustinGrote:
"Worker" Powershell scripts do not execute correctly, especially when trying to use an alternate image like -P ubuntu-latest:mcr.microsoft.com/powershell

This updates the command to execute to use the proper `-f` syntax rather than trying to do an inline call operator

Fixes #467 

Attached is the compiled Windows binary for anyone needing this fix until the PR is merged.

[act-PowershellFixed.zip](https://github.com/nektos/act/files/5797997/act-PowershellFixed.zip)

### DIFF
--- a/pkg/runner/run_context.go
+++ b/pkg/runner/run_context.go
@@ -576,8 +576,8 @@ func (rc *RunContext) withGithubEnv(env map[string]string) map[string]string {
 	github := rc.getGithubContext()
 	env["CI"] = "true"
 	env["HOME"] = "/github/home"
-	env["GITHUB_ENV"] = "/github/workflow/envs.txt"
 
+	env["GITHUB_ENV"] = "/github/workflow/envs.txt"
 	env["GITHUB_WORKFLOW"] = github.Workflow
 	env["GITHUB_RUN_ID"] = github.RunID
 	env["GITHUB_RUN_NUMBER"] = github.RunNumber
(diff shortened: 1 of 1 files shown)


JustinGrote:
@cplee the failing snapshot build is not the fault of my PR, let me know when you fix upstream.

JustinGrote:
I should also note that for maximum accuracy, it would output the scriptblock directly to the command this way:
image

However -f to the file should be functionality equivalent for all intents and purposes but also compatible with the other methods of execution.

cplee:
@JustinGrote - thanks for the PR. I've fixed the test upstream. The commands I got for the shells came from GitHub docs. Can you review and sync with your understanding of how it should work?

https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#using-a-specific-shell

JustinGrote:
@cplee it looks like however this is being executed, it's not being interpreted correctly. Their syntax assumes a bash-style shell that is evaluated, are you starting the process directly from go or starting it in a shell? If so then it may take some more evaluation

Using both the quotes method and the scriptblock method, both times it just comes out as echoing whats in between rather than actually executing it
image

Using -f however executes it correctly

JustinGrote:
@cplee I think I see the issue, when I run it manually there's an exec error thrown in there:
image

I think Powershell is thinking you are trying to execute a binary file rather than read a script because you do the chmod +x to the file. I'm going to try to exclude the chmod for powershell scripts.

catthehacker:
Add
```go
		if step.Shell == "pwsh" || step.Shell == "powershell" {
			scriptName = fmt.Sprintf("%s.ps1", scriptName)
		}
```
after
https://github.com/nektos/act/blob/ec6b1624c0b3c4990b93e8a8c2b4c4a5365ed42a/pkg/runner/step_context.go#L160-L160

Issue #107, opened 2020-02-26
command not found (in image node:12.6-buster-slim, as opposed to GH Actions)


fwilhe:
Hi,

as documented [here](https://help.github.com/en/actions/reference/virtual-environments-for-github-hosted-runners#administrative-privileges-of-github-hosted-runners) GitHub configures passwordless sudo, so I can have sudo in my shell scripts to perform individual commands as root.

Given this minimal workflow:

```
on: push
name: Test
jobs:
  build:
    runs-on: ubuntu-latest
    steps:
    - run: sudo id
```

The result on Actions is `uid=0(root) gid=0(root) groups=0(root)`

Running this workflow in `act` yields this result:

```
$ act
[Test/build] 🚀  Start image=node:12.6-buster-slim
[Test/build]   🐳  docker run image=node:12.6-buster-slim entrypoint=["/usr/bin/tail" "-f" "/dev/null"] cmd=[]
[Test/build]   🐳  docker cp src=/Users/fwilhe/temp/test-actions/. dst=/github/workspace
[Test/build] ⭐  Run sudo id
| /github/workflow/0: line 2: sudo: command not found
[Test/build]   ❌  Failure - sudo id
Error: exit with `FAILURE`: 127
```

Not sure what the most elegant solution would be without re-building the `node:12.6-buster-slim` image to install sudo and configure it passwordless. But I think the goal should be to be 1:1 compatible with actions, if possible.

cplee:
See https://github.com/nektos/act#runners.

You can customize the image used for the runs-on.  I've created an image that is a match to GitHub runners, but its like 15GB!

fwilhe:
Hi @cplee 

thanks for the work you put into this.

I know that the image is configurable, and I've locally tested with an image where I installed sudo. So this is kind of a workaround, but the reason I opened this issue is that this limits compatibility with workflows that run on GitHub. My intention was to point out where the documented (and actual) behavior of Actions is unequal the default of `act`, which makes it less useful compared to what it could be.

As mentioned I'm not sure what solution I would propose, as maintaining separate images seems inappropriate.

cplee:
Tough problem.  The runners that GitHub Actions offers are huge.  I didn't feel right using an 18GB image as the default, requiring users of `act` to download that large of an image to run it.

I chose the `node` image because it was required to run node actions and still light enough.  Not sure how to proceed...not sure what good enough looks like for the default docker image.

Open to ideas 🙏

aidansteele:
[...]

PR #479, opened 2021-01-15
Add config support, add survey on first launch when config doesn't exists

CatTheHacker:
Add config support, add survey on first launch when config doesn't exists


cplee:
@CatTheHacker cool idea. I worry a bit though that the ways to config act get confusing. Any way to leverage existing .actrc support?

CatTheHacker:
@cplee I was thinking of leaving `.actrc` as it is now and just build all options on top of config since config files tend to be cleaner and easier to read.
I was also thinking of making `config` subcommand as a way to interact with config file, so it could be either interactive via `act` or users could manually edit config file.
There is also a bit of context for that here: https://github.com/nektos/act/issues/107#issuecomment-760376329

CatTheHacker:
A default config would look like this (or empty, then assume defaults from `act`). Parameters and `.actrc` would overwrite config file options
```yaml
actor: "nektos/act"
bind: false
defaultbranch: master
directory: .
dryrun: false
envfile: .env
eventpath:
graph: false
insecuresecrets: false
job:
list: false
platforms:
  ubuntu-latest: node:12
  ubuntu-18.04: ghcr.io/catthehacker/docker-images:ubuntu-nodejs-18.04
privileged: false
pull: false
quiet: false
reuse: false
secrets:
  A_SECRET: secret-value
  ANOTHER_SECRET: another-secret-value
secretfile:
verbose: false
watch: false
workflows: "./.github/workflows/"
```

cplee:
It seems we could solve the UX issue without introducing another way of configuration (CLI args, env variables, .actrc)

How about this?
* Check if an explicit image was provided by `.actrc` or CLI arg
* If so, use it
* If not, prompt user about selecting a docker image to use (slim vs 20GB)
* Update `~/.actrc` with the docker image that they chose to use

CatTheHacker:
@cplee yeah, that is possible and it's fine, I'll close this PR and make a new one with above logic for survey.

PR #483, opened 2021-01-17
Add survey during first run for a default image

CatTheHacker:
Add survey during first run for a default image
@cplee Finally had a moment to draft that up, let me know how you see it

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -3,7 +3,6 @@ package cmd
 import (
 	"bufio"
 	"context"
-	"fmt"
 	"os"
 	"path/filepath"
 	"regexp"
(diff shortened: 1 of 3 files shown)


cplee:
Looking great @CatTheHacker - thanks for doing this. I pushed a small change to formatting.

CatTheHacker:
PR looks fine for me. I still have some doubts about whether to give more images for "Large" option but that can be fixed later.

cplee (approved):

theowenyoung:
Hi,

I think the condition should be limited at build command, now, when running just `act --version`, I  have to choose an image.

```bash
act --version
```

And, the args checking should accept command line args, not just `~/.actrc`

How do you think? @cplee @catthehacker

PR #485, opened 2021-01-18
Report error when `if` expression is invalid

### LINKED ISSUE #482: if: syntax without ${{ }} seems to always evaluate to false, even if invalid
This issue is similar to #366, but in this case, it appears that `if` statements without `${{ }}` are not evaluated, and always return `false`.  This sample workflow illustrates the issue:

```
name: "Test if statements"
on: push

jobs:
  test:
    runs-on: ubuntu-latest
    steps:
     - name: "bad expression, but act does not evaluate it and skips this step"
       if: foobar1
       run: echo "This does not run"

     - name: "bad expression, and act evaluates it and detects a bad expression "
       if: ${{ foobar2 }}
       run: echo "act will error before reaching this part"
```
For the second step, the logs show:
```
ERRO[0000] Unable to interpolate string '${{ foobar2 }}' - [ReferenceError: 'foobar2' is not defined]
```

This was tested on `v0.2.18`.

edtan:
Fixes #482

### DIFF
--- a/pkg/runner/run_context.go
+++ b/pkg/runner/run_context.go
@@ -314,7 +314,7 @@ func (rc *RunContext) EvalBool(expr string) (bool, error) {
 		joined := strings.Join(evaluatedParts, " ")
 		v, _, err := rc.ExprEval.Evaluate(fmt.Sprintf("Boolean(%s)", joined))
 		if err != nil {
-			return false, nil
+			return false, err
 		}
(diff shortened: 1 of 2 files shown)


cplee (approved):
thank you!

edtan:
No problem, happy to help!  Thank you for creating this awesome tool!

PR #486, opened 2021-01-18
Add autodetect event flag

### LINKED ISSUE #484: Event type auto-detection not working
The event type auto-detection is not working as intended - it is always defaulting to `push`, even though the workflow defines a different event type.  For example, when running `act` for the following workflow, nothing happens.  You must run `act workflow_dispatch` in order for it to run.

```
name: "Test auto-detection"
on: workflow_dispatch

jobs:
  test:
    runs-on: ubuntu-latest
    steps:
     - name: "foo"
       run: echo "bar"
```

It looks like this is happening due to #157:  `PlanEvent("push")` is never `nil`, because `PlanEvent` always creates and returns a new `Plan` object: https://github.com/nektos/act/blob/e6fcfed458dbccd640e093bffd0fc258d6ce0e6e/pkg/model/planner.go#L108-L122.

CatTheHacker:
Add option to detect event type from workflow, closes #484

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,8 @@ or in a shell by running
 
 ```
 # Command structure:
-act [event name to run] [flags]
+act [<event>] [options]
+If no event name passed, will default to "on: push"
 
 # List the actions for the default event:
 act -l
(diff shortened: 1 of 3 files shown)


cplee (approved):
thanks!

PR #503, opened 2021-01-23
Fix Cartesian product to return empty set if empty set is given

### LINKED ISSUE #499: jobs.*.strategy.matrix.include doesnt seem to be parsed correctly when used in a step's env.
Here's an example workflow file that works in GIthub, but not in act:

```
---
name: "Test Platform GH Actions"

on:
  pull_request:
    types:
      - opened
      - synchronize
      - reopened

jobs:
  build:
    strategy:
      matrix:
        include:
          - example_var_1: "this_is_var_1"
            nested_var_1: "this_is_nested_var_1"
    runs-on: ubuntu-latest
    steps:

      - name: Show the environment
        run: |
          env
        env:
          ENV_VAR_1: "${{ matrix.example_var_1 }}"
          NESTED_ENV_VAR_1: "${{ matrix.nested_var_1 }}"
 ```
 
Expected output (truncated to the relevant parts):
```
[Test Platform GH Actions/build-2]   ✅  Success - Show the environment
| ENV_VAR_1=this_is_var_1
| NESTED_ENV_VAR_1=this_is_nested_var_1
```
 
 Actual output:
 
```
[Test Platform GH Actions/build-2]   ✅  Success - Show the environment
| ENV_VAR_1=
| NESTED_ENV_VAR_1=
```

edtan:
This fixes #499, where a matrix strategy with only `include` keys ends up
causing multiple builds.  This bug appears to have been introduced in #415,
when extra `include` keys are added in the matrix strategy.  The cause
seems to be because the CartesianProduct function returns an item with
empty keys, instead of returning an empty set, when given an empty set.

### DIFF
--- a/pkg/common/cartesian.go
+++ b/pkg/common/cartesian.go
@@ -27,7 +27,7 @@ func cartN(a ...[]interface{}) [][]interface{} {
 	for _, a := range a {
 		c *= len(a)
 	}
-	if c == 0 {
+	if c == 0 || len(a) == 0 {
 		return nil
 	}
 	p := make([][]interface{}, c)
--- a/pkg/common/cartesian_test.go
+++ b/pkg/common/cartesian_test.go
@@ -25,4 +25,16 @@ func TestCartesianProduct(t *testing.T) {
 		assert.Contains(v, "baz")
 	}
 
+  input = map[string][]interface{}{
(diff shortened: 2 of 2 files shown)


cplee (approved):
Thank you!

PR #505, opened 2021-01-27
Fix readme

Rasukarusan:
The README says the docker image for ubuntu-18.04 is `node:12.6-buster-slim`, but isn't it correctly `node:12-buster`?
screenshot_ 2021-01-27 23 08 28

Sorry if i'm wrong! Thanks for the nice tool.

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,7 +126,7 @@ GitHub Actions offers managed [virtual environments](https://help.github.com/en/
 | -------------- | ----------------------------------------------------------------- |
 | ubuntu-latest  | [node:12.6-buster-slim](https://hub.docker.com/_/buildpack-deps)  |
 | ubuntu-20.04   | [node:12.6-buster-slim](https://hub.docker.com/_/buildpack-deps)  |
-| ubuntu-18.04   | [node:12.6-buster-slim](https://hub.docker.com/_/buildpack-deps)  |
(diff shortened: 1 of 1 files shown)


Rasukarusan:
```sh
~/Desktop/test  master ✗
$ tree
.
└── .github
    └── workflows
        └── test.yml
```

test.yml
```sh
name: Test Workflow
on:
  push:
    tags:
      - 'v*'
jobs:
  build:
    name: Run Git Commands
    runs-on: ubuntu-18.04
    steps:
      - name: Checkout code
        uses: actions/checkout@v2
        with:
          fetch-depth: 0
      - name: execute command!
        id: set_value_for_formula
        run: |
          git status
```

catthehacker:
The image in README is correct, defined in https://github.com/nektos/act/blob/master/cmd/platforms.go#L9-L12
image

Rasukarusan:
Oh, you are right.
I don't know why `node:12-buster` is running in my environment, but I'll investigate a bit more.
I'll let you know if I find out any details.

Thanks for the relevant code and screenshot!

catthehacker:
The only thing that comes to my mind if the same image is tagged with both names, but then it also should display the one defined in `act`.
You can run `act` with `-v` to enable verbose output and provide the log.

PR #508, opened 2021-01-29
fix: --version, --help do not need check platform config

theowenyoung:
Hi,

Like what I posted at https://github.com/nektos/act/pull/483#issuecomment-769547060

I made a pr for that, how do you think it?

@cplee @catthehacker


catthehacker:
Looks good to me, I've noticed that it asks when running `--version` but didn't had much time lately to look at the code. Thanks for fixing that ❤️

cplee (approved):

PR #525, opened 2021-02-08
Add option to run custom architecture (container platform)

### LINKED ISSUE #493: enable specifying docker --platform argument
Context: For folks on M1 macs that was to run actions as amd64.

catthehacker:
Closes #493


catthehacker:
Ok, I have currently no idea where it fails...

catthehacker:
Current test will fail as it's dependent on #524

pkg/runner/run_context.go
```diff
@@ -89,6 +89,14 @@ func (rc *RunContext) startJobContainer() common.Executor {
 			binds = append(binds, fmt.Sprintf("%s:%s%s", rc.Config.Workdir, "/github/workspace", bindModifiers))
 		}
 
+		if rc.Config.ContainerPlatform == "" {
+			if runtime.GOOS == "darwin" {
+				rc.Config.ContainerPlatform = fmt.Sprintf("%s/%s", "darwin", runtime.GOARCH)
```
catthehacker, line 94: `darwin/*` platform doesn't exists 🤦 will fix that later

catthehacker:
@cplee Let me know if there is anything wrong.
I've also renamed the parameter lately to `--container-architecture` since `platform` could be confused with `-P, --platform stringArray   custom image to use per platform`.
Whenever it's mentioned in `act` context it's `containerArchitecture` and when it touches Docker API it's `Platform`.

cplee (approved):

pkg/runner/step_context.go
```diff
@@ -354,10 +359,35 @@ func (sc *StepContext) runAction(actionDir string, actionPath string) common.Exe
...
+					Platform:   rc.Config.ContainerArchitecture,
 				})
+				exists, err = container.ImageExistsLocally(ctx, image, rc.Config.ContainerArchitecture)
+				if err != nil {
+					return err
+				}
+
+				if !exists {
+					return err
+				}
```
ZauberNerd, line 390: The code always exits here with err `nil` and the code below won't get executed, which causes local docker actions to be skipped.
This is caused by `container.NewDockerBuildExecutor` returning a function which is not executed here, but in line 483.
/cc @catthehacker @cplee

PR #198, opened 2020-04-17
disable gitignore for actions

alanb-sony:
fixes #193

PR #537, opened 2021-02-19
Add option to allow files/directories in `.gitignore` to be copied to container

### LINKED ISSUE #193: .gitignore should be ignored for actions
I'm trying to use the [get-cmake](https://github.com/lukka/get-cmake) action to download cmake. Strangely it has `dist` listed in its .gitignore so `dist` doesn't get copied into the docker container and the action fails.

### LINKED ISSUE #325: `docker cp` step fails with Python venv inside project directory
I'm using [direnv](https://github.com/direnv/direnv) to automatically create a [Python virtual environment](https://github.com/direnv/direnv/wiki/Python#venv-stdlib-module) inside my project directory. Trying to run `act` gives me the following issue:

```
act         
[test.yml/lint] 🚀  Start image=node:12.6-stretch-slim
[test.yml/lint]   🐳  docker run image=node:12.6-stretch-slim
                      entrypoint=["/usr/bin/tail" "-f" "/dev/null"]
                      cmd=[]
[test.yml/lint]   🐳  docker cp
                      src=/Users/frederic/my_project/.
                      dst=/github/workspace
Error: open /Users/frederic/Kunden/my_project/.direnv/python-venv-3.8.3/bin/python:
no such file or directory
```

Deleting/moving the virtual environment fixes the issue but breaks my project. 😉 

This is related to issue #193 and PR #198, but instead of ignoring files in `.gitignore`, the solution here would be to respect the exclusion patterns in `.gitignore` instead.

It would be great to have a simple switch to enable/disable `.gitignore`, this would solve both issues.

catthehacker:
Continuation of #198
Fixes #193
Fixes #325

### DIFF
--- a/pkg/runner/run_context.go
+++ b/pkg/runner/run_context.go
@@ -129,7 +129,7 @@ func (rc *RunContext) startJobContainer() common.Executor {
 			rc.stopJobContainer(),
 			rc.JobContainer.Create(),
 			rc.JobContainer.Start(false),
-			rc.JobContainer.CopyDir(copyToPath, rc.Config.Workdir+string(filepath.Separator)+".", true).IfBool(copyWorkspace),
+			rc.JobContainer.CopyDir(copyToPath, rc.Config.Workdir+string(filepath.Separator)+".", rc.Config.UseGitIgnore).IfBool(copyWorkspace),
 			rc.JobContainer.Copy(filepath.Dir(rc.Config.Workdir), &container.FileEntry{
 				Name: "workflow/event.json",
 				Mode: 0644,
--- a/pkg/runner/run_context.go
+++ b/pkg/runner/run_context.go
@@ -129,7 +129,7 @@ func (rc *RunContext) startJobContainer() common.Executor {
 			rc.stopJobContainer(),
 			rc.JobContainer.Create(),
 			rc.JobContainer.Start(false),
-			rc.JobContainer.CopyDir(copyToPath, rc.Config.Workdir+string(filepath.Separator)+".").IfBool(copyWorkspace),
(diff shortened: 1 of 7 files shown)


pkg/runner/run_context.go
```diff
@@ -120,7 +120,7 @@ func (rc *RunContext) startJobContainer() common.Executor {
 			rc.stopJobContainer(),
 			rc.JobContainer.Create(),
 			rc.JobContainer.Start(false),
-			rc.JobContainer.CopyDir(copyToPath, rc.Config.Workdir+string(filepath.Separator)+".").IfBool(copyWorkspace),
+			rc.JobContainer.CopyDir(copyToPath, rc.Config.Workdir+string(filepath.Separator)+".", rc.Config.AllowGitignore).IfBool(copyWorkspace),
```
cplee, line 123: I don't think this is working as intended. The `--allow-gitignore` flag is default to `false` so the default behavior is now to "allow git ignore" entries. I think the current default behavior should be preserved.

cplee (changes requested):

pkg/runner/run_context.go
catthehacker: I think I got confused somewhere around looking through code and PR. Will correct that this weekend (or earlier if I'll get some time)

catthehacker: Fixed

wagenet:
I think we still need this.

cplee:
@catthehacker  - i'm ready to merge this once conflicts are resolved

cplee:
@catthehacker  - looks like there was a regression here. We are back to the `disallowGitIgnore` on command line but then `useGitIgnore` internally

cplee (changes requested):
@catthehacker - looks like there was a regression here. We are back to the disallowGitIgnore on command line but then useGitIgnore internally

cplee (approved):

Issue #228, opened 2020-05-08
MODULE_NOT_FOUND during `docker cp` command


pmatheson-greenphire:
Command I'm running
`act -P ubuntu-latest=nektos/act-environments-ubuntu:18.04 `

Error output
```
 ☁  git clone 'https://github.com/gittools/actions' # ref=v0.9.2
[]   🐳  docker cp src=/home/user/.cache/act/gittools-actions-gitversion-setup@v0.9.2 dst=/actions/
| internal/modules/cjs/loader.js:985
|   throw err;
|   ^
| 
| Error: Cannot find module '/actions/gittools-actions-gitversion-setup@v0.9.2/dist/github/gitversion/setup/bundle.js'
|     at Function.Module._resolveFilename (internal/modules/cjs/loader.js:982:15)
|     at Function.Module._load (internal/modules/cjs/loader.js:864:27)
|     at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:74:12)
|     at internal/main/run_main_module.js:18:47 {
|   code: 'MODULE_NOT_FOUND',

```
it looks like the code from gittools-actions-gitversion-setup@v0.9.2 is copied to /actions/ in the container but the action javascript expects the original directory name to be in the path. So the cp should be `cp /home/user/.cache/act/gittools-actions-gitversion-setup@v0.9.2 /actions/gittools-actions-gitversion-setup@v0.9.2`

rdlf0:
I had some hard time with this, too, but after some digging in the code I found that in order to get the correct `dist` path you have to provide the `path` key in the `actions/checkout` step like this:
```yml
steps:
  - name: Checkout
    uses: actions/checkout@v2
    with:
      path: "your-action-root-directory"
```

pmatheson-greenphire:
This doesn't seem like it would solve the problem - it just changes the destination directory of the initial checkout from `/github/workflow/` to `/github/workflow/${PATH}`

rdlf0:
> This doesn't seem like it would solve the problem - it just changes the destination directory of the initial checkout from `/github/workflow/` to `/github/workflow/${PATH}`

Yup! And that's how the path provided to `cp` becomes the one you suggested and the `MODULE_NOT_FOUND` error is gone. As far as I can understand your issue this seems to be the case. In my situation this was the solution.

rstuven:
[...]

PR #552, opened 2021-03-05
Add more context to MODULE_NOT_FOUND errors and how to fix it

alfredodeza:
This update follows many comments on related issues that don't really follow what `"your-action-root-directory"` is. In my case, I stumbled on this before, and couldn't remember what the fix was a few months later. 

The PR gives more context as to what exactly is needed to make this work for running local actions that are getting copied over.

Related comments:

* https://github.com/nektos/act/issues/228#issuecomment-753375358
* https://github.com/nektos/act/issues/228#issuecomment-744428603

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,15 +114,25 @@ It will save that information to `~/.actrc`, please refer to [Configuration](#co
 
 # Known Issues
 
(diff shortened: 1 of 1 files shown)


cplee (approved):

alfredodeza:
Thanks for the merge @cplee !

Issue #465, opened 2021-01-10
Miss steps context outcome, conclusion


theowenyoung:
Hi, Currently, the steps context does not match the Github action.

Docs: https://docs.github.com/en/free-pro-team@latest/actions/reference/context-and-expression-syntax-for-github-actions#steps-context

For example:

```yaml
steps:
  - id: test
    run: echo test
  - run: echo $outcome
    env:
      outcome: steps.test.outcome
```

Currently the step context:

```json
{"success": true,"outputs": {}}
```

Expected:

```json
{
  "outcome": "success",
  "conclusion": "success",
  "outputs": {
  }
}
```

Code: https://github.com/nektos/act/blob/e47a2395a5260aca1cec1506f0243e919ed4e70c/pkg/runner/run_context.go#L40-L43

catthehacker:
Will look at it.

PR #557, opened 2021-03-08
Set step results to map more closely to reality

pezholio:
This partially addresses #465 by setting the outcome to "success" or "failure" depending whether a step passes or fails (see https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#steps-context)

I don't know enough Go to address the conclusion part of the output, but this is an improvement on what we have before.

### DIFF
--- a/pkg/runner/run_context.go
+++ b/pkg/runner/run_context.go
@@ -38,7 +38,7 @@ func (rc *RunContext) String() string {
 }
 
 type stepResult struct {
-	Success bool              `json:"success"`
+	Outcome string            `json:"outcome"`
 	Outputs map[string]string `json:"outputs"`
 }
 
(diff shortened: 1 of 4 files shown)


pkg/runner/run_context.go
```diff
@@ -230,9 +230,9 @@ func (rc *RunContext) newStepExecutor(step *model.Step) common.Executor {
 			if sc.Step.ContinueOnError {
 				common.Logger(ctx).Infof("Failed but continue next step")
 				err = nil
-				rc.StepResults[rc.CurrentStep].Success = true
+				rc.StepResults[rc.CurrentStep].Outcome = "failure"
```
cplee, line 233: @pezholio - won't this break the `continue on error` path?

PR #563, opened 2021-03-12
Add code testing on MacOS

catthehacker:
Add code testing on MacOS. Performance is absolutely terrible on GHA (linux: <2 minutes, macos: >20 minutes) but hey! it works!

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -6,81 +6,96 @@ jobs:
     name: Lint
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2
-    - uses: golangci/golangci-lint-action@v2
-      env:
-        CGO_ENABLED: 0
-      with:
-        version: v1.32.2
+      - uses: actions/checkout@v2
+      - uses: golangci/golangci-lint-action@v2
+        env:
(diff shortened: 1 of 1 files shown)


cplee:
@catthehacker - is the slow part in the execution of the build or tests? Not sure I want to add a 20 min delay on each build 🤔

catthehacker:
@cplee I know that MacOS runners on GitHub Actions have very bad performance especially with Docker running in VirtualBox machine, so I'm pretty sure that affects the testing phase and there shouldn't be any issue with building but I do not own any machine with MacOS so I can't tell how it looks like on normal Apple devices.
I'll try to get something up to verify it.

catthehacker:
@cplee I couldn't find any way to test that myself locally yet. Would it be okay to merge PR if I remove dependency on MacOS test and add `continue-on-error: true` for MacOS?

cplee (approved):

cplee:
@catthehacker - ready to merge once conflict is addressed.

catthehacker:
@cplee Fixed conflicts and I've added `continue-on-error: true` to macos test as a failsafe so it doesn't break rest of workflow

cplee (approved):

PR #566, opened 2021-03-15
$GITHUB_PATH support

torbjornvatn:
Adds support for $GITHUB_PATH as described [here](https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#adding-a-system-path) as a replacement for the deprecated `::add-path` syntax


Shigeki-Nagaya:
@cplee 
Merging Blocked ! PLS marge it.
Thanks in advance.!

pkg/runner/run_context.go
```diff
@@ -578,6 +582,7 @@ func (rc *RunContext) withGithubEnv(env map[string]string) map[string]string {
 	env["HOME"] = "/github/home"
 
 	env["GITHUB_ENV"] = "/github/workflow/envs.txt"
+  env["GITHUB_PATH"] = "/github/workflow/paths.txt"
```
cplee, line 585:
```suggestion
	env["GITHUB_PATH"] = "/github/workflow/paths.txt"
```

cplee:
@torbjornvatn  - ready to merge, just need to resolve conflicts and also fix whitespace on line `585` of `run_context.go`

pkg/runner/run_context.go
```diff
@@ -584,10 +584,9 @@ func withDefaultBranch(b string, event map[string]interface{}) map[string]interf
 func (rc *RunContext) withGithubEnv(env map[string]string) map[string]string {
 	github := rc.getGithubContext()
 	env["CI"] = "true"
-	env["HOME"] = "/github/home"
-
-	env["GITHUB_ENV"] = "/github/workflow/envs.txt"
-  env["GITHUB_PATH"] = "/github/workflow/paths.txt"
+	env["HOME"] = fmt.Sprintf("%s/%s", filepath.Dir(rc.Config.Workdir), "home")
+	env["GITHUB_ENV"] = fmt.Sprintf("%s/%s", filepath.Dir(rc.Config.Workdir), "workflow/envs.txt")
+	env["GITHUB_ENV"] = fmt.Sprintf("%s/%s", filepath.Dir(rc.Config.Workdir), "workflow/paths.txt")
```
cplee, line 589:
```suggestion
	env["GITHUB_PATH"] = fmt.Sprintf("%s/%s", filepath.Dir(rc.Config.Workdir), "workflow/paths.txt")
```

cplee (changes requested):
looks like a typo

pkg/runner/run_context.go
torbjornvatn: Yes, I've fixed it now.

But I just figured out that the test I wrote for setting the PATH from GITHUB_PATH doesn't work anymore (I think it did work earlier)

When I fetch $PATH inside a step in my test it's always `/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin`, so it doesn't seem to be updated even though the envList that is passed to `ContainerExecCreate` in `docker_run.go` contains the updated PATH.
This might be that way `ContainerExecCreate` has worked all along, but this test run worked for instance https://github.com/nektos/act/actions/runs/675631386

torbjornvatn: I've been looking into ways to update the PATH of a container before a ContainerExecCreate, but maybe the only way around this is to prepend the command with a  `fmt.Sprintf("export PATH=%s", env["PATH"])`

catthehacker: Wouldn't that break on different shells (`zsh`, `pwsh`)?

catthehacker: I'm suspecting that #570 breaks `PATH`, currently testing it on your branch

torbjornvatn: Probably, so one would need to set the PATH env in a switch on `s.Shell` like in [workflow.go](https://github.com/nektos/act/blob/master/pkg/model/workflow.go#L257)

Did a quick test now locally with `run = fmt.Sprintf("export PATH=%s; %s", sc.Env["UPDATED_PATH"], run)` in `step_context.go` and that seems to work. 
So now I just have to figure out how to set the PATH env for the different shells returned by `ShellCommand()`

torbjornvatn: > I'm suspecting that #570 breaks `PATH`, currently testing it on your branch

Aha, that's something that has changed since my first commit. 
I can confirm that swapping back to `bash --noprofile --norc -e {0}` does make my test pass

catthehacker: Yep, tried it now and it works for me too. I would be in favour of reverting #570 if there would be any other way to source `/etc/environment`

catthehacker: Also a heads up, there is another PR that touches shells => #575

torbjornvatn: I can see if I get some spare time over Easter to look at the `/etc/environment` merging with `env`

torbjornvatn:
Ah, I never got around to look into this. Sorry!

catthehacker:
@torbjornvatn rebased your branch on latest changes and added composite function and it seems to work fine but GitHub does not allows me to make PR to your repo 😞 
https://github.com/catthehacker/act-fork/tree/GITHUB_PATH

torbjornvatn:
I added your changes @catthehacker and merged with master @cplee, but codecov is a bit unhappy.
And the `go test -v -timeout 30m -cover ./...` command  runs just fine on my mac.

I didn't look into writing tests for the new `extractGithubEnv` and `extractGithubPath` functions as they call out to docker and I don't know how to best handle that in a test.
Suggestions?

pkg/runner/step_context.go
```diff
@@ -252,6 +252,10 @@ func (sc *StepContext) newStepContainer(ctx context.Context, image string, cmd [
 		binds = append(binds, fmt.Sprintf("%s:%s%s", rc.Config.Workdir, rc.Config.Workdir, bindModifiers))
 	}
 
+	if rc.Config.ContainerArchitecture == "" {
+		rc.Config.ContainerArchitecture = fmt.Sprintf("%s/%s", "linux", runtime.GOARCH)
+	}
+
```
catthehacker, line 258: I think this has been added as mistake? It was recently removed as of #619

pkg/runner/step_context.go
```diff
@@ -410,150 +407,170 @@ func (sc *StepContext) runAction(actionDir string, actionPath string) common.Exe
...
-						ImageTag:   image,
-						Platform:   rc.Config.ContainerArchitecture,
-					})
-				} else {
-					log.Debugf("image '%s' for architecture '%s' already exists", image, rc.Config.ContainerArchitecture)
-				}
-			}
+		exists, err := container.ImageExistsLocally(ctx, image, "any")
+		if err != nil {
+			return err
```
catthehacker, line 438: This also seems like a mistaken regress of #619

catthehacker (changes requested):

catthehacker:
Codecov is unhappy about everything since target is 100% so don't mind it :P

> I didn't look into writing tests for the new `extractGithubEnv` and `extractGithubPath` functions as they call out to docker and I don't know how to best handle that in a test.
> Suggestions?

Will look later into it

pkg/runner/step_context.go
torbjornvatn: Yes, sorry about these.
I hope I've fixed them in https://github.com/nektos/act/pull/566/commits/a6fa584e71dd0bcbef366605e03de3e9cd3fbcbe

torbjornvatn:
@cplee and @catthehacker I think I got it now. The linter still complains about cyclomatic complexity, but I think that is material for another PR.

pkg/runner/runner_test.go
```diff
@@ -105,6 +105,7 @@ func TestRunEvent(t *testing.T) {
 		{"testdata", "defaults-run", "push", "", platforms, ""},
 		{"testdata", "uses-composite", "push", "", platforms, ""},
 		{"testdata", "issue-597", "push", "", platforms, ""},
+		{"testdata", "env-and-path", "push", "", platforms, "linux/arm64"},
```
catthehacker, line 108: since tests were updated to not include arch
```suggestion
		{"testdata", "env-and-path", "push", "", platforms, ""},
```

catthehacker (changes requested):

catthehacker (approved):

pkg/runner/runner_test.go
torbjornvatn: Fixed it @catthehacker

cplee:
@torbjornvatn - i'd rather not break tests on `master`...they are currently all green. Can you address the 2 lint errors?

- One is empty line that needs to be deleted
- One can be addressed by adding `// nolint: gocyclo` before the `func`

torbjornvatn:
@cplee WOHOO! All green!

pkg/runner/step_context.go
```diff
@@ -114,7 +114,7 @@ func (sc *StepContext) mergeEnv() map[string]string {
 
 	env["PATH"] = `/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin`
```
catthehacker, line 115: this can perhaps be removed since `PATH` is taken directly from container
```suggestion

```

catthehacker (approved):
sorry for pushing my PR right before yours 😸 i just can't believe I have not seen such simple solution earlier

pkg/runner/step_context.go
catthehacker: Never mind, it should stay

cplee (approved):

catthehacker:
🎉🎉🎉

Issue #555, opened 2021-03-06
Project directory is not mount inside act container


rockandska:
## Act version

```none
act version 0.2.20
```

## Description

Since the act container mount the host docker socket instead of using DinD, tests who use docker and mount project dir as itself (docker run -v $(pwd):$(pwd) example-tests) fail because `/github/workspace` doesn't exist on the host side.

## Expected behaviour

```
$ act 2> /dev/null
[Tests/docker-test] 🚀  Start image=nektos/act-environments-ubuntu:18.04
[Tests/docker-test]   🐳  docker run image=nektos/act-environments-ubuntu:18.04 entrypoint=["/usr/bin/tail" "-f" "/dev/null"] cmd=[]
[Tests/docker-test]   🐳  docker cp src=/tmp/act-test/. dst=/github/workspace
[Tests/docker-test] ⭐  Run actions/checkout@v2
[Tests/docker-test]   ✅  Success - actions/checkout@v2
[Tests/docker-test] ⭐  Run show workflow
| name: Tests
| on: [push, pull_request]
| jobs:
|   docker-test:
|     runs-on: ubuntu-18.04
|     steps:
|       - uses: actions/checkout@v2
|       - name: show workflow
|         run: cat $(pwd)/.github/workflows/test.yml
|       - name: show workflow in docker
|         run: docker run -v $(pwd):$(pwd) bash:3.2.57 cat $(pwd)/.github/workflows/test.yml
[Tests/docker-test]   ✅  Success - show workflow
[Tests/docker-test] ⭐  Run show workflow in docker
| name: Tests
| on: [push, pull_request]
| jobs:
|   docker-test:
|     runs-on: ubuntu-18.04
|     steps:
|       - uses: actions/checkout@v2
|       - name: show workflow
|         run: cat $(pwd)/.github/workflows/test.yml
|       - name: show workflow in docker
|         run: docker run -v $(pwd):$(pwd) bash:3.2.57 cat $(pwd)/.github/workflows/test.yml
[Tests/docker-test]   ✅  Success - show workflow in docker
```

## Actual behaviour

```
$ act 2> /dev/null
[Tests/docker-test] 🚀  Start image=nektos/act-environments-ubuntu:18.04
[Tests/docker-test]   🐳  docker run image=nektos/act-environments-ubuntu:18.04 entrypoint=["/usr/bin/tail" "-f" "/dev/null"] cmd=[]
[...]

Issue #551, opened 2021-03-04
Issue: Running ACT on BigSur 11.2.2 fails due to Mounts denied on /github/workspace


ian-lewis-d:
## Act version

act version 0.2.20

## Expected behaviour

Running $ act -j vetting should run the vetting section in .github/workflows/build.xml

## Actual behaviour

The process runs stops with the following message:

```
[Build/Vetting        ] Extracting content to '/github/'
[Build/Vetting        ] Exec command '[bash --noprofile --norc -eo pipefail /github/workflow/1]'
| docker: Error response from daemon: Mounts denied: 
| The path /github/workspace
| is not shared from OS X and is not known to Docker.
| You can configure shared paths from Docker -> Preferences... -> File Sharing.
| See https://docs.docker.com/docker-for-mac/osxfs/#namespaces for more info.
| .
| ERRO[0000] error waiting for container: context canceled 
```

This is on a Mac OS BigSur 11.2.2 with the default Sandboxing enabled.

I have attempted downgrading Docker Desktop (to 2.5.0.1 from 3.1 latest), Turning off gRPC Fuse and none of this gets me anywhere.

It seems, from what I can work out, that the BigSur MacOS has made root a Read Only file system so adding /github/workflow will fail. This leads to the mount failing as there is no mount point. This is my current theory but I have only limited experience with Docker, Github Actions and Mac computers.

## Workflow and/or repository

<details>
  <summary>workflow</summary>

```none
name: example workflow

on: [push]

jobs:
  vetting:
    name: Vetting
    runs-on: ubuntu-latest
    timeout-minutes: 5
    steps:
      - name: Checkout Repo
        uses: actions/checkout@v2
      - name: Project Validation
        run: docker run --rm -v $(pwd):/app ******/devtools:project-validator
```

</details>

## Steps to reproduce

1. Clone repo (private)
2. Enter cloned repo directory
3. Run `act -j vetting`

## `act` output

<details>
  <summary>Log</summary>

```
DEBU[0000] Loading environment from /Users/user/projects/[REDACTED]/.env 
DEBU[0000] Loading secrets from /Users/user/projects/[REDACTED]/.secrets 
DEBU[0000] Loading workflows from '/Users/user/projects/[REDACTED]/.github/workflows' 
DEBU[0000] Reading workflow '/Users/user/projects/[REDACTED]/.github/workflows/build.yml' 
DEBU[0000] Planning event: push                         
[...]

Issue #410, opened 2020-11-07
Mounting volumes into docker not working


agutenkunst:
If inside the job a volume is mounted into a docker container the mounted folder will be empty in this container.
I create a minimal example
```
name: CI

on: [push, pull_request]

jobs:
  my_job:
    runs-on: ubuntu-18.04
    steps:
      - run: "mkdir /tmp/folder && touch /tmp/folder/file && docker run -t -v /tmp/folder:/folder --entrypoint '' ubuntu:18.04 /bin/ls /folder"
```
In gh actions this works fine: See https://github.com/agutenkunst/nectos-act-docker-volume-mount-issue/runs/1368072778#step:2:20 ("file" is printed by /bin/ls)

Running locally
```
act -P ubuntu-18.04=nektos/act-environments-ubuntu:18.04
```
does not give any output, indicating a failed mount.
```
[CI/my_job] 🚀  Start image=nektos/act-environments-ubuntu:18.04
[CI/my_job]   🐳  docker run image=nektos/act-environments-ubuntu:18.04 entrypoint=["/usr/bin/tail" "-f" "/dev/null"] cmd=[]
[CI/my_job] ⭐  Run mkdir /tmp/folder && touch /tmp/folder/file && docker run -t -v /tmp/folder:/folder --entrypoint '' ubuntu:18.04 /bin/ls /folder
[CI/my_job]   ✅  Success - mkdir /tmp/folder && touch /tmp/folder/file && docker run -t -v /tmp/folder:/folder --entrypoint '' ubuntu:18.04 /bin/ls /folder
```

This is also what I observed in a more complex real world use case. I am not sure if the image is actually Docker-in-Docker ready. Creating docker containers seems to work fine. Just not mounting something into them. 

I check also on github and it seems that the runner user itself is "docker". Sounds like something according to https://docs.docker.com/engine/install/linux-postinstall/ is missing, but I am not sure.

samfoley88:
I've just ran into this problem as well although slightly differently I think my workaround will work for anyone encountering it in future.

**Cause** 
The cause is the fact Docker volume and bind commands (even when running inside an existing container) bind to the HOST file system so will ignore any pre-existing binds (like the bind of /github/workspace).
My best guess is that Github is processing the variables in the command with an awareness that the command is going to docker and so pre-processes the variables and passes the path from the host not the docker container you are already running in.

**Workaround**
[...]

PR #567, opened 2021-03-15
Use project dir instead of /github/workspace

rockandska:
As stated [here](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#file-systems), there is no guaranty that `/github` will be the directory used and actions should not use hard coded `/github` but use the equivalent env var instead.

Using the project dir allow us to mimic this case and in the same way allow users to mount the project dir from inside the container without problem.

Solve #555 
Should works for #551 / #410  too

Also, running `go run main.go` run successfully now.

### DIFF
--- a/pkg/runner/run_context.go
+++ b/pkg/runner/run_context.go
@@ -86,18 +86,18 @@ func (rc *RunContext) startJobContainer() common.Executor {
 			fmt.Sprintf("%s:%s", "/var/run/docker.sock", "/var/run/docker.sock"),
 		}
 		if rc.Config.BindWorkdir {
-			binds = append(binds, fmt.Sprintf("%s:%s%s", rc.Config.Workdir, "/github/workspace", bindModifiers))
+			binds = append(binds, fmt.Sprintf("%s:%s%s", rc.Config.Workdir, rc.Config.Workdir, bindModifiers))
 		}
 
 		rc.JobContainer = container.NewContainer(&container.NewContainerInput{
 			Cmd:        nil,
 			Entrypoint: []string{"/usr/bin/tail", "-f", "/dev/null"},
-			WorkingDir: "/github/workspace",
+			WorkingDir: rc.Config.Workdir,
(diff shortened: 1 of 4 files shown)


JustinGrote:
@rockandska my latest runner checks show the workspace is $HOME/work/<RepoName> on the 'runner' user. Should the default be adjusted? Probably want to be as close to the "real" runner as possible so as to avoid regressions in going from ACT to actions
image

rockandska:
> my latest runner checks show the workspace is $HOME/work/ on the 'runner' user.
> Probably want to be as close to the "real" runner as possible

As explained, the path should bot be static for multiples reasons :
- it is clearly stated in GHA documentation that the path is not guaranteed depending on the runner type/OS (docker, github, self-hosted) and only environment variables should be used in actions / workflows.
- act use the docker host socket to run the containers used in user tests, so the workspace should exist inside the act container and on the host too.

JustinGrote:
@rockandska right, I agree, I'm just saying what the "default" should be if the user doesn't specify it. Just because the docs say it's not guaranteed doesn't mean people won't make actions implicity this way and wonder why they don't work when moved.

That said, maybe it's for the best to catch issues like this early, just wish there was a better way to expose to the user "hey, this action broke in production because you hardcoded a path when you should have used a variable" :)

catthehacker:
@JustinGrote I'm wondering how you get `GITHUB_WORKSPACE` envvar from Docker image since effectively there should be none of those while running `act`.

JustinGrote:
@catthehacker you define it in the Dockerfile with `ENV` and it'll be there when the container starts, then I assume act reads it from the docker container to know where to start placing files via the docker cp and exec commands? I may be reading the code wrong...

rockandska:
@JustinGrote , would you mind to come on gitter to help you
It is off topic and don't want to pollute the PR

Regards,

cplee (approved):

PR #570, opened 2021-03-17
Invoke login shells to source `/etc/environment`

### LINKED ISSUE #265: runner image does not correctly set environment variables
When running with `act -P ubuntu-18.04=nektos/act-environments-ubuntu:18.04` the `CONDA` environment variable is not correctly set:

```
      - name: whatever
        run: echo $CONDA
```

On github this outputs:
```
Run echo $CONDA
  echo $CONDA
  shell: /bin/bash -e {0}
/usr/share/miniconda
```

Locally it outputs:

```
[ci/, ubuntu-18.04-1] ⭐  Run whatever
|
[ci/, ubuntu-18.04-1]   ✅  Success - whatever
```

The file `/etc/environment` does seem to be present in the `nektos/act-environments-ubuntu:18.04` docker image with the correct values, I think it's not being used:

```
docker run --rm -it --entrypoint "" nektos/act-environments-ubuntu:18.04 cat /etc/environment | grep conda
CONDA=/usr/share/miniconda
```

catthehacker:
Fixes #265

### DIFF
--- a/pkg/model/workflow.go
+++ b/pkg/model/workflow.go
@@ -259,13 +259,13 @@ func (s *Step) ShellCommand() string {
 
 	switch s.Shell {
 	case "", "bash":
-		shellCommand = "bash --noprofile --norc -e {0}"
+		shellCommand = "bash --login --norc -e {0}"
 	case "pwsh":
-		shellCommand = "pwsh -command \"& '{0}'\""
+		shellCommand = "pwsh -login -command \"& '{0}'\""
 	case "python":
 		shellCommand = "python {0}"
 	case "sh":
-		shellCommand = "sh -e -c {0}"
+		shellCommand = "sh -l -e -c {0}"
 	case "cmd":
 		shellCommand = "%ComSpec% /D /E:ON /V:OFF /S /C \"CALL \"{0}\"\""
 	case "powershell":


cplee:
These don't match the docs from GitHub: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#using-a-specific-shell

Sure these should be login shells?

catthehacker:
Unfortunately we can't match what GitHub does since they are running full virtual machine. In their case `/etc/environment` is sourced as part of logon process.
Docker containers don't exec the logon process so file has to be either sourced manually (not recommended since it's PAM thing, not shell thing) or force shell to be a login shell (it loads `/etc/profile`, other profile paths and `/etc/environment`).
The only way to fix that issue and not use `--login` would be either through `/etc/profile` and/or `/etc/profile.d/*`, `source`ing `/etc/environment` or reading `/etc/environment` before any workflow execution and merge it into env in `act`.
As far as I know all shells support `-l` or `--login` since it's what POSIX requires
`bash`:
[`-l` Make bash act as if it had been invoked as a login shell (see INVOCATION below).](https://linux.die.net/man/1/bash)
```
--login
Equivalent to -l.
```
`fish`:
[`-l` or `--login` specify that fish is to run as a login shell](https://fishshell.com/docs/current/cmds/fish.html)
PowerShell 6+ (`pwsh`):
```
-Login | -l
    On Linux and macOS, starts PowerShell as a login shell, using /bin/sh to
    execute login profiles such as /etc/profile and ~/.profile. On Windows,
    this switch does nothing.
```

From https://wiki.archlinux.org/index.php/Command-line_shell#Login_shell
> A *login shell* is an invocation mode, in which the shell reads files intended for one-time initialization,
> such as system-wide `/etc/profile` or the user's `~/.profile` or other shell-specific file(s). These files set
> up the initial environment, which is inherited by all other processes started from the shell
> (including other non-login shells or graphical programs). Hence, they are read only once at the beginning
> of a session, which is, for example, when the user logs in to the console or via SSH, changes the user
> with `sudo` or `su` using the `--login` parameter, or when the user manually invokes a login shell (e.g. by `bash --login`).

From http://manpages.ubuntu.com/manpages/bionic/en/man8/pam_env.8.html
> The `pam_env` PAM module allows the (un)setting of environment variables.
> This module can also parse a file with simple `KEY=VAL` pairs on separate lines
> (`/etc/environment` by default).

From https://askubuntu.com/a/863023
> `/etc/environment` is created and maintained by `pam_env` PAM module, not by your shell
> (so you should not `source` it). This file is read by typical login programs. So you
> need to logout/login in the usual manner to see the changes (assuming you are not
> modifying the `PATH` from somewhere afterwards).

catthehacker:
Ideally to fix that problem we would like to:
- spin up container
- grab `/etc/environment`
- shut down container
- merge file content with https://github.com/nektos/act/blob/41b03b581c7aec251d96e9a5ea27009542a798d4/pkg/runner/run_context.go#L75

cplee (approved):

PR #575, opened 2021-03-23
Update Shell Tasks to match ScriptHandlerHelpers

### LINKED ISSUE #467: Doesn't correctly handle worker-level Powershell script runs

JustinGrote:
Code: https://github.com/actions/runner/blob/main/src/Runner.Worker/Handlers/ScriptHandlerHelpers.cs

Docs: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#using-a-specific-shell

Fixes #467
Replaces #466

Thanks to @catthehacker for the direction

### DIFF
--- a/pkg/model/workflow.go
+++ b/pkg/model/workflow.go
@@ -257,11 +257,12 @@ func (s *Step) GetEnv() map[string]string {
 func (s *Step) ShellCommand() string {
 	shellCommand := ""
 
+	//Reference: https://github.com/actions/runner/blob/8109c962f09d9acc473d92c595ff43afceddb347/src/Runner.Worker/Handlers/ScriptHandlerHelpers.cs#L9-L17
 	switch s.Shell {
 	case "", "bash":
-		shellCommand = "bash --login --norc -e {0}"
+		shellCommand = "bash --login --noprofile --norc -e -o pipefail {0}"
 	case "pwsh":
-		shellCommand = "pwsh -login -command \"& '{0}'\""
+		shellCommand = "pwsh -login -command \". '{0}'\""
 	case "python":
 		shellCommand = "python {0}"
 	case "sh":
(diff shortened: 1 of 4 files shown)


JustinGrote:
@cplee should be ready to review. Verified it works correctly on a runner with pwsh installed.
image

cplee (changes requested):
@JustinGrote  - we've got a conflict here with #570 ... can you take a peek? 👀

JustinGrote:
@cplee per the source link above I made mine match exactly what the actions runner does. I'll defer to @catthehacker if there's a specific issue that has to be addressed with act that isn't covered by the action runner cnofiguration.

JustinGrote:
@cplee I see the other branch was already merged so I can rebase it to include the login shell, that's not a big deal and as @catthehacker describes in https://github.com/nektos/act/pull/570#issuecomment-801560100 using login shell is probably as good of an option as exists without breaking the way that it matches too extensively.

catthehacker (changes requested):
`pkg/runner/act/sergioramos-yarn-actions-install@v6` should be removed

JustinGrote:
@cplee uh, I don't think this test failure is me:
image

cplee:
@JustinGrote - just pushed a fix to your branch...let's see if that gets it 🤞

PR #577, opened 2021-03-24
Fix platforms during initial setup

catthehacker:
Currently when running `act` and `.actrc` doesn't exists as well as no `--platform` was specified, chosen default image for platforms will be ignored and `act` will run with default ones defined in https://github.com/nektos/act/blob/41b03b581c7aec251d96e9a5ea27009542a798d4/cmd/platforms.go#L9-L12

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -198,11 +198,11 @@ func newRunCommand(ctx context.Context, input *Input) func(*cobra.Command, []str
 
 		// check platforms
 		if len(input.platforms) == 0 {
-			actrc := []string{
-				filepath.Join(os.Getenv("HOME"), ".actrc"),
-			}
-			_, err = os.Stat(actrc[0])
-			if os.IsNotExist(err) {
+			actrcHome := filepath.Join(os.Getenv("HOME"), ".actrc")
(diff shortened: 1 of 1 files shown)


cplee:
@catthehacker - i'm not sure i understand this change...the `.actrc` file was intended to be for _any_ args you'd pass to act...not just `--platform`. Won't this change break that?

catthehacker:
@cplee This fix is only for the default image question when `.actrc` doesn't exists in current and home directory and `act` was invoked without `-P`/`--platform`.
Any arg can still be added to `.actrc` by user and it will execute properly. If user creates `.actrc` with args or not, the survey is skipped.

Check if `--platform`/`-P` were used:
https://github.com/nektos/act/blob/af5140f13e25a304cba8d3ac8646342440fc2ad6/cmd/root.go#L199-L200

Check if `.actrc` exists:
https://github.com/nektos/act/blob/af5140f13e25a304cba8d3ac8646342440fc2ad6/cmd/root.go#L201-L205

cplee (approved):
Yep, you're right. This works as you described. Thank you!

Issue #586, opened 2021-03-29
Issue: Unable to build latest `act` due to docker daemon API version 1.41


jaypipes:
## Act version

`af5140f13e25a304cba8d3ac8646342440fc2ad6` commit, also tested with `v0.2.20` (see notes below).

Docker version:

```
[jaypipes@thelio act]$ docker version
Client:
 Version:           19.03.13
 API version:       1.40
 Go version:        go1.15.1
 Git commit:        4484c46
 Built:             Fri Oct  2 19:31:30 2020
 OS/Arch:           linux/amd64
 Experimental:      false

Server:
 Engine:
  Version:          19.03.13
  API version:      1.40 (minimum version 1.12)
  Go version:       go1.15.1
  Git commit:       4484c46
  Built:            Fri Oct  2 00:00:00 2020
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.4.4
  GitCommit:        
 runc:
  Version:          1.0.0-rc92+dev
  GitCommit:        c9a9ce0286785bef3f3c3c87cd1232e535a03e15
 docker-init:
  Version:          0.18.0
  GitCommit:        
```

## Expected behaviour

Be able to build with docker daemon 1.40 and earlier by switching off container image platform discovery on versions that don't support it.

## Actual behaviour
I `git clone`'d the act repo at commit `af5140f13e25a304cba8d3ac8646342440fc2ad6` and ran `go run main.go`:

```
[jaypipes@thelio act]$ go run main.go
<snip go mod downloads...>
[Test what expressions result in true and false on Github/test-ifs-and-buts] 🚀  Start image=catthehacker/ubuntu:act-latest
[push/Test                                                                 ] 🚀  Start image=catthehacker/ubuntu:act-latest
[push/Lint                                                                 ] 🚀  Start image=catthehacker/ubuntu:act-latest
[Test how expressions are handled on Github/test-espressions               ] 🚀  Start image=catthehacker/ubuntu:act-latest
Error: "specify container image platform" requires API version 1.41, but the Docker daemon API version is 1.40
```

**Note**: If you run `make test` it will just hang forever with no output until you Ctrl-C to stop it, and then you'll note a bunch of errors like this:

```
   --- FAIL: TestRunEvent/runs-on#01 (0.11s)
[...]

PR #591, opened 2021-03-30
use container image platform only on docker 1.41+

jaypipes:
Commit af5140f13e25a304cba8d3ac8646342440fc2ad6 introduced support for
specifying a container image platform for cross-platform image building.

Unfortunately, attempting to execute a docker command that includes the
`--platform` flag against Docker daemons using API Version 1.40 and
before results in the following error:

```
"specify container image platform" requires API version 1.41, but the Docker daemon API version is 1.40
```

To allow `act` to be used on the 19.03 Docker CE and earlier versions,
this patch simply checks the Docker daemon API version and only
specifies platform specification when the daemon API version is 1.41 or
greater.

Fixes Issue #586

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,7 +87,7 @@ It will save that information to `~/.actrc`, please refer to [Configuration](#co
 ```none
   -a, --actor string                    user that triggered the event (default "nektos/act")
   -b, --bind                            bind working directory to container, rather than copy
-      --container-architecture string   Architecture which should be used to run containers, e.g.: linux/amd64. Defaults to linux/<your machine architecture> [linux/amd64]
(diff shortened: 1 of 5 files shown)


jaypipes:
@catthehacker @cplee not sure if you want me to add unit tests for these changes? A unit test would likely require mocking out the docker client which I thought was a bit overboard for these small changes... but let me know!

jaypipes:
> Hi @jaypipes, could you also update `README.md` to include Docker API requirements for `--container-architecture` flag?
> 
> https://github.com/nektos/act/blob/737dfaff3d0159e486e436f5517b40187994e952/README.md#L90

done! :)

cplee (approved):
@jaypipes - thanks for adding this. don't sweat the test coverage, we'll merge as-is 👍

catthehacker (approved):

PR #600, opened 2021-04-03
Refactor `.actrc` locations and default image survey

### LINKED ISSUE #599: Enhancement: Follow XDG base directories
## Describe feature
Act is now using `$HOME/.actrc`. In my opinion it shouldn't clutter home directory and should be in `$XDG_CONFIG_HOME/actrc` or `$XDG_CONFIG_HOME/act/config`

catthehacker:
Fixes #599 

Moved actrc locations to own func since it can be reused.
Moved survey part to own func due to high cyclomatic complexity.
Added XDG config dir to config locations.
Replaced using HOME envvar with `go-homedir` module since it's already
an indirect dependency and handles home directories better for each GOOS.


cplee (approved):
Awesome

PR #633, opened 2021-04-23
Prefer go-git to find the reference name

snazy:
Walking the directory tree underneath `.git/refs` is not reliable, as it usually does not
return tags, especially for freshly cloned repos and/or tags fetched from a remote.

The go-git library provides an iterator over all git references.

This approach prefers a reference (tag, branch) from go-git, if found. If none is found,
it falls back to the previous implementation.

### DIFF
--- a/pkg/common/git.go
+++ b/pkg/common/git.go
@@ -71,6 +71,40 @@ func FindGitRef(file string) (string, error) {
 
 	log.Debugf("HEAD points to '%s'", ref)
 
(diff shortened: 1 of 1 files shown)


snazy:
(I'm not an experienced Go developer, so there might be some GoLang related style or GoLang pattern issues in there.)

catthehacker (approved):

cplee (approved):

PR #649, opened 2021-05-01
Fix: pwsh errors with "command not found"

JustinGrote:
pwsh includes the -login switch which causes the split to not work correctly. This fixes that issue.


catthehacker:
I don't think we will merge this, I want to regress the `-login`/`-l` thing.

JustinGrote:
@catthehacker no prob, would you like a separate PR to remove the login switch?

catthehacker:
Currently I don't want to touch most of the codebase until other PRs get merged since there are few quite important fixes that should be in master first (643, 566, 635, 648...).
I want to see where we are after all that.

Issue #329, opened 2020-08-03
ACTIONS_RUNTIME_URL, ACTIONS_RUNTIME_TOKEN and ACTIONS_CACHE_URL environment variables are missing.


quisse:
Github actions have undocumented environment variables named `ACTIONS_RUNTIME_URL`, `ACTIONS_RUNTIME_TOKEN` and `ACTIONS_CACHE_URL`. The [actions/cache](https://github.com/actions/cache) action relies on this to create the url to interact with for storing caches.
https://github.com/actions/toolkit/blob/1cc56db0ff126f4d65aeb83798852e02a2c180c3/packages/cache/src/internal/cacheHttpClient.ts#L33-L47

They all get populated over [here](https://github.com/DavidGOrtega/gh-runner/blob/48b040fa804be4a2f43abee064b365838f94055b/src/Runner.Worker/Handlers/ContainerActionHandler.cs#L178-L185)

This is the cause of #285, #169 is linked.

mattes:
Here is a workaround for the time being:
https://github.com/mattes/github-action-extract-node-env

mattes:
This might be helpful: https://github.com/mattes/github-action-extract-node-env

quisse:
The issue is still persistent afaik.

DivoK:
Any news about this problem? Came here from #285

ayozemr:
Coming from https://github.com/nektos/act/issues/169

jshwi:
Hi, I'm trying to understand the problem. Are these values dynamically set. Can they be manually set and read by act? If so what to?

catthehacker:
> Are these values dynamically set?

@jshwi Those environmental variables are only available to GitHub Actions runners, whether it be runner hosted by GitHub or self-hosted one.

> Can they be manually set and read by act?
Probably.

> If so what to?

No one knows, if it was that easy it would be fixed already.

Perhaps guys from GitHub (@joshmgross from https://github.com/actions/cache or @TingluoHuang from https://github.com/actions/runner, top committers) could chime in here, maybe give us some clues on how we could solve that issue.
I presume it won't be so easy as obtaining and just hardcoding those values since, most likely they are undocumented for a reason and the reason is it's supposed to work only on GitHub Actions _service_ (but please, prove me wrong).
There is artefacts API https://docs.github.com/en/rest/reference/actions#artifacts which could be used to replace artefacts actions but I've not found anything regarding cache.

jshwi:
[...]

Issue #285, opened 2020-06-20
actions/cache@v2 error: Cache Service Url not found, unable to restore cache


grische:
I am not sure how the cache is meant to work with nektos/act, but when working around bug #257, the cache will always throw a `cache-hit=false` with the following errors:
```
☁  git clone 'https://github.com/actions/cache' # ref=v2
🐳  docker cp src=/home/grische/.cache/act/actions-cache@v2 dst=/actions/
❓   ::save-state name=CACHE_KEY::Linux-2b7ee04efeb2a8c4ea27558750b1c2dee27f1692a30af56727638c98655a2410
💬  ::debug::Resolved Keys:
💬  ::debug::["Linux-2b7ee04efeb2a8c4ea27558750b1c2dee27f1692a30af56727638c98655a2410"]
💬  ::debug::Checking zstd --version
💬  ::debug::*** zstd command line interface 64-bits v1.3.3, by Yann Collet ***
💬  ::debug::getCacheEntry - Attempt 1 of 2 failed with error: Cache Service Url not found, unable to restore cache.
💬  ::debug::getCacheEntry - Attempt 2 of 2 failed with error: Cache Service Url not found, unable to restore cache.
| [warning]getCacheEntry failed: Cache Service Url not found, unable to restore cache.
⚙  ::set-output:: cache-hit=false
```

benwinding:
Happens with `actions/cache@v1` too.

Seems that every test of a github action requires **_all_** npm modules to be installed....

Has anyone found a work around?

rafilkmp3:
still searching anithin better than comment lines to use act

quisse:
The `ACTIONS_CACHE_URL` is missing.

benwinding:
Any progress on this? seems to be linked to #329

grische:
I guess this can be tracked as part of #329

BevanR:
Google search suggestions and results suggest that this is a popular topic and issue;

![image](https://user-images.githubusercontent.com/325176/144964022-2589061f-d5bb-4c2c-a13e-9a257cef909c.png)

![image](https://user-images.githubusercontent.com/325176/144964070-875c26e4-8627-4c14-850b-5b618bf4078b.png)

BevanR:
If you are using nektos/act to develop GHA workflows locally and want to get `actions/cache` working so that you can get faster feedback when you test-run on your changes to GHA workflow yaml files;

1. #329 is a dead end. Do not bother reading the long thread.
[...]

Issue #169, opened 2020-03-25
Support artifacts


sandorex:
It would be really helpful if artifacts could be supported

SteadBytes:
:+1: For artifacts, or perhaps at the very least produce and explicit warning/error when used. AFAICT a step using the artifact actions will succeed but no artifact will be uploaded/downloaded.

cplee:
Great call, good feature to add.  Anyone have a sample workflow to use?

jromero:
This is use case we have in [`buildpacks/pack`](https://github.com/buildpacks/pack/blob/master/.github/workflows/build.yml). We ultimately want to run test and build always but then only use the artifacts for a release workflow.

The source can be found here:
Upload:
https://github.com/buildpacks/pack/blob/c9745af053eabf812bf2fdab1695488a01022093/.github/workflows/build.yml#L77-L81

Download:
https://github.com/buildpacks/pack/blob/c9745af053eabf812bf2fdab1695488a01022093/.github/workflows/build.yml#L92-L95

zaguiini:
Is there any progress?

jromero:
Typically these artifacts are shipped via the gh API and the referenced by url.

I don't know how other parts, if any, work with mocking or intercepting GH API calls but I would think that is necessary to properly mock this locally. Am I wrong?

krailler:
How works actions/cache? Seems similar no?

wyTrivail:
we also have the same issue while using upload-artifacts,  we got the error 
```
[CI/build]   💬  ::debug::Root artifact directory is /github/workspace/build
[CI/build]   ❗  ::error::Unable to get ACTIONS_RUNTIME_TOKEN env variable
[CI/build]   ❌  Failure - Upload
```

it would be if it can be mocked.

wagenet:
Seems like this should have remained open. It's a pretty big blocker for full use.

ayozemr:
Artifacts if one of those key actions that many workflows use, and bc of that its a blocker for local dev working... Having this along with cache support would make act the tool we all need to fully develop workflows locally, instead of running workflows  to try things and waiting a lot of time

ayozemr:
Still waiting for this :-(

CatTheHacker:
It's a bit complicated since https://github.com/nektos/act/issues/329

nicolo-ribaudo:
Sorry bot, I hate to just :+1: issues

CreMindES:
[...]

PR #652, opened 2021-05-02
Fix: Allow known failing actions to gracefully continue

JustinGrote:
Some actions such as cache, upload-artifact, and download artifact do not work due to lack of an emulated api/server. 

A potential path forward exists with #169, but until that is implemented we should silently continue on some actions that would otherwise fail but do not really impact the flow of the action for testing, to allow other actions in the flow to be tested.

This is intended to be temporary until the feature is implemented.

Related: #329, #285

(cherry picked from commit 523ad078bac2610989b1162bdf884675843cbdce)

Output from a run against JustinGrote/Press:
image

### DIFF
--- a/pkg/runner/step_context.go
+++ b/pkg/runner/step_context.go
@@ -81,7 +81,7 @@ func (sc *StepContext) Executor() common.Executor {
 		for _, blacklistItem := range unsupportedActionBlacklist {
 			if blacklistItem == actionName {
 				return func(ctx context.Context) error {
-					common.Logger(ctx).Infof("\U0001F6A7  Skipping currently unsupported remote action %v", actionName)
+					common.Logger(ctx).Infof("\U0001F6A7  Skipping remote action that act currently does not support: %v", actionName)
 					return nil
 				}
 			}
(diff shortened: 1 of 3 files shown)


catthehacker:
I don't think we should implement that. There is already a simple way to fix that through `continue-on-error:`.

JustinGrote:
@catthehacker I suppose the env path is a good workaround but it still requires modifying the action itself to accomodate act. Further, new users are going to be confused as to why these very common functions do not work as expected, this is a less hostile approach IMHO.

catthehacker:
But this will break workflows for people who have it working already + https://github.com/nektos/act#skipping-steps

JustinGrote:
@catthehacker I'm confused, how would it break them if they're already skipping the task? it would still skip it regardless as the skip check comes before the remote action check as I recall.

That said for this PR to be viable there should absolutely be a force override command line switch that I should have included.

catthehacker:
It will break for people who are using https://github.com/anthonykawa/artifact-server

catthehacker:
> That said for this PR to be viable there should absolutely be a force override command line switch that I should have included.

This seems to be too much work to just workaround the problem instead of fixing it in `act`.

JustinGrote:
> > That said for this PR to be viable there should absolutely be a force override command line switch that I should have included.
> 
> This seems to be too much work to just workaround the problem instead of fixing it in `act`.

Fair, just wasn't sure how far off that would be since an API has to be emulated, etc.

PR #659, opened 2021-05-04
Mergify: configuration update

cplee:
This change has been made by @cplee from https://mergify.io config editor.


cplee:
@catthehacker - i'm looking to use mergify to improve/cleanup the PR process for act. can you review/approve this for further testing?

catthehacker:
`test-*.yml` will be re-generated from `pkg/runner/*_test.go` so I don't see benefit in deleting them (also since they verify behaviour between `act` and GHA.

catthehacker:
other than that everything looks good to me, I don't know anything about mergify so I leave the decision up to you :)

catthehacker:
@cplee

cplee:
@catthehacker - I'm trying to remove cruft from the PR checks. Those "test-*.yml" end up running on each push. They also don't actually test anything...they just `echo` out a bunch of data to help troubleshoot. 

I'd be fine if they're added back in but with a trigger of `workflow_dispatch` instead of `push`

PR #660, opened 2021-05-04
Remove pwsh -login and add Pwsh test

JustinGrote:
This resolves the issue with pwsh failing due to -login and the command not splitting correctly.

It also adds a test pwsh run using a image that has Powershell integrated, hopefully this doesn't introduce too much into the testing time, but the image can be cached in the CI if necessary to speed things up.

EDIT: Literally took less than half a second to download and extract the container, I guess that's a benefit of using the github container registry :)

### DIFF
--- a/pkg/runner/runner_test.go
+++ b/pkg/runner/runner_test.go
@@ -83,14 +83,13 @@ func TestRunEvent(t *testing.T) {
 	platforms := map[string]string{
 		"ubuntu-latest": "node:12.20.1-buster-slim",
 	}
-	pwshplatforms := map[string]string{
-		"ubuntu-latest": "ghcr.io/justingrote/act-pwsh:latest",
-	}
+
 	tables := []TestJobFileInfo{
 		{"testdata", "basic", "push", "", platforms, ""},
 		{"testdata", "fail", "push", "exit with `FAILURE`: 1", platforms, ""},
 		{"testdata", "runs-on", "push", "", platforms, ""},
-		{"testdata", "powershell", "push", "", pwshplatforms, ""},
+		// Pwsh is not available in default worker (yet) so we use a separate image for testing
+		{"testdata", "powershell", "push", "", map[string]string{"ubuntu-latest": "ghcr.io/justingrote/act-pwsh:latest"}, ""},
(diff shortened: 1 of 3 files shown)


catthehacker (approved):

pkg/runner/runner_test.go
```diff
@@ -83,10 +83,14 @@ func TestRunEvent(t *testing.T) {
...
 		"ubuntu-latest": "node:12.20.1-buster-slim",
 	}
+	pwshplatforms := map[string]string{
+		"ubuntu-latest": "ghcr.io/justingrote/act-pwsh:latest",
+	}
 	tables := []TestJobFileInfo{
 		{"testdata", "basic", "push", "", platforms, ""},
 		{"testdata", "fail", "push", "exit with `FAILURE`: 1", platforms, ""},
 		{"testdata", "runs-on", "push", "", platforms, ""},
+		{"testdata", "powershell", "push", "", pwshplatforms, ""},
```
catthehacker, line 93: perhaps it would be better as inline map instead of single variable

JustinGrote: @catthehacker I did it expecting more tests to utilize it, but I can make it an inline map if you prefer.

JustinGrote: @catthehacker I did it expecting more tests to utilize it, but I can make it an inline map if you prefer.

JustinGrote: Fixed in https://github.com/nektos/act/pull/660/commits/a91679c83c55128751c14c6c5bbb6bdaf02c413c

JustinGrote:
@cplee @catthehacker looks like the composite test is flaky on MacOS, maybe a race condition. Empty commit and it worked this time :)

cplee:
@catthehacker - I'd like to test mergify workflow with this one...can you add an `approve` review to this PR?

catthehacker (approved):

cplee (approved):

cplee:
@Mergifyio refresh

cplee:
Took some persistence, but Mergify is all dialed in

PR #666, opened 2021-05-05
WIP: Enable Non-Root Runner Scenarios

JustinGrote:
Currently act requires the docker container to have root privileges to work because it both binds and mounts to the "same path" as the host.

While the bind method is useful if using act as a task runner, the mount method will fail in non-root containers.

In order to more accurately test Github Actions for both permission and pathing errors, this PR makes mounts use the runner workdir by default and sets permissions accordingly. Binds with the -b flag will be unaffected and will continue to use the same path as what is on the container.

It will also add a parameter to customize the starting container workdir separately from the host workdir, if there is such a situation where you dont' want to use the default (usually /home/runner/workdir)

# Example that this will fix
Try a simple action that does `mkdir -p /my/random/path`. This will work just fine in act, but fail in Github Actions, and you won't find out till you run it in github actions.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,3 @@
+# Set the default behavior, in case people don't have core.autocrlf set.
+# Dockerfile shell scripts will not compile correctly on windows without this
+* text=lf
--- a/pkg/common/git.go
+++ b/pkg/common/git.go
@@ -55,7 +55,7 @@ func FindGitRevision(file string) (shortSha string, sha string, err error) {
 		refBuf = []byte(ref)
 	}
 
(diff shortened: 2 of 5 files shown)


catthehacker:
There is quite a mix of old code and different branches from your fork

JustinGrote:
@catthehacker yeah as mentioned it's draft and WIP, I just rebased to my ActForWindows branch and haven't cleaned up the stuff that the merge missed.

JustinGrote:
Requesting design review and that you agree with the approach before I finish tests.

PR #604, opened 2021-04-04
Add various GitHub environment variables

jsoref:
Using a tool I wrote for something else... https://github.com/jsoref/gha-debug/runs/2017914814?check_suite_focus=true

<details><summary>
GitHub
</summary>

```
AGENT_TOOLSDIRECTORY=/opt/hostedtoolcache
ANDROID_HOME=/usr/local/lib/android/sdk
ANDROID_NDK_HOME=/usr/local/lib/android/sdk/ndk-bundle
ANDROID_NDK_ROOT=/usr/local/lib/android/sdk/ndk-bundle
ANDROID_SDK_ROOT=/usr/local/lib/android/sdk
ANT_HOME=/usr/share/ant
AZURE_EXTENSION_DIR=/opt/az/azcliextensions
BOOST_ROOT_1_72_0=/opt/hostedtoolcache/boost/1.72.0/x64
BOOTSTRAP_HASKELL_NONINTERACTIVE=1
CHROMEWEBDRIVER=/usr/local/share/chrome_driver
CHROME_BIN=/usr/bin/google-chrome
CI=true
CONDA=/usr/share/miniconda
DEBIAN_FRONTEND=noninteractive
DEPLOYMENT_BASEPATH=/opt/runner
DOTNET_MULTILEVEL_LOOKUP=0
DOTNET_NOLOGO=1
DOTNET_SKIP_FIRST_TIME_EXPERIENCE=1
GECKOWEBDRIVER=/usr/local/share/gecko_driver
GITHUB_ACTION=run2
GITHUB_ACTIONS=true
GITHUB_ACTION_REF=
GITHUB_ACTION_REPOSITORY=
GITHUB_ACTOR=jsoref
GITHUB_API_URL=https://api.github.com
GITHUB_BASE_REF=
GITHUB_ENV=/home/runner/work/_temp/_runner_file_commands/set_env_b9659a7c-aad0-4984-83d7-702fbfbb212b
GITHUB_EVENT_NAME=push
GITHUB_EVENT_PATH=/home/runner/work/_temp/_github_workflow/event.json
GITHUB_GRAPHQL_URL=https://api.github.com/graphql
GITHUB_HEAD_REF=
GITHUB_JOB=build
GITHUB_PATH=/home/runner/work/_temp/_runner_file_commands/add_path_b9659a7c-aad0-4984-83d7-702fbfbb212b
GITHUB_REF=refs/heads/main
GITHUB_REPOSITORY=jsoref/gha-debug
[...]

PR #673, opened 2021-05-06
feat: add repository_owner to github context

KnisterPeter:
This features adds the missing `repository_owner` to the github context.

https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#github-context

### DIFF
--- a/pkg/runner/run_context.go
+++ b/pkg/runner/run_context.go
@@ -450,21 +450,22 @@ func (rc *RunContext) getStepsContext() map[string]*stepResult {
 }
 
 type githubContext struct {
-	Event      map[string]interface{} `json:"event"`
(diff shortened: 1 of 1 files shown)


catthehacker:
perhaps it could be added already in #604 ?

KnisterPeter:
Yes, I will add it over there.
But according to the official docs, there is no ENV-Var GITHUB_REPOSITORY_OWNER.
We can create it nontheless but that would be a spec-extension.

KnisterPeter:
Ah, thats a fork. I doubt that I can push on that.

catthehacker:
I added it to #604, but let's keep this open for now

catthehacker:
Merged in other PR!

PR #677, opened 2021-05-07
Asset server implementation

### LINKED ISSUE #169: Support artifacts

KnisterPeter:
# Description

This PR contains a naive github-actions (partially-)compatible asset server.

We only reverse engineered the success case.
The implementation is based on the awesome work of @anthonykawa: https://github.com/anthonykawa/artifact-server

This implementation should be seen as a draft and requires an architecture/design review.

## Missing things

- [x] CLI flags for configuration (port, asset path)
- [x] Server start need to be integrated properly into the runner (we do start it for all commands currently)
- [x] Tests
- [x] Add docs

## Prerequisites

- [x] `--env ACTIONS_RUNTIME_URL=http://localhost:8080/` must be set because it's internally used by the `@action/toolkit`
- [x] `--env ACTIONS_RUNTIME_TOKEN=foo` must be set (we do not have authentication implemented)

~This should ideally be implemented (either as port of this PR or #329), so users don't need to set it on their own.~

These variables are set if act is started with `--artifact-server-path` and configured for the internal server.
It is also possible to define them as environment variables using `--env` for an external artifact-server.

Relates to #329

--

Closes #169

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -285,7 +285,7 @@ func newRunCommand(ctx context.Context, input *Input) func(*cobra.Command, []str
 			return err
 		}
 
-		artifacts.Serve(ctx, input.artifactServerPath, input.artifactServerPort)
+		cancel := artifacts.Serve(ctx, input.artifactServerPath, input.artifactServerPort)
 
 		ctx = common.WithDryrun(ctx, input.dryrun)
 		if watch, err := cmd.Flags().GetBool("watch"); err != nil {
@@ -294,7 +294,11 @@ func newRunCommand(ctx context.Context, input *Input) func(*cobra.Command, []str
 			return watchAndRun(ctx, r.NewPlanExecutor(plan))
 		}
 
-		return r.NewPlanExecutor(plan)(ctx)
+		executor := r.NewPlanExecutor(plan).Finally(func(ctx context.Context) error {
+			cancel()
+			return nil
+		})
+		return executor(ctx)
 	}
 }
 
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -17,7 +17,7 @@ import (
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 
-	"github.com/nektos/act/pkg/assets"
+	"github.com/nektos/act/pkg/artifacts"
 	"github.com/nektos/act/pkg/common"
 	"github.com/nektos/act/pkg/model"
 	"github.com/nektos/act/pkg/runner"
@@ -67,12 +67,12 @@ func Execute(ctx context.Context, version string) {
 	rootCmd.PersistentFlags().StringVarP(&input.containerArchitecture, "container-architecture", "", "", "Architecture which should be used to run containers, e.g.: linux/amd64. If not specified, will use host default architecture. Requires Docker server API Version 1.41+. Ignored on earlier Docker server platforms.")
 	rootCmd.PersistentFlags().StringVarP(&input.containerDaemonSocket, "container-daemon-socket", "", "/var/run/docker.sock", "Path to Docker daemon socket which will be mounted to containers")
 	rootCmd.PersistentFlags().StringVarP(&input.githubInstance, "github-instance", "", "github.com", "GitHub instance to use. Don't use this if you are not using GitHub Enterprise Server.")
+	rootCmd.PersistentFlags().StringVarP(&input.artifactServerPath, "artifact-server-path", "", "", "Defines the path where the artifact server stores uploads and retrieves downloads from. If not specified the artifact server will not start.")
 	rootCmd.SetArgs(args())
 
 	// todo: where to start the server?
-	// todo: instead of env ASSET_PATH and PORT would prefer flags. how to integrate with these?
 	go func() {
-		assets.ServeAssets()
+		artifacts.Serve(input.artifactServerPath)
 	}()
 
 	if err := rootCmd.Execute(); err != nil {
@@ -281,6 +281,7 @@ func newRunCommand(ctx context.Context, input *Input) func(*cobra.Command, []str
 			ContainerCapAdd:       input.containerCapAdd,
 			ContainerCapDrop:      input.containerCapDrop,
 			AutoRemove:            input.autoRemove,
+			ArtifactServerPath:    input.artifactServerPath,
 		}
 		r, err := runner.New(config)
 		if err != nil {
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -68,12 +68,8 @@ func Execute(ctx context.Context, version string) {
 	rootCmd.PersistentFlags().StringVarP(&input.containerDaemonSocket, "container-daemon-socket", "", "/var/run/docker.sock", "Path to Docker daemon socket which will be mounted to containers")
 	rootCmd.PersistentFlags().StringVarP(&input.githubInstance, "github-instance", "", "github.com", "GitHub instance to use. Don't use this if you are not using GitHub Enterprise Server.")
 	rootCmd.PersistentFlags().StringVarP(&input.artifactServerPath, "artifact-server-path", "", "", "Defines the path where the artifact server stores uploads and retrieves downloads from. If not specified the artifact server will not start.")
+	rootCmd.PersistentFlags().StringVarP(&input.artifactServerPort, "artifact-server-port", "", "34567", "Defines the port where the artifact server listens (will only bind to localhost).")
 	rootCmd.SetArgs(args())
-
-	// todo: where to start the server?
-	go func() {
-		artifacts.Serve(input.artifactServerPath)
-	}()
 
 	if err := rootCmd.Execute(); err != nil {
 		os.Exit(1)
@@ -282,11 +278,16 @@ func newRunCommand(ctx context.Context, input *Input) func(*cobra.Command, []str
 			ContainerCapDrop:      input.containerCapDrop,
 			AutoRemove:            input.autoRemove,
 			ArtifactServerPath:    input.artifactServerPath,
+			ArtifactServerPort:    input.artifactServerPort,
 		}
 		r, err := runner.New(config)
 		if err != nil {
 			return err
 		}
+
+		go func() {
+			artifacts.Serve(input.artifactServerPath, input.artifactServerPort)
+		}()
 
 		ctx = common.WithDryrun(ctx, input.dryrun)
 		if watch, err := cmd.Flags().GetBool("watch"); err != nil {
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -285,9 +285,7 @@ func newRunCommand(ctx context.Context, input *Input) func(*cobra.Command, []str
 			return err
 		}
 
-		go func() {
-			artifacts.Serve(input.artifactServerPath, input.artifactServerPort)
-		}()
+		artifacts.Serve(ctx, input.artifactServerPath, input.artifactServerPort)
 
 		ctx = common.WithDryrun(ctx, input.dryrun)
 		if watch, err := cmd.Flags().GetBool("watch"); err != nil {
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -17,6 +17,7 @@ import (
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 
+	"github.com/nektos/act/pkg/assets"
 	"github.com/nektos/act/pkg/common"
 	"github.com/nektos/act/pkg/model"
 	"github.com/nektos/act/pkg/runner"
@@ -67,6 +68,10 @@ func Execute(ctx context.Context, version string) {
 	rootCmd.PersistentFlags().StringVarP(&input.containerDaemonSocket, "container-daemon-socket", "", "/var/run/docker.sock", "Path to Docker daemon socket which will be mounted to containers")
 	rootCmd.PersistentFlags().StringVarP(&input.githubInstance, "github-instance", "", "github.com", "GitHub instance to use. Don't use this if you are not using GitHub Enterprise Server.")
 	rootCmd.SetArgs(args())
+
+	go func() {
+		assets.AssetServer()
+	}()
 
 	if err := rootCmd.Execute(); err != nil {
 		os.Exit(1)
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -69,8 +69,10 @@ func Execute(ctx context.Context, version string) {
 	rootCmd.PersistentFlags().StringVarP(&input.githubInstance, "github-instance", "", "github.com", "GitHub instance to use. Don't use this if you are not using GitHub Enterprise Server.")
 	rootCmd.SetArgs(args())
 
+	// todo: where to start the server?
+	// todo: instead of env ASSET_PATH and PORT would prefer flags. how to integrate with these?
 	go func() {
-		assets.AssetServer()
+		assets.ServeAssets()
 	}()
 
 	if err := rootCmd.Execute(); err != nil {
--- a/pkg/artifacts/server.go
+++ b/pkg/artifacts/server.go
@@ -266,7 +266,7 @@ func Serve(ctx context.Context, artifactPath string, port string) context.Cancel
 	go func() {
 		<-serverContext.Done()
 
-		if err := server.Shutdown(serverContext); err != nil {
+		if err := server.Shutdown(ctx); err != nil {
 			log.Errorf("Failed shutdown gracefully: %v", err)
 		}
 	}()
--- a/pkg/artifacts/server.go
+++ b/pkg/artifacts/server.go
@@ -259,7 +259,9 @@ func Serve(ctx context.Context, artifactPath string, port string) context.Cancel
 	// run server
 	go func() {
 		log.Infof("Start server on http://%s:%s", ip, port)
-		log.Fatal(server.ListenAndServe())
+		if err := server.ListenAndServe(); err != nil && err != http.ErrServerClosed {
+			log.Fatal(err)
+		}
 	}()
 
 	// wait for cancel to gracefully shutdown server
--- a/pkg/artifacts/server.go
+++ b/pkg/artifacts/server.go
@@ -6,13 +6,14 @@ import (
 	"fmt"
 	"io"
 	"io/fs"
-	"log"
 	"net/http"
 	"os"
 	"path"
 	"path/filepath"
+	"strings"
 
 	"github.com/julienschmidt/httprouter"
+	log "github.com/sirupsen/logrus"
 )
 
 type FileContainerResourceURL struct {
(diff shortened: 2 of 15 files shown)


pkg/assets/server.go
```diff
@@ -0,0 +1,227 @@
...
+		}
+
+		_, err = io.Copy(w, file)
+		if err != nil {
+			panic(err)
+		}
+	})
+}
+
+func ServeAssets() {
```
catthehacker, line 203:
```suggestion
func ServeAssets(p string) {
```

pkg/assets/server.go
```diff
@@ -0,0 +1,227 @@
...
+	// todo: how to name this env var?
+	assetPath := os.Getenv("ASSET_PATH")
+	if assetPath == "" {
+		absPath, err := filepath.Abs("./.act")
+		if err != nil {
+			panic(err)
+		}
+
+		assetPath = absPath
+	}
```
catthehacker, line 215:
```suggestion
	if p == "" {
		p = os.TempDir()
	}
```

catthehacker: or
`$XDG_DATA_HOME`
https://specifications.freedesktop.org/basedir-spec/latest/ar01s03.html

cmd/root.go
```diff
@@ -64,6 +65,12 @@ func Execute(ctx context.Context, version string) {
 	rootCmd.PersistentFlags().StringVarP(&input.githubInstance, "github-instance", "", "github.com", "GitHub instance to use. Don't use this if you are not using GitHub Enterprise Server.")
 	rootCmd.SetArgs(args())
 
+	// todo: where to start the server?
+	// todo: instead of env ASSET_PATH and PORT would prefer flags. how to integrate with these?
+	go func() {
+		assets.ServeAssets()
+	}()
```
catthehacker, line 72:
```suggestion
	go func() {
		assets.ServeAssets(input.ArtifactServerPath)
	}()
```

catthehacker: + add flag 🚩 in `Execute()`

catthehacker (changes requested):

KnisterPeter:
Let me explain a bit on why we thing that configuration might help here (for our usecase):

This assets could be used for caching things across runs as well. In fact the github `@actions/cache` is also doing this (but unfortunatelly not availabel for GHE). That wouldn't work out of the box and is probably not even target with act.

We are using act to run github actions in a non-local environment. And we would like to use this feature (and extend on that outside of act) to build a whole caching layer.
To do that we need to provide different configuration (at least for the URL part).

An implicit overridable env-var would be fine for us.

catthehacker:
Ok, I see server is running locally on host. That won't work if the Docker host is a remote server, for example `ssh://`

pkg/assets/server.go
KnisterPeter: We might have another problem here.
The artifacts would be stored across act executions (which is fine) but the workflowId of act is always the 1.
That means we would override and add files to that folder on consecutive executions.
Using a folder local to the project makes this at least discoverable to the user. I like your idea of using the os tempdir, but we should then scope it with a run id.

catthehacker:
> We are using act to run github actions in a non-local environment. And we would like to use this feature (and extend on that outside of act) to build a whole caching layer.
> To do that we need to provide different configuration (at least for the URL part).
> 
> An implicit overridable env-var would be fine for us.

We should at least follow some standard of non-conflicting envvars, like `ACT_*`

KnisterPeter:
> Even better if it's internal network for containers only.

What do you mean with that? Do you suggest to make the server only available in the docker internal network for the run - for sharing across jobs?

One use case which is in my head would be to expose test result data or coverage information after an act execution.
If that's internal to the container users cannot do that.

catthehacker:
IMO, *server* should be available only in containers created by `act`, artefacts can be accessible via specified path.

catthehacker:
Don't worry, proceed with your PR as you planned 😺 I have some idea but that requires much more changes in `act`

catthehacker:
> ## Missing things
> * [x]  CLI flags for configuration (port, asset path)

since server will be on host for now, please add port flag

> * [x]  `--env ACTIONS_RUNTIME_URL=http://localhost:8080/` must be set because it's internally used by the `@action/toolkit`
> * [x]  `--env ACTIONS_RUNTIME_TOKEN=foo` must be set (we do not have authentication implemented)

Please add them to GitHub context struct and set them in `withGithubEnv()`
https://github.com/nektos/act/blob/cc4e23d96c24effff203c86d679c22c9ab092f27/pkg/runner/run_context.go#L457
https://github.com/nektos/act/blob/cc4e23d96c24effff203c86d679c22c9ab092f27/pkg/runner/run_context.go#L652

KnisterPeter:
@catthehacker @cplee Can you guide us where the proper place for these lines are? 

https://github.com/nektos/act/pull/677/files#diff-ab967ab1a2f3a1b769106eeb7bfe892ef0e81d1d27811fa15be08e6749feee1fR69-R72

I think it might not be right in the middle of the CLI parsing.

pkg/artifacts/server.go
```diff
@@ -0,0 +1,209 @@
...
+		})
+		if err != nil {
+			panic(err)
+		}
+
+		_, err = w.Write(json)
+		if err != nil {
+			panic(err)
+		}
+	})
```
ChristopherHX, line 95: Does this work for asset files > 8 MB?
I ask this because the actions/toolkit uploads big files in 8MB chunks with range upload.
Currently 2 chunks are uploaded parallel to the server by the toolkit.
_I haven't tested this implementation, but the nodejs version also missed chunk upload_

KnisterPeter: I guess it doesn't work. I think think the default go web server does this.
But I also think the support for chunked upload could be defered to a later PR. The API wouldn't change for that.
Splitting this into multiple PRs would add support for smaller chunks earlier/faster.

catthehacker:
Could you also update `README` with added flags?
`go run main.go --help` and copy paste flags

KnisterPeter:
@catthehacker Yes, good point. I'm currently writing tests. So this is still WIP.

KnisterPeter:
@catthehacker I'm very inexperienced with go. I think the tests are fine now, can you do a review. Specially with the MemFS parts, I'm unsure if this is 'go-like'.

**P.S.:** This is still WIP. I need to re-test if the implementation still works.

crawler:
Hi, @KnisterPeter
Thank you for doing the really useful feature. Once I tried this awesome project I immediately noticed that it missing cache and every run makes the dependencies downloaded every time. So despite I not common with GO I installed your's fork and tried different options with no luck. With the lsof command I see that the server is started, and I can connect to it using 'curl' but the getCacheEntry can't connect to it:

`act -s GITHUB_TOKEN=$GITHUB_TOKEN -j test --artifact-server-path=cache`

results to 

`::debug::getCacheEntry - Attempt 2 of 2 failed with error: connect ECONNREFUSED 127.0.0.1:34567`

same thing with the ACTIONS_RUNTIME_URL

`act -s GITHUB_TOKEN=$GITHUB_TOKEN -j test --artifact-server-path=cache --artifact-server-port=7676 --env ACTIONS_RUNTIME_URL=http://spitfire.lan:7676`

results to

`::debug::getCacheEntry - Attempt 2 of 2 failed with error: connect ECONNREFUSED 127.0.0.1:7676`

spitfire.lan - is local domain name of the my workstation

Will appreciate any information)

KnisterPeter:
@crawler 

> missing cache

First of all I have to say that this is not about the cache, but about artifacts created during the build. The github actions caching is only using parts of this and is not compatible.

Second to note, this is not a fully tested solution and it should be integrated when I'm ready with testing. I haven't had so much time lately.

crawler:
@KnisterPeter Thank you for having time to clear this out for me.

catthehacker:
@crawler Use https://github.com/nektos/act#skipping-steps with `-r, --reuse | reuse action containers to maintain state`

KnisterPeter:
@catthehacker From our perspective this is finished now.
There is a last step: we tried to add an [integration test](https://github.com/xing/act/blob/asset-server/pkg/artifacts/server_test.go#L218-L370). It creates a workflow file. From our understanding of the [Makefile](https://github.com/xing/act/blob/asset-server/Makefile#L32), act should be executed right after the `go test` run. But it isn't.
Can you help us with this missing part?

catthehacker:
`test-*.yml` are ignored, you should make it a standalone workflow file in `testdata/`
https://github.com/nektos/act/blob/8a9167da82b3824272fdbfcc25b3671cd1260e7d/.github/workflows/.gitignore#L1

ChristopherHX:
I think you need to add your test workflow into this list, as far as I understand the project structure
https://github.com/nektos/act/blob/8a9167da82b3824272fdbfcc25b3671cd1260e7d/pkg/runner/runner_test.go#L87-L115

_I didn't noticed that the artifact client uses gzip in some cases, good catch_

KnisterPeter:
@ChristopherHX I think that wouldn't be sufficient.
First of all, that testdata list belongs to the `runner` package.
Second, we need to start act with the artifact server, which is currently in the `root.go`. I'm already refactoring this, so we can create the server in the tests as well, and then run act based on that.
It will look similar to the runner tests.

KnisterPeter:
@catthehacker @ChristopherHX @cplee 
I'm kind of stuck verifying that this works on mac os. The test is running in a docker-machine/boot2docker and the actions cannot access `localhost` (I think because `localhost` is not the host where act was started and the webserver lives).
I tried several aproaches to determine the hosts real IP address, but none of my tries succeeds.

Any help or idea would be nice. Otherwise I don't see a solution to this (not even have a mac to test this locally which is kind of hard to debug).

catthehacker:
This should help you
https://github.com/nektos/act/blob/8a9167da82b3824272fdbfcc25b3671cd1260e7d/.github/workflows/checks.yml#L94-L99
`DOCKER_HOST` contains IP of machine running Docker

catthehacker:
But that is what I mentioned earlier, that artefact server will not work in scenario where Docker is accessed through SSH since you will have to do gimmicks to make it connect to host from VM.
Feel free to lock out artefacts server to `linux`-only GOOS.
Also when you operate in Docker, you shouldn't use `localhost` but `host.docker.internal`

KnisterPeter:
For production access we have the ACTION_RUNTIME_URL inside the container.

ChristopherHX:
@KnisterPeter Do it similar to https://github.com/ChristopherHX/act/commit/b647b6e594270bb74666fc39608c0a81dbc3dac3, then it works on macOS.

But your tests still fail both on linux / mac github for me.
```
2021-06-04T16:24:24.9828060Z time="2021-06-04T16:24:11Z" level=debug msg="context env => map[ACT:true ACTIONS_RUNTIME_TOKEN:token ACTIONS_RUNTIME_URL:http://10.79.0.165:12345/ CI:true GITHUB_ACTION:18 GITHUB_ACTIONS:true GITHUB_ACTION_REF: GITHUB_ACTION_REPOSITORY: GITHUB_ACTOR:nektos/act GITHUB_API_URL:https://api.github.com GITHUB_BASE_REF: GITHUB_ENV:/var/run/act/workflow/envs.txt GITHUB_EVENT_NAME:push GITHUB_EVENT_PATH:/var/run/act/workflow/event.json GITHUB_GRAPHQL_URL:https://api.github.com/graphql GITHUB_HEAD_REF: GITHUB_JOB:test-artifacts GITHUB_PATH:/var/run/act/workflow/paths.txt GITHUB_REF:refs/heads/act-cache GITHUB_REPOSITORY:ChristopherHX/act GITHUB_REPOSITORY_OWNER:ChristopherHX GITHUB_RETENTION_DAYS:0 GITHUB_RUN_ID:1 GITHUB_RUN_NUMBER:1 GITHUB_SERVER_URL:https://github.com GITHUB_SHA:b647b6e594270bb74666fc39608c0a81dbc3dac3 GITHUB_TOKEN: GITHUB_WORKFLOW:Test that artifact uploads and downloads succeed GITHUB_WORKSPACE:/Users/runner/work/act/act/pkg/artifacts/testdata ImageOS:ubuntu20 PATH:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin RUNNER_PERFLOG:/dev/null RUNNER_TRACKING_ID:]"
2021-06-04T16:24:24.9837040Z [Test that artifact uploads and downloads succeed/test-artifacts] setupEnv => map[ACT:true ACTIONS_RUNTIME_TOKEN:token ACTIONS_RUNTIME_URL:http://10.79.0.165:12345/ CI:true GITHUB_ACTION:18 GITHUB_ACTIONS:true GITHUB_ACTION_REF: GITHUB_ACTION_REPOSITORY: GITHUB_ACTOR:nektos/act GITHUB_API_URL:https://api.github.com GITHUB_BASE_REF: GITHUB_ENV:/var/run/act/workflow/envs.txt GITHUB_EVENT_NAME:push GITHUB_EVENT_PATH:/var/run/act/workflow/event.json GITHUB_GRAPHQL_URL:https://api.github.com/graphql GITHUB_HEAD_REF: GITHUB_JOB:test-artifacts GITHUB_PATH:/var/run/act/workflow/paths.txt GITHUB_REF:refs/heads/act-cache GITHUB_REPOSITORY:ChristopherHX/act GITHUB_REPOSITORY_OWNER:ChristopherHX GITHUB_RETENTION_DAYS:0 GITHUB_RUN_ID:1 GITHUB_RUN_NUMBER:1 GITHUB_SERVER_URL:https://github.com GITHUB_SHA:b647b6e594270bb74666fc39608c0a81dbc3dac3 GITHUB_TOKEN: GITHUB_WORKFLOW:Test that artifact uploads and downloads succeed GITHUB_WORKSPACE:/Users/runner/work/act/act/pkg/artifacts/testdata ImageOS:ubuntu20 PATH:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin RUNNER_PERFLOG:/dev/null RUNNER_TRACKING_ID:]
2021-06-04T16:24:24.9842530Z [Test that artifact uploads and downloads succeed/test-artifacts] ⭐  Run Verify Artifact #4
2021-06-04T16:24:24.9845770Z time="2021-06-04T16:24:11Z" level=debug msg="Wrote command '\nfile1=\"multi/artifact/dir-1/file1.txt\"\nfile2=\"multi/artifact/dir-2/file2.txt\"\nif [ ! -f $file1 -o ! -f $file2 ] ; then\n  echo \"Expected files do not exist\"\n  exit 1\nfi\nif [ \"$(cat $file1)\" != \"Lorem ipsum dolor sit amet\" -o \"$(cat $file2)\" != \"Hello world from file #2\" ] ; then\n  echo \"File contents of downloaded artifacts are incorrect\"\n  exit 1\nfi\n\n' to 'workflow/18'"
2021-06-04T16:24:24.9848190Z time="2021-06-04T16:24:11Z" level=debug msg="Writing entry to tarball workflow/18 len:352"
2021-06-04T16:24:24.9849810Z [Test that artifact uploads and downloads succeed/test-artifacts] Extracting content to '/Users/runner/work/act/act/pkg/artifacts/testdata'
2021-06-04T16:24:24.9853020Z [Test that artifact uploads and downloads succeed/test-artifacts]   🐳  docker exec cmd=[bash --noprofile --norc -e -o pipefail /Users/runner/work/act/act/pkg/artifacts/testdata/workflow/18] user=
2021-06-04T16:24:24.9855630Z [Test that artifact uploads and downloads succeed/test-artifacts] Exec command '[bash --noprofile --norc -e -o pipefail /Users/runner/work/act/act/pkg/artifacts/testdata/workflow/18]'
2021-06-04T16:24:24.9857430Z [Test that artifact uploads and downloads succeed/test-artifacts]   ✅  Success - Verify Artifact #4
2021-06-04T16:24:24.9859710Z [Test that artifact uploads and downloads succeed/test-artifacts] Removed container: c8e3d0ed64d9312c891f284e67ba00974d62eda3a7ada4fbe26fa7fd267b038a
2021-06-04T16:24:24.9862760Z [Test that artifact uploads and downloads succeed/test-artifacts]   🐳  docker volume rm act-Test-that-artifact-uploads-and-downloads-succeed-test-artifacts
2021-06-04T16:24:24.9865780Z --- PASS: TestArtifactFlow (78.82s)
2021-06-04T16:24:24.9867290Z     --- PASS: TestArtifactFlow/upload-and-download (78.82s)
2021-06-04T16:24:24.9867850Z PASS
2021-06-04T16:24:24.9868890Z time="2021-06-04T16:24:24Z" level=error msg="Failed shutdown gracefully: context canceled"
2021-06-04T16:24:24.9870000Z time="2021-06-04T16:24:24Z" level=fatal msg="http: Server closed"
2021-06-04T16:24:24.9870660Z FAIL	github.com/nektos/act/pkg/artifacts	79.024s
```

KnisterPeter:
@ChristopherHX Interesting solution. But I think it isn't good to depend on external services. 8.8.8.8 is a DNS managed by Google. If that's down, act will not work anymore.

ChristopherHX:
No it doesn't connect to google (udp is connectionless), read the linked stackoverflow answer.
You can still use localhost as fallback

KnisterPeter:
Not stale

KnisterPeter:
@catthehacker @cplee We consider this ready for review and merge now.

- We implemented the check for the outbound ip (for macos support)
- We shutdown the server gracefully if act is done
- We rebased and resolved all conflicts

catthehacker (approved):
Looks sound 😸

KnisterPeter:
@catthehacker @cplee What need to be done to enable the merge? I wanted to tick the `allow maintainers to edit` box, but it isn't there.

catthehacker:
I assume @cplee can merge it manually (since I don't have permissions to do so) or you could rebase on master and force push branch, but it would require approval again.

cplee:
@KnisterPeter - the issue is the base branch `xing:asset-server` isnt allowing updates. I'll force merge though

jonodrew:
I can't see this documented in the README for this project - is it ready to use?

KnisterPeter:
It is documented as 'artifact server' in the readme.
And yes, it's ready to use.

debuggerpk:
Thankyou for your great work team. Having said that, I am unable to run this. The command I am using is

`act --secret-file .env.act --artifact-server-path=artifacts`

The error I am getting is

```
 💬  ::debug::Artifact Url: http://10.5.0.2:34567/_apis/pipelines/workflows/1/artifacts?api-version=6.0-preview
| Create Artifact Container - Attempt 1 of 5 failed with error: connect ECONNREFUSED 10.5.0.2:34567
| Create Artifact Container - Attempt 2 of 5 failed with error: connect ECONNREFUSED 10.5.0.2:34567
| Create Artifact Container - Attempt 3 of 5 failed with error: connect ECONNREFUSED 10.5.0.2:34567
```

Am I missing something?

PR #680, opened 2021-05-08
Hotfix: Skip Checkout Regression

JustinGrote:
The testing logic for isLocalCheckout was incorrect, fix and add a test specifically for checkout.

### DIFF
--- a/pkg/runner/runner_test.go
+++ b/pkg/runner/runner_test.go
@@ -89,6 +89,7 @@ func TestRunEvent(t *testing.T) {
 		{"testdata", "basic", "push", "", platforms, ""},
 		{"testdata", "fail", "push", "exit with `FAILURE`: 1", platforms, ""},
 		{"testdata", "runs-on", "push", "", platforms, ""},
+		{"testdata", "checkout", "push", "", platforms, ""},
 		// Pwsh is not available in default worker (yet) so we use a separate image for testing
 		{"testdata", "powershell", "push", "", map[string]string{"ubuntu-latest": "ghcr.io/justingrote/act-pwsh:latest"}, ""},
 		{"testdata", "job-container", "push", "", platforms, ""},
--- a/pkg/runner/run_context.go
+++ b/pkg/runner/run_context.go
@@ -579,7 +579,7 @@ func (rc *RunContext) getGithubContext() *githubContext {
 }
 
 func (ghc *githubContext) isLocalCheckout(step *model.Step) bool {
-	if step.Type() != model.StepTypeInvalid {
+	if step.Type() == model.StepTypeInvalid {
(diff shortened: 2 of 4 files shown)


catthehacker:
Can we remove `.secrets` and `.env` loading from tests as well?

JustinGrote:
> Can we remove `.secrets` and `.env` loading from tests as well?

does that need to be done in the context of this PR or should it be done separately? I should have just fixed the last test and this should "go green" on this commit.

JustinGrote:
> @JustinGrote this pull request has failed checks 🛠
This is probably related to why I put the invalid step check in the first place, let me nail it down.

pkg/runner/runner_test.go
```diff
@@ -89,6 +89,7 @@ func TestRunEvent(t *testing.T) {
 		{"testdata", "basic", "push", "", platforms, ""},
 		{"testdata", "fail", "push", "exit with `FAILURE`: 1", platforms, ""},
 		{"testdata", "runs-on", "push", "", platforms, ""},
+		{"testdata", "checkout", "push", "", platforms, ""},
```
catthehacker, line 92: do we really need that?
image

catthehacker (changes requested):

catthehacker:
> > Can we remove `.secrets` and `.env` loading from tests as well?
> 
> does that need to be done in the context of this PR or should it be done separately? I should have just fixed the last test and this should "go green" on this commit.

It was added because of that specific issue so I would prefer it to be gone otherwise it will mess with testing.

JustinGrote:
Sure, I can regress that as well, but it will still show up if you, for instance, want to check out a specific ref as part of your flow.

catthehacker:
> Sure, I can regress that as well, but it will still show up if you, for instance, want to check out a specific ref as part of your flow.

Sounds like we need more test cases 😼

JustinGrote:
> > Sure, I can regress that as well, but it will still show up if you, for instance, want to check out a specific ref as part of your flow.
> 
> Sounds like we need more test cases 😼

Probably :)

The current kernel panic is in the empty workflow case, it's generating a nil remote action that when it hits IsCheckout() causes a nil reference panic. It looks like the old flow was to get past this and drop an error later, sorting that out.

pkg/runner/runner_test.go
JustinGrote: I think it's important for it to be specifically called out as opposed to being a side-effect of other tests. It's like saying do we really need basic because the other commands run shell scripts too?

JustinGrote:
@catthehacker fixed and the .secrets regressed as well, LGTM

cplee (approved):

ayozemr:
Thanks for this!! Will this be part of a release anytime soon?

PR #687, opened 2021-05-17
fix: use auth when fetching (same as with cloning)

KnisterPeter:


KnisterPeter:
@cplee @catthehacker This is a regression from 0.2.21 where it did work.

cplee (approved):

PR #752, opened 2021-07-16
Fix format function {{, }} are escapes

ChristopherHX:
Github transforms multiple expressions into one format expression while parsing and every non expression `{` is replaced by `{{` and `}` by `}}`.

I assume vmformat should mimic githubs behavior, issues appeard during testing of https://github.com/ChristopherHX/github-actions-act-runner.

Haven't searched any documentation about this yet.

Some corner cases fixed too
`format('{0}', github.undefined_value)` should be an empty string not `undefined`.
`format('{0}', '{1}', 'Hello')` should be `{1}` not `Hello`.

How do I test the error handler?
vmFormat has no error output, so the tests cannot test for error messages or did I miss something?

### DIFF
--- a/pkg/runner/expression.go
+++ b/pkg/runner/expression.go
@@ -4,11 +4,11 @@ import (
 	"crypto/sha256"
 	"encoding/hex"
 	"encoding/json"
-	"fmt"
 	"io"
 	"os"
 	"path/filepath"
 	"regexp"
+	"strconv"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


catthehacker:
> vmFormat has no error output, so the tests cannot test for error messages [...]?

correct

catthehacker (approved):

cplee (approved):

PR #768, opened 2021-08-08
Cache dockerhub images from linux for macOS

ChristopherHX:
An attempt to avoid hitting the dockerhub Pull Rate Limit.

There is a small overhead due to caching (ca. +2 min, next run within 7 days only some seconds), but it might avoid rerunning the macOS job due to this error.
Restoreing the cache doesn't take much extra job execution time, because the test step is pulling less images from dockerhub.

This PR failed due to this limit #776.
It seems to be triggered by multiple PR's are opened within some days.

_Needs more testing, if this solves the dockerhub limit problems or only delays the problem to happen less frequently._

Add new dockerhub images to `CACHED_DOCKER_IMAGES` and increment the cache key (`CACHED_DOCKER_IMAGES_KEY`)

What do you think?

_I need  ACT_OWNER: ${{ github.repository_owner }} and ACT_REPOSITORY: ${{ github.repository }} to have all tests green within my fork_


catthehacker (dismissed):
My plans are to escape DockerHub and move everything to other registries but this will suffice for now.

cplee (dismissed):

ChristopherHX:
**This PR doesn't need work, it only needs to be merged manually**
However mergify added this label weeks later, once another PR was merged

cplee (approved):

Issue #757, opened 2021-07-20
Environment variables defined in the container are not evaluatable


mulle-nat:
## System information

- Operating System: macOS
- Architecture: x86_64

- Docker version: Docker version 20.10.7, build f0df350
- Docker image used in `act`: catthehacker/ubuntu:act-20.04
- `act` version: 0.2.23

## Expected behaviour

I expected to be able to use container defined environment variables in if: statements.

```
name: CI

on: [push]

env:
# ENV2: # undefined here, but defined in local container 

jobs:
  build:
    runs-on: ${{ matrix.os }}
    strategy:
      matrix:
        os: [ ubuntu-latest] # macos-latest,

    steps:
    - name: Dump Environment
      run:  env | sort

    - name: Test
      if: ${{ env.ENV2 }} 
      run:  echo Test
```

## Actual behaviour

`ENV2` value is ignored, unless defined in workflow file.

## Workflow and/or repository

See the complete workflow with step by step instructions on how to reproduce:

https://github.com/mulle-nat/act-and-github

</details>

catthehacker:
We don't inspect Docker image for envvars, they should be in `/etc/environment`

PR #771, opened 2021-08-09
Add reading environment from Docker image

catthehacker:
This change will add reading image definition and merging it's environment variables with our current env in style of:
- missing envvar, creates new one
- existing envvar is overwritten by our env
- `PATH` envvar is merged, container `PATH` is last to not overwrite any tools installed during workflow

It's not required for running our main container but it will be probably more intuitive if it will work like standard docker and this change is required for nested containers started in workflow

Fixes https://github.com/nektos/act/issues/757


joonas:
Glad to see this PR, I just recently came across an instance myself where the `golang:1.17` container sets up a `PATH` variable, which unfortunately currently [gets overridden](https://github.com/nektos/act/blob/de44a505da37f9405db195902b277821df2742b6/pkg/runner/step_context.go#L145-L147) unless an explicit `PATH` environment variable is set.

In turn, this results in the `go` not being available in the image.

After spelunking a bit into the `actions/runner` codebase, I found that they are [parsing it out](https://github.com/actions/runner/blob/main/src/Runner.Worker/Container/DockerUtil.cs#L35-L47) from the [`docker inspect --format "{{range .Config.Env}}{{println .}}{{end}}" ...` output](https://github.com/actions/runner/blob/f61dcad5bb126b028385764a87ef121e7416d5e5/src/Runner.Worker/ContainerOperationProvider.cs#L321-L324).

cplee (approved):

PR #773, opened 2021-08-09
fix: log in to dockerhub to avoid rate limits

catthehacker:
not required on linux runners since they have accounts
provided by GitHub


cplee (approved):

catthehacker:
I'm so confused right now

catthehacker:
This won't work for now.

PR #781, opened 2021-08-20
CI: Add CodeQL analysis for discovering code vulnerabilities

HarshCasper:
## Description

This PR adds CodeQL analysis for discovering code vulnerabilities.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,31 @@
+name: "CodeQL analysis"
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+  schedule:
+    - cron: '30 12 * * 1'
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+      matrix:
+        language: [ 'go' ]
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v2
+
+    - name: Initialize CodeQL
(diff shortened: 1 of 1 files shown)


catthehacker (approved):
I don't see how that would benefit us or any other reason to have security scanning

cplee:
@HarshCasper - where do I see the results?

catthehacker:
@cplee https://github.com/nektos/act/pull/781/checks?check_run_id=3384696784#step:4:3359

catthehacker:
All results will be also in https://github.com/nektos/act/security/code-scanning

cplee:
@catthehacker - ok, i see it (and screenshot below)...but is this good or bad? I see errors but shouldn't this fail the build if there are errors?

I'm all for security scanning but I think it should be enforced through failed builds rather than buried in the build output...but I may be misunderstanding how it works

image

catthehacker:
From what I've read about CodeQL action it's not made to fail in any circumstance and even `codeql-action` succeeds with errors :)
image

cplee:
Sorry, dumb question then - what value does this provide?

catthehacker:
I assume it should present us with some more security errors in checks or show some issues in https://github.com/nektos/act/security but I don't know if it's just not configured or weak.
There are some configs mentioned in https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/configuring-code-scanning#running-additional-queries
but I never bothered to set up CodeQL so don't know much about it.

As I said earlier, I don't see need for this due to the way `act` works, it executes everything on demand in Docker, if someone were to try something malicious they would go for the actions instead of `act`.

PR #790, opened 2021-08-30
Add Act Path

ChristopherHX:
Please make actpath relocatable, I need this for my hostexecutor mode. I'm working on merging my whole fork back to nektos/act.
On windows `/var/run/act` cannot even be created, this make it more dynamic.

Also this moves run scripts to actpath away from your workspace.

### DIFF
--- a/pkg/runner/run_context.go
+++ b/pkg/runner/run_context.go
@@ -19,8 +19,6 @@ import (
 	"github.com/nektos/act/pkg/model"
 )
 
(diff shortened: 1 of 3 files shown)


catthehacker:
I don't doubt it is being used but I would prefer that the change to code would come with something that improves act itself otherwise change can be kept in fork/repo, as I would like to minimise chance of act breaking (especially since that path is very sensitive and I broke it myself once or twice).

Issue #158, opened 2020-03-16
Possible improvements on parallelism


aidansteele:
## Background

I'll start with an example from one of my projects. Here's the graph rendered by `act -l`:

```
 ╭──────────╮ ╭─────────╮ ╭───────╮
 │ frontend │ │ backend │ │ certs │
 ╰──────────╯ ╰─────────╯ ╰───────╯
                 ⬇
   ╭────────────────╮ ╭────────╮
   │ deployfrontend │ │ deploy │
   ╰────────────────╯ ╰────────╯
```

In this case, `deployfrontend` needs `frontend` and `deploy` needs `backend`. 

## My understanding of the current state

Right now, it appears that `act` forms a dependency graph of workflows, jobs, steps, matrixes, etc. These are then transformed into `model.Plan`, `model.Stage` and `model.Run` structs. A `model.Run` is effectively an instance of a job -- it isn't a 1:1 mapping because matrixes mean that there could be from 0-n instances of the same job, each with different variables.

All runs within a stage are run in parallel and a stage is only complete when all runs have completed. At this point, execution can proceed to the next stage. 

For stage N (where N is the zero-based stage to run in order) stage to run, all runs have a list of transitive dependencies N long. So runs with no dependencies are all in stage 0, runs with 1 dependency are in stage 1, etc.

Is this correct?

## Proposal / discussion

Right now `deployfrontend` won't run until `frontend`, `backend` and `certs` have all completed. In my case, `frontend` completes *much* faster than `backend`. I would love for `deployfrontend` to start execution as soon as `frontend` has completed, even if `backend` is still executing.

I think this would mean eliminating the concept of stages entirely. Can I ask why they were introduced?  Without diving into it in *real* depth, it kind of feels like the existing abstractions around `common.Executor` would be sufficient to model the DAG necessary to represent execution, so I'm not sure why `model.Stage` is there. I feel like I can't confidently suggest eliminating stages without understanding what problem they are currently solving.

One other random point: maybe it's not an issue in practice, but I feel like maybe it would be useful also limiting the number of actively executing runs to the number of CPUs present. Thoughts on that?

I'd be happy to submit a PR or at least ideas for a PR proposal if you are happy with the idea of refactoring to allow runs to start independently of each other, but I feel I should get some feedback first.

aidansteele:
Sorry for the verbosity 🤯

cplee:
@aidansteele - sorry for the delay, love the verbosity.

[...]

PR #823, opened 2021-09-24
fix: add simple concurrency limit

catthehacker:
Adds concurrency limit based on `max-parallel` key from `strategy`
In case of missing key, it uses default value `4`
If there is less jobs than default max parallel, it will set max parallel to count of jobs.

Currently it will run max parallel jobs, wait for finish, then start another batch of jobs.

Related issue: https://github.com/nektos/act/issues/158

TODO: `job.Strategy` should always be available even if it's not specified in workflow, so then we won't have to redefine `maxParallel` limit to `4`


catthehacker:
macos failed again when setting up docker

cplee (approved):

ZauberNerd:
@catthehacker @cplee this seems to introduce a regression as now unrelated jobs aren't run in parallel anymore.
From the code it looks like only matrix jobs are still running in parallel.

Example workflow:
```yaml
name: CI

on: [push]

jobs:
  build:
    runs-on: ubuntu-latest
    steps:
      - run: echo "build"

  setup:
    runs-on: ubuntu-latest
    steps:
      - run: echo "setup"

  test:
    runs-on: ubuntu-latest
    steps:
      - run: echo "test"
```

catthehacker:
imdumb

PR #857, opened 2021-11-02
lint: make shellcheck happier

jsoref:
I'm not a huge fan of linters, but when I tried to run `make`, I tripped over a bunch of things. After a bit of thought, I eventually figured out some of them...

Sprinkling some quotation marks here seems like a reasonable compromise.

Note: I can't figure out how to make the linter happy about: `for image in ${{ env.CACHED_DOCKER_IMAGES }}` -- it is confident that the thing after `in` is a single value and not possibly multiple values. I suspect it's possible to trick it by writing that item to a file and then using `cat` to retrieve the contents of the file, but that seems quite a bit much.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -82,7 +82,7 @@ jobs:
           docker pull registry:2
           docker image save -o registry registry:2
           mkdir -p docker-registry
-          docker run -d -p 5000:5000 --name registry -v $PWD/docker-registry:/var/lib/registry registry:2
+          docker run -d -p 5000:5000 --name registry -v "$PWD/docker-registry:/var/lib/registry" registry:2
           npx wait-on tcp:5000
           for image in ${{ env.CACHED_DOCKER_IMAGES }}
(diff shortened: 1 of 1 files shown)


catthehacker:
shell files shouldn't be validated at all
https://github.com/nektos/act/blob/6a3a47c2175c5f6438370d28febf8933097cdc55/Makefile#L55-L66

jsoref:
🤷 , I think that would be `.sh` files directly by shellcheck, as opposed to actionlint calling shellcheck...

https://github.com/rhysd/actionlint/blob/0b7375279d2caf63203701eccc39ab091cc83a50/docs/usage.md#ignore-some-errors
I think you'd need an actionlint.yml file, but, I've never met this before, so I'm not entirely certain.

catthehacker:
I want to replace the linter but didn't had enough time to do so yet

cplee (approved):

jsoref:
This was fixed by #860

Issue #858, opened 2021-11-02
> macOS CI broken VirtualBox >= 6.1.28


ChristopherHX:
> Error creating machine: Error in driver during machine creation: Error setting up host only network on machine start: /usr/local/bin/VBoxManage hostonlyif ipconfig vboxnet0 --ip 192.168.99.1 --netmask 255.255.255.0 failed:
VBoxManage: error: Code E_ACCESSDENIED (0x80070005) - Access denied (extended info not available)
VBoxManage: error: Context: "EnableStaticIPConfig(Bstr(pszIp).raw(), Bstr(pszNetmask).raw())" at line 242 of file VBoxManageHostonly.cpp

😕

_Originally posted by @jsoref in https://github.com/nektos/act/issues/856#issuecomment-957240793_

Solution change the docker-machine create command to `docker-machine create --driver virtualbox --virtualbox-boot2docker-url ${{ env.ISO_PATH }} --virtualbox-hostonly-cidr "192.168.56.1/24" default`

Sources
https://github.com/kubernetes/minikube/pull/12811
https://stackoverflow.com/questions/48500031/what-does-virtualbox-hostonly-cidr-do-during-docker-machine-creation

ChristopherHX:
No longer needed, macOS CI is removed.

PR #860, opened 2021-11-02
ci: remove macOS tests

catthehacker:
Running Docker on non-Linux platforms is very painful,
especially in CI. It's better to remove macOS for now,
until a better implementation is present.

Signed-off-by: hackercat <me@hackerc.at>

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -4,8 +4,6 @@ on: [pull_request, workflow_dispatch]
 env:
   ACT_OWNER: ${{ github.repository_owner }}
   ACT_REPOSITORY: ${{ github.repository }}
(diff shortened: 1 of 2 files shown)


ChristopherHX:
I agree that removing macOS ci for now is a good idea.

> until a better implementation is present

If you only mean the macOS failures since today, then you find a simple solution in this issue #858. It is only one new cli argument and it works as before.

catthehacker:
> If you only mean the macOS failures since today,

I mean as a whole, since we don't actually need to run Docker on macOS, same goes for Windows.
Runners can just interact with Docker host over SSH. unfortunately anything involves using secrets which don't work in PRs.

cplee (approved):

ChristopherHX:
Did you forget to remove the macOS ci from your branch protection rules? I still see:

> test-macos Expected — Waiting for status to be reported

However repo admins can ignore branch protection rules.

PR #887, opened 2021-11-17
Add more steps context support

victorpopkov:
Add more steps context support to match GitHub:
https://docs.github.com/en/actions/learn-github-actions/contexts#steps-context

Should reference https://github.com/nektos/act/issues/465 and https://github.com/nektos/act/pull/557.

#### Add step outcome

`steps.<step id>.outcome` string is the result of a completed step **before** `continue-on-error`.

#### Remove step success in favour of conclusion

`steps.<step id>.conclusion` string is the result of a completed step **after** `continue-on-error` is applied, which currently matches the original `stepResult.Success` value. So, respectively, all checks have been replaced:

- `Success == true` &rarr; `Conclusion == stepStatusSuccess`
- `Success == false` &rarr; `Conclusion == stepStatusFailure`

### Tested on

- Linux (Ubuntu 20.04.3 LTS x86_64)

All `runner` package tests have passed. The following workflow has been used to check the behaviour as well:

```yml
name: test
on: push

jobs:
  action-check:
    runs-on: ubuntu-latest
    steps:
      - name: First
        id: first
        run: exit 0
      - name: Second
        continue-on-error: true
        id: second
        run: exit 1
      - run: echo '${{ toJSON(steps.first) }}'
      - run: echo '${{ toJSON(steps.second) }}'
```

<details>
<summary><b>Output (before)</b></summary>

```txt
[test/action-check] 🚀  Start image=node:12-buster-slim
[test/action-check]   🐳  docker pull image=node:12-buster-slim platform= username= forcePull=false
[test/action-check]   🐳  docker create image=node:12-buster-slim platform= entrypoint=["/usr/bin/tail" "-f" "/dev/null"] cmd=[]
[test/action-check]   🐳  docker run image=node:12-buster-slim platform= entrypoint=["/usr/bin/tail" "-f" "/dev/null"] cmd=[]
[test/action-check]   🐳  docker exec cmd=[mkdir -m 0777 -p /var/run/act] user=root workdir=
[test/action-check] ⭐  Run First
[test/action-check]   🐳  docker exec cmd=[bash --noprofile --norc -e -o pipefail /home/victor/Forks/act/workflow/first] user= workdir=
[test/action-check]   ✅  Success - First
[test/action-check] ⭐  Run Second
[test/action-check]   🐳  docker exec cmd=[bash --noprofile --norc -e -o pipefail /home/victor/Forks/act/workflow/second] user= workdir=
[test/action-check]   ❌  Failure - Second
[test/action-check] Failed but continue next step
[test/action-check] ⭐  Run echo '${{ toJSON(steps.first) }}'
[test/action-check]   🐳  docker exec cmd=[bash --noprofile --norc -e -o pipefail /home/victor/Forks/act/workflow/2] user= workdir=
| {
|   "success": true,
|   "outputs": {}
| }
[test/action-check]   ✅  Success - echo '${{ toJSON(steps.first) }}'
[test/action-check] ⭐  Run echo '${{ toJSON(steps.second) }}'
[test/action-check]   🐳  docker exec cmd=[bash --noprofile --norc -e -o pipefail /home/victor/Forks/act/workflow/3] user= workdir=
| {
|   "success": true,
|   "outputs": {}
| }
[test/action-check]   ✅  Success - echo '${{ toJSON(steps.second) }}'
INFO[0002] Cleaning up container for job action-check
```
</details>

<details>
<summary><b>Output (after)</b></summary>

```txt
[test/action-check] 🚀  Start image=node:12-buster-slim
[test/action-check]   🐳  docker pull image=node:12-buster-slim platform= username= forcePull=false
[test/action-check]   🐳  docker create image=node:12-buster-slim platform= entrypoint=["/usr/bin/tail" "-f" "/dev/null"] cmd=[]
[test/action-check]   🐳  docker run image=node:12-buster-slim platform= entrypoint=["/usr/bin/tail" "-f" "/dev/null"] cmd=[]
[test/action-check]   🐳  docker exec cmd=[mkdir -m 0777 -p /var/run/act] user=root workdir=
[test/action-check] ⭐  Run First
[test/action-check]   🐳  docker exec cmd=[bash --noprofile --norc -e -o pipefail /home/victor/Forks/act/workflow/first] user= workdir=
[test/action-check]   ✅  Success - First
[test/action-check] ⭐  Run Second
[test/action-check]   🐳  docker exec cmd=[bash --noprofile --norc -e -o pipefail /home/victor/Forks/act/workflow/second] user= workdir=
[test/action-check]   ❌  Failure - Second
[test/action-check] Failed but continue next step
[test/action-check] ⭐  Run echo '${{ toJSON(steps.first) }}'
[test/action-check]   🐳  docker exec cmd=[bash --noprofile --norc -e -o pipefail /home/victor/Forks/act/workflow/2] user= workdir=
| {
|   "conclusion": "success",
|   "outcome": "success",
|   "outputs": {}
| }
[test/action-check]   ✅  Success - echo '${{ toJSON(steps.first) }}'
[test/action-check] ⭐  Run echo '${{ toJSON(steps.second) }}'
[test/action-check]   🐳  docker exec cmd=[bash --noprofile --norc -e -o pipefail /home/victor/Forks/act/workflow/3] user= workdir=
| {
|   "conclusion": "success",
|   "outcome": "failure",
|   "outputs": {}
| }
[test/action-check]   ✅  Success - echo '${{ toJSON(steps.second) }}'
INFO[0002] Cleaning up container for job action-check
```
</details>

### DIFF
--- a/pkg/runner/run_context.go
+++ b/pkg/runner/run_context.go
@@ -50,9 +50,44 @@ func (rc *RunContext) String() string {
 	return fmt.Sprintf("%s/%s", rc.Run.Workflow.Name, rc.Name)
 }
 
+type stepStatus int
+
+const (
+	stepStatusSuccess stepStatus = iota
+	stepStatusFailure
+)
+
+var stepStatusStrings = [...]string{
+	"success",
+	"failure",
+}
+
+func (s stepStatus) MarshalText() ([]byte, error) {
+	return []byte(s.String()), nil
+}
+
+func (s *stepStatus) UnmarshalText(b []byte) error {
+	str := string(b)
+	for i, name := range stepStatusStrings {
+		if name == str {
+			*s = stepStatus(i)
+			return nil
+		}
+	}
+	return fmt.Errorf("invalid step status %q", str)
+}
+
+func (s stepStatus) String() string {
+	if int(s) >= len(stepStatusStrings) {
+		return ""
+	}
+	return stepStatusStrings[s]
+}
+
 type stepResult struct {
-	Success bool              `json:"success"`
-	Outputs map[string]string `json:"outputs"`
+	Outputs    map[string]string `json:"outputs"`
+	Conclusion stepStatus        `json:"conclusion"`
+	Outcome    stepStatus        `json:"outcome"`
 }
 
 // GetEnv returns the env for the context
@@ -261,8 +296,9 @@ func (rc *RunContext) newStepExecutor(step *model.Step) common.Executor {
 	return func(ctx context.Context) error {
 		rc.CurrentStep = sc.Step.ID
 		rc.StepResults[rc.CurrentStep] = &stepResult{
-			Success: true,
-			Outputs: make(map[string]string),
+			Outcome:    stepStatusSuccess,
+			Conclusion: stepStatusSuccess,
+			Outputs:    make(map[string]string),
 		}
 		runStep, err := rc.EvalBool(sc.Step.If.Value)
 
@@ -273,7 +309,8 @@ func (rc *RunContext) newStepExecutor(step *model.Step) common.Executor {
 				return err
 			}
 			rc.ExprEval = exprEval
-			rc.StepResults[rc.CurrentStep].Success = false
+			rc.StepResults[rc.CurrentStep].Conclusion = stepStatusFailure
+			rc.StepResults[rc.CurrentStep].Outcome = stepStatusFailure
 			return err
 		}
 
@@ -295,12 +332,13 @@ func (rc *RunContext) newStepExecutor(step *model.Step) common.Executor {
 		} else {
 			common.Logger(ctx).Errorf("  \u274C  Failure - %s", sc.Step)
 
+			rc.StepResults[rc.CurrentStep].Outcome = stepStatusFailure
 			if sc.Step.ContinueOnError {
 				common.Logger(ctx).Infof("Failed but continue next step")
 				err = nil
-				rc.StepResults[rc.CurrentStep].Success = true
+				rc.StepResults[rc.CurrentStep].Conclusion = stepStatusSuccess
 			} else {
-				rc.StepResults[rc.CurrentStep].Success = false
+				rc.StepResults[rc.CurrentStep].Conclusion = stepStatusFailure
 			}
 		}
 		return err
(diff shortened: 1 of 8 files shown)


pkg/runner/run_context.go
```diff
@@ -50,9 +50,44 @@ func (rc *RunContext) String() string {
 	return fmt.Sprintf("%s/%s", rc.Run.Workflow.Name, rc.Name)
 }
 
+type stepStatus int
```
ZauberNerd, line 53: What is the reason for the `stepStatus` being an `int`? To me (I'm no golang expert) it seems overly complicated to implement the marshaling interface here when a simple string would suffice?

ZauberNerd (commented):
Tested it locally with the following workflow and everything looks fine 🎉 :
```yaml
name: CI

on: [push]

jobs:
  contexts:
    runs-on: ubuntu-latest
    steps:
      - id: succeeds
        run: echo "success"
      - id: continues
        continue-on-error: true
        run: echo "continue" && exit 1
      - id: steps-context
        env:
          STEPS_CONTEXT: ${{ toJson(steps) }}
        run: echo "$STEPS_CONTEXT"
```
But it would be nice if you could add some tests, for example to test for the differing values of `conclusion` and `outcome` in case `continue-on-error` is applied.

pkg/runner/run_context.go
victorpopkov: At first, I've used a simple string. However, having an enum type makes sense because you don't have to depend on the strings themselves. It should become easier to reuse, change or extend step statuses in future. For example:

- More statuses could be added like "cancelled" or "skipped" if needed, and there wouldn't be a need to change the already existing business logic.
- It's easier to rename them by taking advantage of the integrated code editor refactor features.
- It's easier to rename the strings themselves, like changing "success" to "successful" or something like that.
- Performance. String comparisons are usually not as efficient as simply comparing enums, etc.

I'm not A Go expert, so for me, it's more like a force of habit using enum types in cases like that.

ZauberNerd: Thanks for the explanation, that makes sense to me. 👍

victorpopkov:
I could extend `pkg/runner/testdata` to add a test case there. And/Or extend `run_context_test.go`. Going to look into that.

victorpopkov:
Force-pushed to rebase. Let me know if you need commits to be squashed.

ZauberNerd (approved):
Thanks, looks good to me. No need to squash, as the mergify bot does that automatically.

catthehacker (approved):

cplee (approved):

PR #889, opened 2021-11-17
feat: try to read ref and sha from event payload if available

ZauberNerd:
With this change `act` will try to populate the `githubContext.ref` and
`githubContext.sha` with values read from the event payload.

Caveats:
- `page_build` should not have a ref
- `status` should not have a ref
- `registry_package` should set the ref to the branch/tag but the
  payload isn't documented
- `workflow_call` should set ref to the same value as its caller but the
  payload isn't documented
- most of the events should set the sha to the last commit on the ref
  but unfortunately the sha is not always included in the payload,
  therefore we use the sha from the local git checkout

### DIFF
--- a/pkg/model/github_context.go
+++ b/pkg/model/github_context.go
@@ -1,5 +1,12 @@
 package model
 
+import (
+	"fmt"
+
+	"github.com/nektos/act/pkg/common"
+	log "github.com/sirupsen/logrus"
+)
+
 type GithubContext struct {
 	Event            map[string]interface{} `json:"event"`
 	EventPath        string                 `json:"event_path"`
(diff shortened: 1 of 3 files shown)


catthehacker:
Could you add some test cases?

catthehacker (approved):

catthehacker (changes requested):
Still missing test cases

ZauberNerd:
@catthehacker I added some tests, can you please take another look?

KnisterPeter (approved):

catthehacker (approved):
Looks more than "some" 😄 thanks a lot

cplee (approved):

PR #904, opened 2021-11-25
fix: do not replace '.' in floating point numbers

ZauberNerd:


ChristopherHX:
FYI Such a fix is included in https://github.com/nektos/act/pull/792/files#diff-1b1bd72cb2c57bae52ca138d34eafb6521db6556f94db086dfb2b6efdb83613b and much more. I supended this and let it to be closed, because I was to lazy to test it and merge it into my fork.

However I come to the conclusion it would be better to translate the c# interpreter to golang instead of using the otto engine and hacking javascript to match github.

The official interpreter is here https://github.com/actions/runner/tree/main/src/Sdk/DTExpressions2/Expressions2. I use it in my own local c# github actions runner, it works a lot of better than the act equivalent.

ZauberNerd:
@ChristopherHX we've found more issues with the current otto expression parser and were looking for a different way.
We've now found https://github.com/rhysd/actionlint which has a golang API: https://pkg.go.dev/github.com/rhysd/actionlint and we are trying to implement an interpreter using it. You can see our progress here on this branch: https://github.com/ZauberNerd/act/tree/exprparser (not sure, if we'll finish that though - for now it is simple, but later on it will get more complex).

ZauberNerd:
Btw, that actionlint package has a few other goodies which would make great additions to act
/cc @cplee @catthehacker

catthehacker:
I'm aware of actionlint but it's not something I looked into for various reasons.

ChristopherHX:
Actionslint seems to have a far more conforming parser than act and haven't found any problem during my first test.

@ZauberNerd 
I like your approach much more than what act does today, on top of that better error messages. Not shure what would speak against using it.

ZauberNerd:
Closing this in favor of: https://github.com/nektos/act/pull/908

PR #908, opened 2021-11-29
Refactor expression evaluator to use parser from actionlint package

ZauberNerd:
~~This is an early draft to showcase the general direction we're going with this change.~~

~~We will of course rebase the commits, so that we can create separate PRs for the preparation-changes (refactoring of contexts, etc).~~

~~Also there's still a few TODO comments that we have to resolve before this PR is ready to merge.~~

~~Anyways, feedback is welcome :)~~

Next steps:
- [x] manually track position in evaluator (actionlint parser gives us only the starting position of a token, but we need to know how far we've parsed)
- [x] extend `toString` to be able to print all types
- [x] get `act` tests to pass with the `exprparser` package
- [x] clean-up errors, todos, tests, etc


catthehacker:
I think `model` is better place than `common` for contexts and stepresults

ZauberNerd:
@catthehacker @ChristopherHX we're a bit stumped by how `if: xxx` works in GitHub actions and wanted to know if you have done any testing there or have a clue why the conditions evaluate the way they do:
This is the relevant section of the workflow: https://github.com/ZauberNerd/act-test/actions/runs/1578574757/workflow#L1137-L1294
This is how they evaluate: https://github.com/ZauberNerd/act-test/runs/4522196500?check_suite_focus=true

For example:
- `if: null` executes, while `if: ${{ null }}` does not execute
- `if: ''` executes, while `if: env.*` does not execute (and here's no difference between putting it in `${{}}` or not

ChristopherHX:
Weird..
One job of yours fails in my fork at the same comment, I enabled ACTIONS_STEP_DEBUG=true to see evaluation steps
https://github.com/ChristopherHX/act-test/runs/4522756256?check_suite_focus=true

Test if: env.TRUE does execute on my github action replicatation with a semi official c# runner.

ZauberNerd:
> Weird.. One job of yours fails in my fork at the same comment, I enabled ACTIONS_STEP_DEBUG=true to see evaluation steps https://github.com/ChristopherHX/act-test/runs/4522756256?check_suite_focus=true
> 
> Test if: env.TRUE does execute on my github action replicatation with a semi official c# runner.
> 
> EDIT: I think for more details I need to diff the official runner message vs. my custom crafted one.

Thanks, `ACTIONS_STEP_DEBUG` is a good idea and gives some insight into what is happening - although not why :D 
Anyway, I think that was the nudge we needed to continue our investigation here.

ChristopherHX:
Selfhosted runner windows and your ifs job fails to 100%
https://github.com/ChristopherHX/act-test/runs/4523021564?check_suite_focus=true

ChristopherHX:
> * `if: null` executes, while `if: ${{ null }}` does not execute

`if: null` yaml null is the same as removing the if so only `success()`
`if: ${{ null }}` null in expression syntax  is the same as `success() && null`, null is falsy

> * `if: ''` executes, while `if: env.*` does not execute (and here's no difference between putting it in `${{}}` or not

`if: ''` same as with null, github evaluates `success()`

`if: env.*` Report this as bug to github, your env variables are lowercased by the actions service, but the unix runner is casesentive unlike case insensitive on windows

pkg/runner/expression.go
```diff
@@ -247,15 +244,58 @@ func (ee expressionEvaluator) toString(in interface{}) string {
...
+	case int:
+		return evaluated.(int) != 0, nil
+	case float64:
+		num := evaluated.(float64)
+		if math.IsNaN(num) {
+			return false, nil
+		}
+		return num != 0, nil
+	default:
+		return false, fmt.Errorf("Unable to map return type to boolean for '%s'", expr)
```
catthehacker, line 277: You can do this
```suggestion
	switch t := evaluated.(type) {
	case bool:
		return t, nil
	case string:
		return t != "", nil
	case int:
		return t != 0, nil
	case float64:
		if math.IsNaN(t) {
			return false, nil
		}
		return t != 0, nil
	default:
		return false, fmt.Errorf("Unable to map return type to boolean for '%s'", expr)
```

pkg/exprparser/interpreter.go
```diff
@@ -539,7 +539,11 @@ func (impl *interperterImpl) evaluateFuncCall(funcCallNode *actionlint.FuncCallN
 	case "format":
 		return impl.format(args[0], args[1:]...)
 	case "join":
-		return impl.join(args[0], args[1])
+		if len(args) == 1 {
+			return impl.join(args[0], reflect.ValueOf(","))
+		} else {
+			return impl.join(args[0], args[1])
+		}
```
catthehacker, line 546: unnecessary `else` block

ZauberNerd:
@catthehacker @cplee this PR is now ready for review.

catthehacker:
Looks good

pkg/exprparser/functions_test.go
```diff
@@ -0,0 +1,247 @@
...
+		{`contains(3.141592, 3.14) }}`, true, "contains-number-number"},
+		{`contains(true, 'u') }}`, true, "contains-bool-str"},
+		{`contains(null, '') }}`, true, "contains-null-str"},
+		{`contains(fromJSON('["first","second"]'), 'first') }}`, true, "contains-item"},
+		{`contains(fromJSON('[null,"second"]'), '') }}`, true, "contains-item-null-empty-str"},
+		{`contains(fromJSON('["","second"]'), null) }}`, true, "contains-item-empty-str-null"},
+		// TODO: investigate why this evaluates to false
+		// {`contains(fromJSON('[true,"second"]'), 'true') }}`, true, "contains-item-bool-arr"},
+		// TODO: investigate why this evaluates to false
+		// {`contains(fromJSON('["true","second"]'), true) }}`, true, "contains-item-str-bool"},
```
catthehacker, line 31: missed?

pkg/runner/expression.go
```diff
@@ -1,655 +1,257 @@
...
-	case c >= '0' && c <= '9':
-		return true
-	case c == '_' || c == '-':
-		return true
-	default:
-		return false
+// NewExpressionEvaluator creates a new evaluator
+func (sc *StepContext) NewExpressionEvaluator() ExpressionEvaluator {
+	rc := sc.RunContext
+	// todo: cleanup EvaluationEnvironment creation
```
catthehacker, line 68: missed?

pkg/exprparser/functions_test.go
KnisterPeter: No, these are behaving strange on github and we are not sure what to do with them.
Therefore we left the TODO comment and defered that to a later point.

pkg/runner/expression.go
KnisterPeter: @ZauberNerd Do you still see a reason for this todo?

pkg/exprparser/functions_test.go
catthehacker: they evaluate to `false` because you compare `true` with `"true"`
```suggestion
		// this looks for string true in array containing boolean true
		// {`contains(fromJSON('[true,"second"]'), 'true') }}`, true, "contains-item-bool-arr"},
		// this looks for boolean true in array containing string true
		// {`contains(fromJSON('["true","second"]'), true) }}`, true, "contains-item-str-bool"},
```

catthehacker: can you link me the docs article?

KnisterPeter: Sure: https://docs.github.com/en/actions/learn-github-actions/expressions#functions

catthehacker: I think the docs are correct. Since the inserted values are an `Array` in this case. And searched `item` can't be casted to other type otherwise it would introduce a bug.
> Some functions cast values to a string to perform comparisons. GitHub casts data types to a string using these conversions:

image

pkg/runner/expression.go
ZauberNerd: I don't think there's something urgent to do here. I mean we could try to consolidate some things between StepContext and RunContext, but for now it works, so I'd rather see it merged and if need be addressed in a separate PR.

pkg/exprparser/functions_test.go
KnisterPeter: @zaubernerd Then we should change the impl to not coerce Array items.

ZauberNerd: @catthehacker see also this workflow, where we tried the above cases on GH actions: https://github.com/ZauberNerd/act-test/runs/4559124896?check_suite_focus=true#step:3:251

ZauberNerd: @KnisterPeter we don't coerce the Array to a string. `contains` accepts either an array or a string as its input.
And the weird thing is that `contains([3.14, 'foo'], '3.14')` and `contains(['3.14', 'bar'], 3.14)` seem to cast values in the array to a string while comparing.

catthehacker: @ZauberNerd I'm unsure what you're trying to show me. All cases in that workflow align with what I'm saying and the docs.

KnisterPeter: @zaubernerd we are coercing all Array items to string. That's what we need to change

catthehacker: @KnisterPeter although now that I think of it, there might be a bug. This seem to be type casted while it shouldn't
```yaml
[ "${{ contains(fromJSON('[null,"second"]'), '') }}" = "true" ]
[ "${{ contains(fromJSON('["","second"]'), null) }}" = "true" ]
```

KnisterPeter: As I've said before, we are coercing all Array items which we shouldn't do. Our implementation is just wrong.

ZauberNerd: @catthehacker yes, exactly that. GitHub's behaviour seems inconsistent here.
The above linked workflow has these lines in it, some of which indicate that GH does type coercions, but then the `true` case does not which contradicts the observation and the docs:
```
[ "${{ contains(fromJSON('[null,"second"]'), '') }}" = "true" ]
[ "${{ contains(fromJSON('["","second"]'), null) }}" = "true" ]
# TODO: investigate why this evaluates to false
# [ "${{ contains(fromJSON('[true,"second"]'), 'true') }}" = "true" ]
[ "${{ contains(fromJSON('["true","second"]'), 'true') }}" = "true" ]
# TODO: investigate why this evaluates to false
# [ "${{ contains(fromJSON('["true","second"]'), true) }}" = "true" ]
[ "${{ contains(fromJSON('[3.14,"second"]'), '3.14') }}" = "true" ]
[ "${{ contains(fromJSON('[3.14,"second"]'), 3.14) }}" = "true" ]
[ "${{ contains('3.141592', 3.14) }}" = "true" ]
[ "${{ contains(3.141592, '3.14') }}" = "true" ]
[ "${{ contains(3.141592, 3.14) }}" = "true" ]
```

ZauberNerd: Maybe @ChristopherHX can help us out here too :)

catthehacker: ref: 
https://github.com/actions/runner/blob/be9632302ceef50bfb36ea998cea9c94c75e5d4d/src/Sdk/DTPipelines/Pipelines/ContextData/JTokenExtensions.cs#L15

catthehacker: What happens is if left side is `string`, it compares `string == string`
If left side is an `Array` (https://github.com/actions/runner/blob/be9632302ceef50bfb36ea998cea9c94c75e5d4d/src/Sdk/DTExpressions2/Expressions2/Sdk/Functions/Contains.cs#L26-L29) it does special comparison of objects inside the array (https://github.com/actions/runner/blob/be9632302ceef50bfb36ea998cea9c94c75e5d4d/src/Sdk/DTExpressions2/Expressions2/EvaluationResult.cs#L227-L272)

catthehacker: and most important code for you
```csharp
            // Same kind
            if (leftKind == rightKind)
            {
            [...]
            }

            return false;
```
https://github.com/actions/runner/blob/be9632302ceef50bfb36ea998cea9c94c75e5d4d/src/Sdk/DTExpressions2/Expressions2/EvaluationResult.cs#L234

catthehacker: ref: https://github.com/actions/runner/blob/be9632302ceef50bfb36ea998cea9c94c75e5d4d/src/Sdk/DTExpressions2/Expressions2/EvaluationResult.cs#L81

ChristopherHX:
Great work, your parser can handle
- newlines in string literals
- ${{test}} in string literals
- the array operator

unlike act's current implementation.
I guess, this makes my workarounds for the expressions parser of my fork redundant thanks a lot.

During stress testing this PR, I discovered that your code fails to execute by accessing undefined properties of the matrix object. `${{matrix.test6}}` should evaluate to nothing. I created this workflow, to avoid to reintroduce an expressions bug in github-act-runner. 

Hmm removing `${{matrix.test6}}` isn't enought either? Invalid format string, need to look deeper.
```yaml
name: format-substitution
on: push
jobs:
  format-substitution-test:
    strategy:
      matrix:
        test1: [ false ]
        test2: [ true ]
        test3: [ 0 ]
        test4: [ 99 ]
        test5: [ 'Hello World' ]
    runs-on: [ubuntu-latest]
    steps:
    - run: |
        echo ${{matrix.test1}}-${{matrix.test2}}-${{matrix.test3}}-${{matrix.test4}}-${{matrix.test5}}${{matrix.test6}}
        echo ${{github.actor}}
        echo ${{github.event_name}}
        test="Hello World"
        echo $test
        echo ${test}
```
Src: https://github.com/ChristopherHX/github-act-runner/blob/main/.github/testworkflows/format-substitution.yml

This workflow succeeds with an empty run step.

ChristopherHX:
Ah I got it, you forget to escape `${test}` to `${{test}}`, during conversion from multiple expressions to a big format expression

ChristopherHX:
@ZauberNerd @KnisterPeter 
Can you please add this commit to your PR? https://github.com/ChristopherHX/act/commit/312d39590e15444cc9c7818450b994a7c562e3f5
This fixes the two things I mentioned today.
Github actions don't fail to evaluate on most errors, like access undefined map property.
However invalid json string in fromJSON and an invalid format string does fail your workflow.

ZauberNerd:
@ChristopherHX I think something with inputs and secrets broke while rebasing against master and now `TestRunEvent/uses-nested-composite` and `TestRunEvent/local-action-via-composite-dockerfile` are failing.
I guess it is related to this change: https://github.com/nektos/act/commit/9868e1377274a7657d4aac6bd2b66c0d564405cb#diff-4fe45e900ed33a3395bb42c8d8a85c1afb94987b2368bfe31d62e9e1ebd69060R449-R579 which I tried to incorporate here: https://github.com/nektos/act/blob/686898ecdeea992bd2810c1b15096604afb237cf/pkg/runner/expression.go#L39-L56 and here: https://github.com/nektos/act/blob/686898ecdeea992bd2810c1b15096604afb237cf/pkg/runner/expression.go#L112
Since you made some changes there, do you have an immediate idea what I'm doing wrong?

ChristopherHX:
I'm debugging it carefully, there are expression evaluation errors in the log. Some values are empty.

ChristopherHX:
@ZauberNerd The inputs context is always available, also on workflowlevel see reusable workflows.
The error was, that the inputs context was empty on runcontext level. 
https://github.com/ChristopherHX/act/commit/4c99bd7ca7151eff66b293a743c5f684b387c544

pkg/exprparser/functions_test.go
ZauberNerd: @catthehacker updated: https://github.com/nektos/act/pull/908/commits/73cf7e35b4765b1133f95dfcb303428997237ab2

ZauberNerd:
The last open item (wrong implementation of `contains()`) has been fixed. Please take another look.
/cc @catthehacker @cplee

catthehacker (approved):
Thanks for that massive work, I tried to review whole PR carefully and so far it seems good to me

cplee (approved):

PR #913, opened 2021-12-03
feat: print plan summary upfront to executing it

KnisterPeter:
This change will print a toc at debug level upfront to running a workflow. It will help prepare outputs for (e.g. UI) before executing and parsing the log output.

We would like to add this to the output of `act`, since we are parsing the output with a computer to create a more meaningful output in a website for it. The log boundaries are created to be easily machine parsable.


catthehacker:
I would like to not have that, since it's visual noise and doesn't help act

KnisterPeter:
Okay, we will then keep in on our side.
I'll close the PR as soon as I've moved into our Org.

PR #923, opened 2021-12-11
ci: replace superlinter with megalinter

catthehacker:
- replaced https://github.com/github/super-linter with https://github.com/megalinter/megalinter (megalinter is maintained better than superlinter)
- pass `context` to `Executor` for `gitClone`, this was recommended by `contextcheck` linter but later on I noticed it's a required 
  fix anyway for logging, also pass context to Docker engine version check func
- fix `.goreleaser.yml`, in-editor schema validation (and schema linter) screams about `int`s being `int`s
- correct `dependabot.yml` quotes
- move `CGO_ENABLED` to global `env`
- remove unnecessary `nolint` comments


cplee:
Thanks for the quick fix @catthehacker

PR #931, opened 2021-12-16
refactor: export and move shared contexts into pkg/model

ZauberNerd:
This commit moves the `githubContext`, `jobContext` and `stepResult` structs
from the `runner` package to the `model` package in preparation for #908
because the `expression.go` file lives in the `runner` package and would
introduce cyclic dependencies with the `exprparser` package.


ZauberNerd:
@catthehacker if this PR is a change you'd be willing to merge before #908 then it is ready for review from our side.

catthehacker (approved):
I guess it looks ok, @cplee opinion on this will be more valuable

cplee (dismissed):
These should have been in `model` package from start...i was lazy 🤷‍♂️

catthehacker (approved):

cplee (approved):

PR #932, opened 2021-12-16
fix: GetEnv need to be initialized before first access

ZauberNerd:
This commit ensures that the run context env is initialized before its
first access.
This is done in preparation for #908

- [ ] polish this patch and remove the `fixme` comment (cc @KnisterPeter)


catthehacker:
can't this be part of #908 ?

ZauberNerd:
> can't this be part of #908 ?

Yes, I can do that. I just thought it would be better to keep #908 smaller and since this change technically doesn't belong there, I extracted it into a separate preparation PR.

PR #934, opened 2021-12-20
fix: all copied files now owned by runner inside the container

### LINKED ISSUE #398: Write files as user, not root
Hi there, first of all, I'd really love to thank you for this wonderful project. It's been a wonderful aid for my personal projects. 

I would like to suggest a feature to add to this project. When new files are created (e.g. documentations / MAKE files), the owner of the file is `root` and not _user_, even with the `--bind` flag turned on. I feel it would make more sense for the files to be owned by _user_ and not `root`, especially since the container is supposed to be 'bound' to the working directory.

I have tried to change the ownership by explicitly including `chown ` _user_, but being a containerised environment, _user_ does not exist and returns the error message `chown: invalid user:` _user_

joeyparrish:
This recursively changes ownership of all copied files and other working folders to `runner` inside the container, or whichever user is specified by `--container-user`.

Fixes #398 both with and without `--bind`, thanks to the `--container-user` patch by @catthehacker:
https://github.com/catthehacker/act-fork/commit/a4dd723dd182201a99db24a86a590d28749661ce

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -54,7 +54,7 @@ func Execute(ctx context.Context, version string) {
 	rootCmd.Flags().BoolVar(&input.useGitIgnore, "use-gitignore", true, "Controls whether paths specified in .gitignore should be copied into container")
 	rootCmd.Flags().StringArrayVarP(&input.containerCapAdd, "container-cap-add", "", []string{}, "kernel capabilities to add to the workflow containers (e.g. --container-cap-add SYS_PTRACE)")
 	rootCmd.Flags().StringArrayVarP(&input.containerCapDrop, "container-cap-drop", "", []string{}, "kernel capabilities to remove from the workflow containers (e.g. --container-cap-drop SYS_PTRACE)")
-	rootCmd.Flags().StringVarP(&input.containerUser, "container-user", "", "", "UID or username which will be used for running act containers")
+	rootCmd.Flags().StringVarP(&input.containerUser, "container-user", "", "runner", "UID or username which will be used for running act containers")
 	rootCmd.Flags().BoolVar(&input.autoRemove, "rm", false, "automatically remove container(s)/volume(s) after a workflow(s) failure")
 	rootCmd.PersistentFlags().StringVarP(&input.actor, "actor", "a", "nektos/act", "user that triggered the event")
 	rootCmd.PersistentFlags().StringVarP(&input.workflowsPath, "workflows", "W", "./.github/workflows/", "path to workflow file(s)")
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -54,6 +54,7 @@ func Execute(ctx context.Context, version string) {
 	rootCmd.Flags().BoolVar(&input.useGitIgnore, "use-gitignore", true, "Controls whether paths specified in .gitignore should be copied into container")
 	rootCmd.Flags().StringArrayVarP(&input.containerCapAdd, "container-cap-add", "", []string{}, "kernel capabilities to add to the workflow containers (e.g. --container-cap-add SYS_PTRACE)")
 	rootCmd.Flags().StringArrayVarP(&input.containerCapDrop, "container-cap-drop", "", []string{}, "kernel capabilities to remove from the workflow containers (e.g. --container-cap-drop SYS_PTRACE)")
+	rootCmd.Flags().StringVarP(&input.containerUser, "container-user", "", "", "UID or username which will be used for running act containers")
 	rootCmd.Flags().BoolVar(&input.autoRemove, "rm", false, "automatically remove container(s)/volume(s) after a workflow(s) failure")
 	rootCmd.PersistentFlags().StringVarP(&input.actor, "actor", "a", "nektos/act", "user that triggered the event")
 	rootCmd.PersistentFlags().StringVarP(&input.workflowsPath, "workflows", "W", "./.github/workflows/", "path to workflow file(s)")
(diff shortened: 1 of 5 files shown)


pkg/runner/run_context.go
```diff
@@ -202,8 +203,10 @@ func (rc *RunContext) startJobContainer() common.Executor {
 			rc.JobContainer.Start(false),
 			rc.JobContainer.UpdateFromImageEnv(&rc.Env),
 			rc.JobContainer.UpdateFromEnv("/etc/environment", &rc.Env),
-			rc.JobContainer.Exec([]string{"mkdir", "-m", "0777", "-p", ActPath}, rc.Env, "root", ""),
+			rc.JobContainer.Exec([]string{"mkdir", "-m", "0777", "-p", ActPath}, rc.Env, "0", ""),
 			rc.JobContainer.CopyDir(copyToPath, rc.Config.Workdir+string(filepath.Separator)+".", rc.Config.UseGitIgnore).IfBool(copyWorkspace),
+			rc.JobContainer.Exec([]string{"chown", "-R", rc.Config.ContainerUser, copyToPath}, rc.Env, "root", "").IfBool(copyWorkspace),
+			rc.JobContainer.Exec([]string{"chown", "-R", rc.Config.ContainerUser, "/home/runner"}, rc.Env, "root", ""),
```
catthehacker, line 209: This will fail on image that don't have such path

cmd/root.go
```diff
@@ -54,6 +54,7 @@ func Execute(ctx context.Context, version string) {
 	rootCmd.Flags().BoolVar(&input.useGitIgnore, "use-gitignore", true, "Controls whether paths specified in .gitignore should be copied into container")
 	rootCmd.Flags().StringArrayVarP(&input.containerCapAdd, "container-cap-add", "", []string{}, "kernel capabilities to add to the workflow containers (e.g. --container-cap-add SYS_PTRACE)")
 	rootCmd.Flags().StringArrayVarP(&input.containerCapDrop, "container-cap-drop", "", []string{}, "kernel capabilities to remove from the workflow containers (e.g. --container-cap-drop SYS_PTRACE)")
+	rootCmd.Flags().StringVarP(&input.containerUser, "container-user", "", "runner", "UID or username which will be used for running act containers")
```
catthehacker, line 57: This will fail on image that doesn't have `runner`

pkg/runner/run_context.go
joeyparrish: Which images would those be?  I thought the runner software was always installed in act images.

cmd/root.go
catthehacker: Should be `0`, otherwise any other value has chance to break

pkg/runner/run_context.go
catthehacker: All images except `large` will not have that path

catthehacker, line 209:
```suggestion
			rc.JobContainer.Exec([]string{"chown", "-R", rc.Config.ContainerUser, copyToPath}, rc.Env, "0", "").IfBool(copyWorkspace),
			rc.JobContainer.Exec([]string{"chown", "-R", rc.Config.ContainerUser, "/home/runner"}, rc.Env, "0", ""),
```
In case image doesn't have `root`

joeyparrish: I've been tinkering with some of the other images in the README, and I see what you mean.  I'm going to experiment first with creating that folder and setting the value of the `HOME` environment variable to that.

What I've found so far is that if the UID specified in --container-user doesn't exist in the container, `$HOME` ends up being `/`, which breaks a bunch of my workflows.  It should be an easy fix to create and set a home directory, though.

Let me know if you think this is the wrong direction.

joeyparrish: Thanks!

cmd/root.go
joeyparrish: I tested with `ghcr.io/catthehacker/ubuntu:full-latest` because that has everything I need for my workflows.  But this has some coincidental features that made everything work.  My local UID is 1000, and ubuntu:full-latest has a user with that UID (runneradmin).  This allows workflows using `sudo` to succeed.

When my local UID doesn't exist in a container, such as `ubuntu:act-latest`, I get errors like `sudo: you do not exist in the passwd database`.

It seems to me that in order to make this work in the general case and be compatible with GitHub's environment, we would need to:

1. Create a `runner` user (the name is arbitrary) during container setup if it doesn't exist
2. Add `runner` to sudoers if it's not already there
3. Change the UID of `runner` to match `--container-user`
4. Change ownership of the working directory and runner home directory to match that UID

What are your thoughts on that?

I'm also starting to wonder what the advantage of the `--container-user` option is.  For me, I only ever want to use it in combination with `--bind`, set to my local UID, so that I can get the output of a job with the correct local ownership.  Is there another major use-case for it?

If not, perhaps we could drop it and always use the local user UID.  In my opinion, this would make `act` work for workflows where it currently doesn't, without the need for users to start using new command-line options.

pkg/runner/run_context.go
catthehacker: I don't think it's inherently wrong but it's easy to get it wrong.

We used to set `HOME` in act and that broke things. Best option would be to look for home path in `/etc/passwd`

joeyparrish: In the case I'm describing, though, there is no corresponding /etc/passwd entry.

catthehacker: Yes, that's correct, I misread that

catthehacker, line 209:
```suggestion
			rc.JobContainer.Exec([]string{"chown", "-R", rc.Config.ContainerUser, "/home/runner", ActPath}, rc.Env, "root", ""),
```

cmd/root.go
catthehacker: > I'm also starting to wonder what the advantage of the `--container-user` option is.

The advantage of not breaking program completely.

PR #938, opened 2021-12-22
lint(megalinter): disable markdown link check

catthehacker:
Disable link check linter since it will cause false positives, example: https://github.com/nektos/act/runs/4609894750


catthehacker:
funny, gitter is actually down
image

cplee (approved):

PR #939, opened 2021-12-23
SSH-agent authentication for cloning actions

ZauberNerd:
This makes it easier to run act locally with GitHub enterprise, because
cloning remote actions otherwise always requires setting `GITHUB_TOKEN`.

It will also help with remote actions in private repositories.

Unfortunately a `GITHUB_TOKEN` still needs to be provided for the
`actions/checkout` action when cloning a remote repository or different
ref (i.e. not a local checkout). This *might* change, if:
actions/checkout#660 is acceptable.

Also a token needs to be provided in case the workflow contains actions
that talk to the GitHub API.

### DIFF
--- a/pkg/common/git.go
+++ b/pkg/common/git.go
@@ -15,7 +15,9 @@ import (
 	git "github.com/go-git/go-git/v5"
 	"github.com/go-git/go-git/v5/config"
 	"github.com/go-git/go-git/v5/plumbing"
+	"github.com/go-git/go-git/v5/plumbing/transport"
 	"github.com/go-git/go-git/v5/plumbing/transport/http"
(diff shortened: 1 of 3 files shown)


ZauberNerd:
@catthehacker @cplee do you have any thoughts/feedback on this draft and should I continue with it?

ZauberNerd:
Closing this, as it does not solve the `GITHUB_TOKEN` issue and forwarding the `SSH_AUTH_SOCK` to the clone action is unlikely.
I'll continue on another PR to read an oauth token from the `gh` CLI config if it is available or use the `gh` packages to create a new token (like this: https://github.com/ZauberNerd/act/commit/558eb6d4375b09c1039297b03de36c2e998015c6)

PR #948, opened 2022-01-02
fix: docker cp of dangling symlink (#943)

### LINKED ISSUE #943: Issue: act actions/checkout fails for dangling symlinks
## System information

- Operating System: macOS
- Architecture: x64 (64-bit)
- Docker version: 20.10.11
- `act` version: 0.2.25

## Expected behaviour

checkout should run and the echo step should run

## Actual behaviour

pseudo checkout failed

## Workflow and/or repository

https://github.com/jsoref/act-dangling

## Steps to reproduce

```
git clone https://github.com/jsoref/act-dangling
cd act-dangling
act
```

## `act` output

<details>
  <summary>Log</summary>

```
[Test/Test] 🚀  Start image=catthehacker/ubuntu:act-latest
[Test/Test]   🐳  docker pull image=catthehacker/ubuntu:act-latest platform= username= forcePull=false
[Test/Test]   🐳  docker create image=catthehacker/ubuntu:act-latest platform= entrypoint=["/usr/bin/tail" "-f" "/dev/null"] cmd=[]
[Test/Test]   🐳  docker run image=catthehacker/ubuntu:act-latest platform= entrypoint=["/usr/bin/tail" "-f" "/dev/null"] cmd=[]
[Test/Test]   🐳  docker exec cmd=[mkdir -m 0777 -p /var/run/act] user=root workdir=
[Test/Test]   🐳  docker cp src=/Users/jsoref/code/act/dangling/. dst=/Users/jsoref/code/act/dangling
[Test/Test]   🐳  docker exec cmd=[mkdir -p /Users/jsoref/code/act/dangling] user= workdir=
Error: open /Users/jsoref/code/act/dangling/baz: no such file or directory
```

</details>

jsoref:
Fixes #943


pkg/container/docker_run.go
```diff
@@ -579,6 +579,10 @@ func (cr *containerReference) copyDir(dstPath string, srcPath string, useGitIgno
 			// open files for taring
 			f, err := os.Open(file)
 			if err != nil {
+				if fi.Mode()&os.ModeSymlink == os.ModeSymlink {
+					// symlinks don't need to be copied, ignore this error
+					err = nil
+				}
```
ChristopherHX, line 585: I agree with
> symlinks don't need to be copied

but act shouldn't even try to do that, it's just wrong to copy the content of a symlink into the tar. Act just had luck that docker ignored the content of the symlink.

I prefer to return just before line 579 on symlinks like
```golang
if fi.Mode()&os.ModeSymlink == os.ModeSymlink {
   return nil
}
```
instead of patching the error code, if it is a dangling symlink.

ChristopherHX (changes requested):

ChristopherHX (dismissed):
Thanks, ( dangling ) symlinks are now  also working on windows.

ZauberNerd:
There's symlink error handling in the `io.Copy` block: https://github.com/nektos/act/blob/9f9eacf427a228631901dd2c8e02a44b479b1574/pkg/container/docker_run.go#L592 shoudn't that be removed as we will never reach it?

Also: How is this change macOS only? I don't have a mac handy so I tested on Linux and could reproduce it just as well.

jsoref:
I use act on macOS, and thus describe things based on this. It seemed unlikely to me that such a thing wouldn't have been tripped on already by Linux users, so I described things based on how I encountered them. That there is some difference between how it behaves on Linux may have contributed to how I described it.

If the problem happens for everyone, then I can change the description to not specify.

I'll remove the code you've identified.

ZauberNerd:
Thanks, yeah looks like it happens on all platforms. With the fix #971 from Christopher applied the error is now also printed for me (`DEBU[0000] open /.../act-dangling/baz: no such file or directory`).

ChristopherHX (approved):

ZauberNerd (approved):

Issue #927, opened 2021-12-14
Issue: Dynamic Matrix not working


aurimasniekis:
## System information

- Operating System: macOS
- Architecture: x64 (64-bit)
- Apple M1: no
- Docker version: 20.10.8
- Docker image used in `act`:
- `act` version: act version HEAD-cad4bc8

## Expected behaviour

The following workflow file should work, the matrix dynamic value should be evaluated after the first job is done and passed to matrix of second job.

## Actual behaviour

Matrix is not evaluated

## Workflow and/or repository

<details>
  <summary>workflow</summary>

```none
name: "Dynamic Matix"

on:
  pull_request:
  push:

jobs:
  bar:
    name: Bar
    runs-on: ubuntu-latest
    steps:
      - id: output_data
        run: |
          echo '::set-output name=matrix::"{"package": ["a", "b"]}"'
    outputs:
      matrix: ${{ steps.output_data.outputs.matrix }}

  foo:
    name: Foo
    needs: [ bar ]
    runs-on: ubuntu-latest
    strategy:
      matrix: ${{ fromJson(needs.bar.outputs.matrix) }}
    steps:
      - name: foo ${{ matrix.package }}
        run: |
          echo 'Matrix "${{ matrix.package }}" JSON: ${{ toJSON(matrix) }} ${{ needs.bar.outputs.matrix }}'
```

</details>

## `act` output

<details>
  <summary>Log</summary>

```noneDEBU[0000] Loading environment from /tmp/foo/.env
DEBU[0000] Loading secrets from /tmp/foo/.secrets
DEBU[0000] Loading workflows from '/tmp/foo/.github/workflows'
DEBU[0000] Loading workflows recursively
DEBU[0000] Found workflow 'sample.yaml' in '/tmp/foo/.github/workflows/sample.yaml'
DEBU[0000] Reading workflow '/tmp/foo/.github/workflows/sample.yaml'
DEBU[0000] Correcting if statements '/tmp/foo/.github/workflows/sample.yaml'
DEBU[0000] Planning event: push
WARN[0000] unable to get git repo: unable to find git repo
WARN[0000] unable to get git revision: unable to find git repo
WARN[0000] unable to get git ref: unable to find git repo
DEBU[0000] context steps => map[]
DEBU[0000] context env => map[ACT:true]
DEBU[0000]
WARN[0000] unable to get git repo: unable to find git repo
WARN[0000] unable to get git revision: unable to find git repo
[...]

PR #964, opened 2022-01-23
fix: deep evaluate matrix strategy

ChristopherHX:
We need to wait for the previous stage to finish, otherwise we don't know how the final strategy will look like.
Then we need to evaluate all expression nodes of the strategy.matrix property, even the mapping keys.

I also added support for the insert directive and array merge operations, both are undocumented features of the github/actions which are available since 2019.

My tests aren't cover all possibilities, but the fixed issue has a test.

Most likely this change collides with @catthehacker plans of rewriting the yaml parser to actionslint, since I use the current yaml.Node types.

I have no idea how big the regressions regarding parallelism are with this change.

Resolves https://github.com/nektos/act/issues/927

### DIFF
--- a/pkg/runner/expression.go
+++ b/pkg/runner/expression.go
@@ -132,79 +132,93 @@ func (ee expressionEvaluator) evaluate(in string, isIfExpression bool) (interfac
 	return evaluated, err
 }
 
-func (ee expressionEvaluator) EvaluateYamlNode(node *yaml.Node) error {
-	if node.Kind == yaml.ScalarNode {
-		var in string
-		if err := node.Decode(&in); err != nil {
+func (ee expressionEvaluator) evaluateScalarYamlNode(node *yaml.Node) error {
+	var in string
+	if err := node.Decode(&in); err != nil {
+		return err
+	}
+	if !strings.Contains(in, "${{") || !strings.Contains(in, "}}") {
+		return nil
+	}
+	expr, _ := rewriteSubExpression(in, false)
+	if in != expr {
+		log.Debugf("expression '%s' rewritten to '%s'", in, expr)
+	}
+	res, err := ee.evaluate(in, false)
+	if err != nil {
+		return err
+	}
+	return node.Encode(res)
+}
+
+func (ee expressionEvaluator) evaluateMappingYamlNode(node *yaml.Node) error {
+	var m map[interface{}]yaml.Node
+	if err := node.Decode(&m); err != nil {
+		return err
+	}
+	// GitHub has this undocumented feature to merge maps, called insert directive
+	insertDirective := regexp.MustCompile(`\${{\s*insert\s*}}`)
+	mout := make(map[interface{}]yaml.Node)
+	for k := range m {
+		v := m[k]
+		if err := ee.EvaluateYamlNode(&v); err != nil {
 			return err
 		}
-		if !strings.Contains(in, "${{") || !strings.Contains(in, "}}") {
-			return nil
-		}
-		expr, _ := rewriteSubExpression(in, false)
-		if in != expr {
-			log.Debugf("expression '%s' rewritten to '%s'", in, expr)
-		}
-		res, err := ee.evaluate(in, false)
-		if err != nil {
-			return err
-		}
-		return node.Encode(res)
-	} else if node.Kind == yaml.MappingNode {
-		var m map[interface{}]yaml.Node
-		if err := node.Decode(&m); err != nil {
-			return err
-		}
-		// GitHub has this undocumented feature to merge maps, called insert directive
-		insertDirective := regexp.MustCompile(`\${{\s*insert\s*}}`)
-		mout := make(map[interface{}]yaml.Node)
-		for k, v := range m {
-			if err := ee.EvaluateYamlNode(&v); err != nil {
-				return err
-			}
-			if sk, ok := k.(string); ok {
-				// Merge the nested map of the insert directive
-				if insertDirective.MatchString(sk) {
-					var vm map[interface{}]yaml.Node
-					if err := v.Decode(&vm); err != nil {
-						return err
-					}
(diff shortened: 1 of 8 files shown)


pkg/runner/expression.go
```diff
@@ -130,6 +132,83 @@ func (ee expressionEvaluator) evaluate(in string, isIfExpression bool) (interfac
...
+			return err
+		}
+		return node.Encode(res)
+	} else if node.Kind == yaml.MappingNode {
+		var m map[interface{}]yaml.Node
+		if err := node.Decode(&m); err != nil {
+			return err
+		}
+		// GitHub has this undocumented feature to merge maps, called insert directive
+		insertDirective := regexp.MustCompile(`\${{\s*insert\s*}}`)
```
ChristopherHX, line 159: Sample usage of `${{insert}}`, key collisions result in template error on github and actions/runner
```yaml
on: push
jobs:
  a:
    strategy:
      matrix:
        a:
        - env:
              key1: val1
              ${{insert}}:
                 key2: val2
              ${{insert}}: ${{fromJSON('{"key3": "val3"}')}}
    runs-on: ubuntu-latest
    steps:
    - run: exit 0
```

pkg/runner/expression.go
```diff
@@ -130,6 +132,97 @@ func (ee expressionEvaluator) evaluate(in string, isIfExpression bool) (interfac
...
+	aout := make([]yaml.Node, 0)
+	for i := range a {
+		v := a[i]
+		// Preserve nested sequences
+		wasseq := v.Kind == yaml.SequenceNode
+		if err := ee.EvaluateYamlNode(&v); err != nil {
+			return err
+		}
+		// GitHub has this undocumented feature to merge sequences / arrays
+		// We have a nested sequence via evaluation, merge the arrays
```
ChristopherHX, line 201: Sample usage of the undocumented array merge operation
```
on: push
jobs:
  a:
    strategy:
      matrix:
        a:
        - env:
              key1: val1
        - ${{fromJSON('[{"key2": "val2"}]')}}
    runs-on: ubuntu-latest
    steps:
    - run: exit 0

File: pkg/runner/expression.go [hunk 3]
```diff
@@ -130,6 +132,97 @@ func (ee expressionEvaluator) evaluate(in string, isIfExpression bool) (interfac
...
+
+func (ee expressionEvaluator) EvaluateYamlNode(node *yaml.Node) error {
+	if node.Kind == yaml.ScalarNode {
+		return ee.evaluateScalarYamlNode(node)
+	} else if node.Kind == yaml.MappingNode {
+		return ee.evaluateMappingYamlNode(node)
+	} else if node.Kind == yaml.SequenceNode {
+		return ee.EvaluateSequenceYamlNode(node)
+	}
+	return nil
```
catthehacker, line 223: `switch` would be better here

ChristopherHX: Done

ChristopherHX:
I noticed that my first map approuch of evaluateMappingYamlNode failed with yaml validation errors, if I used `${{insert}}` as yaml key multiple times. It seems to be a somewhat complex change, loosly based on how actions/runner works internally.
More parts of act should migrate to use this method for evaluation e.g.
- job.strategy itself the whole block is evaluatable
- job.env the whole block is evaluatable
- job.defaults the whole block is evaluatable
- job.runs-on
- job.steps.*.with the whole block is evaluatable
- job.steps.*.env the whole block is evaluatable
- job.steps.*.continue-on-error
- job.steps.*.timeout-minutes

ZauberNerd (approved):
So if I understand it correctly, you take the parsed yaml node, turn it back into its string representation, run `Evaluate` on that and then encode it as a `yaml.Node` again?
Feels a bit hack-y to me, but I think it is fine for now until we have refactored the whole workflow parsing to use actionlint and then we can remove this again.

pkg/runner/runner.go
```diff
@@ -115,59 +115,71 @@ func New(runnerConfig *Config) (Runner, error) {
...
-			}
-
-			b := 0
-			for i, matrix := range matrixes {
-				rc := runner.newRunContext(run, matrix)
-				rc.JobName = rc.Name
-				if len(matrixes) > 1 {
-					rc.Name = fmt.Sprintf("%s-%d", rc.Name, i+1)
+	stagePipeline := make([]common.Executor, 0)
+	for i := range plan.Stages {
```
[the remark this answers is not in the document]
ChristopherHX: golang reuses the loop variables `s`, `stage`, the linter warns you if you do that after my changes.
So what happens if you reference `s`
- you capture the loop variable in a function
- the loop finishes
- the callbacks with references to the loop variables get called
- `s`, `stage` are always assigned to the last stage.

_I also believed that I can safely reference loop variables in `func(){...}`_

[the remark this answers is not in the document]
ZauberNerd: Ah, yes. We ran into the same issue here: https://github.com/nektos/act/blob/master/pkg/runner/job_executor_test.go#L147-L155 ;)

KnisterPeter (approved):
Looks fine from my side

cplee (approved):
lgtm - thanks

ChristopherHX:
> you take the parsed yaml node, turn it back into its string representation

I only look for expressions in scalar values could be boolean, number or string. Maybe we could read the Value field directly like in if statements. I just find it myself safer to use the decode function.

> run Evaluate on that and then encode it as a yaml.Node again

Everything which isn't an expression is skipped and not encoded again
After evaluate we get a raw boolean, number, string, array or mapping which needs to be encoded as yaml.Node.


For actionslint, we may need to replace all yaml types. I rely much on the current act api with yaml.Node, with a lot of convertion functions to translate job requests of the github actions service into act's layout.

> workflow parsing to use actionlint and then we can remove this again.

I doubt we can remove this later without breaking my test cases.

PR #968, opened 2022-01-26
docs: Expand the GITHUB_TOKEN section

### LINKED ISSUE #967: Enhancement: Add pointers for GITHUB_TOKEN to README
## Describe feature
When I started using act, I kept running into issues because I was missing the `GITHUB_TOKEN` and I couldn't be bothered to figure it out for a long time.

I believe this is such a common use-case that it bears spelling out briefly, what this token is and where to get it.

gzur:
When I started using act, it kept failing because I was missing the `GITHUB_TOKEN` and I couldn't be bothered to figure it out.

I was unsure how to get this token, so I did some research and wanted to lessen the confusion of future users.

I believe this is such a common use-case that it bears spelling out.

Fixes #967

### DIFF
--- a/README.md
+++ b/README.md
@@ -202,7 +202,15 @@ It will save that information to `~/.actrc`, please refer to [Configuration](#co
   -W, --workflows string                 path to workflow file(s) (default "./.github/workflows/")
 ```
 
-In case you want to pass a value for `${{ github.token }}`, you should pass `GITHUB_TOKEN` as secret: `act -s GITHUB_TOKEN=[insert token or leave blank for secure input]`.
+## `GITHUB_TOKEN`
+
+Github [automatically provides](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#about-the-github_token-secret) a `GITHUB_TOKEN` secret when running workflows inside Github.
+
+If your workflow depends on this token, you need to create a [personal access token](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token) and pass it to `act`  as a secret:
+
+```bash
+act -s GITHUB_TOKEN=[insert token or leave blank for secure input]
+```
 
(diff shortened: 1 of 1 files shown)


KnisterPeter (approved):
Thank you

gzur:
Something in the test suite is timing out.

Probably not related to my markdown 😂

catthehacker:
If we are to extend the documentation regarding that, it would be good to include information that `GITHUB_TOKEN` can be logged in shell file history if not inserted through secure input and command is not prefixed with whitespace

gzur:
> [...] `GITHUB_TOKEN` can be logged in shell file history if not inserted through secure input a

Though I agree with the sentiment, this is hardly the forum. Leaking credentials The problem is mentioned, but  - prefixing by whitespace is not a portable solution

> and command is not prefixed with whitespace

This is not true for every shell / platform

gzur:
_Accidentally closed and re-opened this PR 😂_

I added a warning about leaking the `GITHUB_TOKEN` via shell history. I tried to be as concise as possible, since secure use of the shell is a huge topic and extremely platform specific and this README is hardly the forum for such a broad discussion.

catthehacker (approved):

catthehacker:
> > If we are to extend the documentation regarding that, it would be good to include information that GITHUB_TOKEN can be logged in shell file history if not inserted through secure input and command is not prefixed with whitespace
> 
> I added a warning about leaking the `GITHUB_TOKEN` via shell history. I tried to be as concise as possible, since secure use of the shell is a huge topic and extremely platform specific and this README is hardly the forum for such a broad discussion. @catthehacker

Thanks, we are discussing better format for documentation which can have more descriptive information but for now this will suffice for sure.

ChristopherHX (approved):

cplee (approved):

Issue #946, opened 2021-12-30
Issue: Panic after failing to load an action


ChristopherHX:
## System information

- Operating System: Windows
- Architecture: x64 (64-bit)
- Apple M1: no 
- Docker version: 20.10.6
- Docker image used in `act`: `catthehacker/ubuntu:act-latest`
- `act` version: 0.2.26-master ( e9c4609dcadf70739e34f79680e69df973b24d8e since 1891c72ab158508e36009d16b24913fa5836422b )

## Expected behaviour

Print a proper errormessage and don't panic.

## Actual behaviour

Panic after some failures without any cleanup.

## Workflow and/or repository

<details>
  <summary>workflow</summary>

```yaml
on: push
name: no-panic-after-failure
jobs:
  nopanic:
    runs-on: ubuntu-latest
    steps:
    - uses: ./path/to/non-existent-action
```

</details>

## Steps to reproduce

- execute the provided workflow
`act -W .\no-panic-after-failure\push.yml -v`
- older versions of act like v0.2.25 don't panic.

## `act` output

<details>
  <summary>Log</summary>

```none
time="2021-12-30T20:30:55+01:00" level=debug msg="Loading environment from /act-panic-after-failure\\pkg\\runner\\testdata\\.env"
time="2021-12-30T20:30:55+01:00" level=debug msg="Loading secrets from /act-panic-after-failure\\pkg\\runner\\testdata\\.secrets"
time="2021-12-30T20:30:55+01:00" level=debug msg="Loading workflow '/act-panic-after-failure\\pkg\\runner\\testdata\\no-panic-after-failure\\push.yml'"
time="2021-12-30T20:30:55+01:00" level=debug msg="Reading workflow '/act-panic-after-failure\\pkg\\runner\\testdata\\no-panic-after-failure\\push.yml'"
time="2021-12-30T20:30:55+01:00" level=debug msg="Correcting if statements '/act-panic-after-failure\\pkg\\runner\\testdata\\no-panic-after-failure\\push.yml'"
time="2021-12-30T20:30:55+01:00" level=debug msg="Planning event: push"
time="2021-12-30T20:30:55+01:00" level=debug msg="Loading slug from git directory '/.git'"
time="2021-12-30T20:30:55+01:00" level=debug msg="Found revision: b5929156e58dec0605e18a50e2168c626bee9b6c\n"
time="2021-12-30T20:30:55+01:00" level=debug msg="Loading revision from git directory '/.git'"
[...]

PR #971, opened 2022-01-27
Fix: regression run after failure

ChristopherHX:
Restore the old behavior of the then executor. A lot of code in act relies on the fact that the pipeline executor fails fast instead of failing later.

Here a list of regressions
- https://github.com/nektos/act/issues/946 failing to read an action => panic
- returning an error at this line https://github.com/nektos/act/pull/964/commits/f538b28a53f03e082feca382f18b1cdea0192d0c#diff-3ca8bf2cdff0ec928622b093c6af8793a24f0f5f9a397c86e1df94a385be6e44R130 => panic
- failed to connect to docker => panic
- no cleanup on panic, the finally executor doesn't run
This broke error handling of act. I prefer to revert this function to it's previous version and create a new finally executor to archive the same in a less error prone.

Also print the error of the step, since it is useful to know why my action failed. E.g. if you used continue-on-error in a composite action.

Resolves https://github.com/nektos/act/issues/946

@KnisterPeter @ZauberNerd You introduced the regression in https://github.com/nektos/act/commit/1891c72ab158508e36009d16b24913fa5836422b

The observation of the edit of this comment https://github.com/nektos/act/pull/948#issuecomment-1022949930 is caused by the same regression, job failure without reason.

### DIFF
--- a/pkg/common/executor.go
+++ b/pkg/common/executor.go
@@ -136,12 +136,13 @@ func (e Executor) Then(then Executor) Executor {
 			case Warning:
 				log.Warning(err.Error())
 			default:
-				SetJobError(ctx, err)
+				log.Debugf("%+v", err)
+				return err
 			}
-		} else if ctx.Err() != nil {
-			SetJobError(ctx, ctx.Err())
 		}
-
+		if ctx.Err() != nil {
+			return ctx.Err()
+		}
 		return then(ctx)
 	}
 }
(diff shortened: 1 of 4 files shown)


ZauberNerd (approved):

KnisterPeter (approved):

KnisterPeter:
@ChristopherHX Thanks for taking care. We where not fully aware that our change was breaking the cleanup.

PR #894, opened 2021-11-20
Don't interpolate joboutputs, before job is done

BlackDex:
All credits go to @ChristopherHX which fixed this issue (https://github.com/ChristopherHX/act/commit/fd9f40cdc02dc72c40a9c33450d638d6e8e8ee57).
I only created a PR for this so it will be fixed in the upstream binary.

This fixes #758

Issue #758, opened 2021-07-20
Issue: Only first step is able to set output variables


domenix:
## System information

- Operating System: Windows 10
- Architecture: x64 (64-bit)
- Apple M1: no
- Docker version: 20.10.7
- `act` version: c865a56 (current commit on `master` at the time of submission)

## Expected behaviour

Output variables should be able to be defined in multiple steps under the same job

## Actual behaviour

Currently, output variables can only be successfully defined in the first step inside a job (see official [documentation](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idoutputs) for reference).

## Workflow and/or repository

<details>
  <summary>workflow</summary>

```yml
name: Test
on:
  workflow_dispatch:

jobs:
  set_vars:
    name: Set variables
    runs-on: ubuntu-latest
    steps:
    - id: set_1
      run: |
        echo "::set-output name=var_1::$(echo hello)"
        echo "::set-output name=var_2::$(echo world)"
    - id: set_2
      run: echo "::set-output name=var_3::$(echo 1234)"
    outputs:
      variable_1: ${{ steps.set_1.outputs.var_1 }}
      variable_2: ${{ steps.set_1.outputs.var_2 }}
      variable_3: ${{ steps.set_2.outputs.var_3 }}
  printing:
    name: Printing
    needs: set_vars
    runs-on: ubuntu-latest
    steps:
      - name: Print variables from another job
        run: |
          echo "${{ needs.set_vars.outputs.variable_1 }}" # Gets printed
          echo "${{ needs.set_vars.outputs.variable_2 }}" # As well
          echo "${{ needs.set_vars.outputs.variable_3 }}" # Empty space in logs, variable not resolved
```

</details>

## Steps to reproduce

Run: `act workflow_dispatch -W <path to the workflow file with content above>`

## `act` output

<details>
  <summary>Log</summary>

```none
[Test/Set variables] 🚀  Start image=catthehacker/ubuntu:act-latest
[Test/Set variables]   🐳  docker run image=catthehacker/ubuntu:act-latest platform= entrypoint=["/usr/bin/tail" "-f" "/dev/null"] cmd=[]
[Test/Set variables]   🐳  docker exec cmd=[mkdir -m 0777 -p /var/run/act] user=root
[Test/Set variables] ⭐  Run echo "::set-output name=var_1::$(echo hello)"
echo "::set-output name=var_2::$(echo world)"
[...]

PR #984, opened 2022-02-08
refactor: extract RunContext Executor in JobExecutor

Poltergeist:
This splits the executor from the `RunContext` into its own function
called `newJobExecutor`.
We defined an interface called `jobInfo` which is implemented by the `RunContext`.
This enables better unit testing because only a small interface needs to
be mocked.

This is a preparation for implementing pre and post actions.

### DIFF
--- a/pkg/runner/run_context.go
+++ b/pkg/runner/run_context.go
@@ -244,57 +244,38 @@ func (rc *RunContext) interpolateOutputs() common.Executor {
 	}
 }
 
-// Executor returns a pipeline executor for all the steps in the job
-func (rc *RunContext) Executor() common.Executor {
-	steps := make([]common.Executor, 0)
+func (rc *RunContext) startContainer() common.Executor {
+	return rc.startJobContainer()
+}
 
-	steps = append(steps, func(ctx context.Context) error {
-		if len(rc.Matrix) > 0 {
-			common.Logger(ctx).Infof("\U0001F9EA  Matrix: %v", rc.Matrix)
-		}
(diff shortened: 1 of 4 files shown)


KnisterPeter (approved):

ZauberNerd (approved):

ChristopherHX (approved):
I prefer to see refactoring like this merged instead of causing unneeded merge conflicts, like with my composite PR.

pkg/runner/run_context.go
```diff
@@ -244,57 +244,38 @@ func (rc *RunContext) interpolateOutputs() common.Executor {
...
+	return rc.Matrix
+}
 
-		return nil
-	})
+func (rc *RunContext) result(result string) {
+	rc.Run.Job().Result = result
+}
 
-	return common.NewPipelineExecutor(steps...).Finally(rc.interpolateOutputs()).Finally(func(ctx context.Context) error {
```
cweckesser, line 292: This issue #758 was closed on November 2021, and from the comments it could be understood that the issue was actually fixed here #894. It seems like the changes introduced in this file might have caused the issue to respawn. I reproduced it in my local environment by creating a simple job with two steps, and it was only possible to set output variables in the first one.
@Poltergeist I would really appreciate it if you could verify that the fix from November 2021 was affected by these changes.

KnisterPeter: Hi @cweckesser, can you create a new issue with a reproducible workflow to show this case?

PR #988, opened 2022-02-09
fix: always execute closeContainer() executor

ZauberNerd:
During our earlier refactoring in https://github.com/nektos/act/pull/984 we accidentally changed the
behaviour in such a way that the `closeContainer()` executor was never
called.

This commit restores the earlier behaviour.

Ref:
* https://github.com/nektos/act/pull/984/files#diff-c057d66dc9657d8428e290c69871596e2b567bb8fecad62a99cab54398131a84L294
* https://github.com/nektos/act/pull/984/files#diff-ea9d5c93d769ef9b268932dd9990363e97fc3bec8a00114979d049bead5dd718R68

There are two other commits in this PR that add tests for the above regression and rename some interfaces.


KnisterPeter (approved):
Thanks

ChristopherHX (approved):
I better ran my fd leak test on the previous change... This is not something you would notice easily within act.

catthehacker (approved):

PR #990, opened 2022-02-10
refactor: default empty if: statement to if: success() in evaluator & remove `FixIfStatement()`

ZauberNerd:
Previously the code to default an empty `if:` statement in the yaml file
was implemented in different files in the model package.
Now an empty `if:` statement defaults to `success()` in the expression
evaluator.

The introduction of the expression evaluator seems to have made the
`FixIfStatement` functions obsolete, as the test case
`TestRunEvent/issue-598` works even without these functions.


KnisterPeter (approved):

ChristopherHX (approved):
Much cleaner now thanks.

cplee (approved):

PR #993, opened 2022-02-11
Added instructions to install on fedora using copr package

### LINKED ISSUE #992: Enhancement: Added support for Fedora
At this moment it's not possible to install act into fedora using  the fedora official mirrors or even copr.
Since act is targeting Debian and Arch, i think that it will be a nice improvement to include Fedora too.

Rubemlrm:
Added information on README.MD about the installation of act on Fedora based distros.

At this moment this is installed by COPR only, despite this i had to use the `act-cli`instead of `act`since exists already a package with the name of `act` that is the short name for another software.

Fix #992

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,6 +77,15 @@ scoop install act
 yay -S act
 ```
 
+### [COPR](https://copr.fedorainfracloud.org/coprs/rubemlrm/act-cli/) (Linux)
+
+[![copr-shield](https://img.shields.io/copr/version/act)](https://copr.fedorainfracloud.org/coprs/rubemlrm/act-cli/)
+
+```shell
+dnf copr enable rubemlrm/act-cli
+dnf install act-cli
+```
+
 ### [Nix](https://nixos.org) (Linux/macOS)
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -77,6 +77,15 @@ scoop install act
 yay -S act
 ```
 
+### [COPR](https://copr.fedorainfracloud.org/coprs/rubemlrm/act-cli/) (Linux)
+
+[![copr-shield](https://img.shields.io/copr/version/act)](https://copr.fedorainfracloud.org/coprs/rubemlrm/act-cli/)
```
catthehacker, line 82: Doesn't seem like there is shield for COPR

Rubemlrm: My bad. I will do a commit to fix this. Thanks a lot for the reply. Can I just leave a link like was made for the unix step?

catthehacker: Sure

Rubemlrm: done thanks a lot

catthehacker (approved):

KnisterPeter (approved):

ChristopherHX (approved):

Issue #977, opened 2022-02-03
Issue: Git commands on private repos cause act to hang up


ajrice6713:
## System information

- Operating System: macOS latest
- Architecture: x64 (64-bit)
- Apple M1: no 
- Docker version: 20.10.12
- Docker image used in `act`: ghcr.io/catthehacker/ubuntu
- `act` version: act version 0.2.25

## Expected behaviour

Should be able to clone/push to private repositories in act

## Actual behaviour

Act simply stops dead in its tracks on `git clone` or git push`

### Workflow Example

```yml
name: Listner Test

on:
  workflow_call:
    inputs:
      branch_name:
        description: 'Branch name passed from the caller workflow'
        required: true
        type: string
 
 workflow_dispatch: 
    inputs:
      branch_name:
        description: 'Branch name passed from the caller workflow'
        required: true
        type: string

jobs:
  run_listner_test:
    name: Publish Staging Docsite
    runs-on: ubuntu-latest
    steps:
    - uses: actions/checkout@v2

    - name: Create and Push Branch
      run: |
        git checkout -b ${{ github.event.inputs.branch_name }}
        mkdir dest-directory
        git clone --branch ${{ github.event.inputs.branch_name }} https://github.com/{my-org}/{my-private-repo} dest-directory
        rm -r dest-directory/*.
        git add .
        git commit -m "Clone Specs from my-private-repo repo"
        git push origin ${{ github.event.inputs.branch_name }}
```

## Steps to reproduce

Call git clone or git pull on a private repository

## `act` output

```none
act -j run_listner_test --secret-file ./.secrets --eventpath ./listener.event
[Listner Test/Publish Staging Docsite] 🚀  Start image=ghcr.io/catthehacker/ubuntu:act-latest
[Listner Test/Publish Staging Docsite]   🐳  docker pull image=ghcr.io/catthehacker/ubuntu:act-latest platform= username= forcePull=false
[Listner Test/Publish Staging Docsite]   🐳  docker create image=ghcr.io/catthehacker/ubuntu:act-latest platform= entrypoint=["/usr/bin/tail" "-f" "/dev/null"] cmd=[]
[Listner Test/Publish Staging Docsite]   🐳  docker run image=ghcr.io/catthehacker/ubuntu:act-latest platform= entrypoint=["/usr/bin/tail" "-f" "/dev/null"] cmd=[]
[...]

PR #995, opened 2022-02-13
Set default GIT_TERMINAL_PROMPT env var to disable git prompts

sosedoff:
I've recently hit the same issue as https://github.com/nektos/act/issues/977 - `act` hangs when project needs to pull a private repo. While there are a few ways to go about fixing the access (configure github tokem, setup creds, use ssh endpoint, etc), we should def drop any git prompts as the act environment is not interactive. 

With this PR, `act` will error out with ```fatal: could not read Username for 'https://github.com': terminal prompts disabled``` message.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -23,6 +23,12 @@ import (
 	"github.com/nektos/act/pkg/runner"
 )
 
+var (
+	defaultEnvVars = map[string]string{
+		"GIT_TERMINAL_PROMPT": "0", // Disable terminal prompts from Git
+	}
+)
+
 // Execute is the entry point to running the CLI
 func Execute(ctx context.Context, version string) {
 	input := new(Input)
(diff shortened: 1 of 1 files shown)


ChristopherHX (approved):
This might be a solution for git, but what is if you call..
- python3
- node

...?

The main reason for this is that act tells docker to allocate a pseudo tty, while not sending End of Terminal ( ascii 4 ) in the input stream. 
Sending End of Terminal leads to this message
```
Username for 'https://github.com': fatal: could not read Username for 'https://github.com': Success
```

actions/runner doesn't allocate a pseudo tty, therefore git never tries to read a password.

**It would be great to have a test, but it seems to be complicated to do it where you added the env variable.**

sosedoff:
Any other approvers? Not sure if i need to make any adjustments.

catthehacker:
If GHA runners don't set that envvar, then we shouldn't either.

sosedoff:
But GHA runners aren't in the same environment as `act`, hence why it exhibits this issue. If my proposed changes aren't aligned with what you have in mind, feel free to close the PR. Im just trying to fix the obvious thing here, thats all.

sosedoff:
Closing due to lack of interest from maintainers.

Issue #114, opened 2020-02-28
Feature request: specify matrix on command line


dobesv:
For matrix builds, it would be nice if I could limit the build to a subset (or just one) matrix option value for a build.

e.g.

```
act --job build --matrix java:13
```

my current workaround is to just edit the workflows file to remove the other matrix options before running `act`, which works fine but isn't ideal.

Julian:
Is this actually done? I can't seem to find how to do it.

Julian:
Got it, thanks for the confirmation I wasn't missing it.

codeandcats:
This would be a handy feature.

My workflow as matrix of 5 node versions to test against and when running locally act seems to spin them all up in parallel and my system drops to a crawl.

Being able to specify matrix items to run or at least how many containers at most to spin up at the same time would make this useable again.

m-mcgowan:
+1 for this being a useful feature - often there's a CI problem with a specific matrix combination. Having to run all of them slows down development. Being able to specify the matrix variables on the command line as overrides would allow for a might tighter focus.

johnboiles:
> just edit the workflows file to remove the other matrix options before running act

Handy workaround at least!

PR #1003, opened 2022-02-16
feat: run jobs in parallel

KnisterPeter:
This changes fixes and restructures the parallel execution of jobs.
The previous changes limiting the parallel execution did break this
and allowed only one job in parallel.

While we run #CPU jobs in parallel now, the jobs added per job-matrix
add to this. So we might over-commit to the capacity, but at least
it is limited.

### DIFF
--- a/pkg/runner/runner.go
+++ b/pkg/runner/runner.go
@@ -121,8 +121,8 @@ func (runner *runnerImpl) NewPlanExecutor(plan *model.Plan) common.Executor {
 		stage := plan.Stages[i]
 		stagePipeline = append(stagePipeline, func(ctx context.Context) error {
 			pipeline := make([]common.Executor, 0)
-			stageExecutor := make([]common.Executor, 0)
 			for r, run := range stage.Runs {
+				stageExecutor := make([]common.Executor, 0)
 				job := run.Job()
 				if job.Strategy != nil {
 					strategyRc := runner.newRunContext(run, nil)
@@ -140,7 +140,6 @@ func (runner *runnerImpl) NewPlanExecutor(plan *model.Plan) common.Executor {
 					maxParallel = len(matrixes)
 				}
 
-				b := 0
 				for i, matrix := range matrixes {
 					rc := runner.newRunContext(run, matrix)
 					rc.JobName = rc.Name
@@ -167,13 +166,8 @@ func (runner *runnerImpl) NewPlanExecutor(plan *model.Plan) common.Executor {
 							return nil
 						})(common.WithJobErrorContainer(WithJobLogger(ctx, jobName, rc.Config.Secrets, rc.Config.InsecureSecrets)))
 					})
-					b++
-					if b == maxParallel {
-						pipeline = append(pipeline, common.NewParallelExecutor(maxParallel, stageExecutor...))
-						stageExecutor = make([]common.Executor, 0)
-						b = 0
-					}
 				}
+				pipeline = append(pipeline, common.NewParallelExecutor(maxParallel, stageExecutor...))
 			}
 			return common.NewParallelExecutor(runtime.NumCPU(), pipeline...)(ctx)
 		})
--- a/pkg/runner/runner.go
+++ b/pkg/runner/runner.go
@@ -169,13 +169,13 @@ func (runner *runnerImpl) NewPlanExecutor(plan *model.Plan) common.Executor {
 					})
 					b++
 					if b == maxParallel {
-						pipeline = append(pipeline, common.NewParallelExecutor(stageExecutor...))
+						pipeline = append(pipeline, common.NewParallelExecutor(maxParallel, stageExecutor...))
 						stageExecutor = make([]common.Executor, 0)
 						b = 0
 					}
 				}
 			}
-			return common.NewPipelineExecutor(pipeline...)(ctx)
+			return common.NewParallelExecutor(runtime.NumCPU(), pipeline...)(ctx)
 		})
 	}
 
--- a/pkg/common/executor.go
+++ b/pkg/common/executor.go
@@ -91,39 +91,37 @@ func NewErrorExecutor(err error) Executor {
 }
 
 // NewParallelExecutor creates a new executor from a parallel of other executors
-func NewParallelExecutor(executors ...Executor) Executor {
+func NewParallelExecutor(parallel int, executors ...Executor) Executor {
 	return func(ctx context.Context) error {
-		errChan := make(chan error)
+		work := make(chan Executor, len(executors))
+		errs := make(chan error, len(executors))
 
-		for _, executor := range executors {
-			e := executor
-			go func() {
-				err := e.ChannelError(errChan)(ctx)
-				if err != nil {
-					log.Fatal(err)
+		for i := 0; i < parallel; i++ {
+			go func(work <-chan Executor, errs chan<- error) {
+				for executor := range work {
+					errs <- executor(ctx)
 				}
-			}()
+			}(work, errs)
 		}
+
+		for i := 0; i < len(executors); i++ {
+			work <- executors[i]
+		}
+		close(work)
 
 		// Executor waits all executors to cleanup these resources.
 		var firstErr error
 		for i := 0; i < len(executors); i++ {
-			if err := <-errChan; err != nil && firstErr == nil {
+			err := <-errs
+			if firstErr == nil {
 				firstErr = err
 			}
 		}
+
 		if err := ctx.Err(); err != nil {
 			return err
 		}
 		return firstErr
-	}
-}
-
-// ChannelError sends error to errChan rather than returning error
-func (e Executor) ChannelError(errChan chan error) Executor {
-	return func(ctx context.Context) error {
-		errChan <- e(ctx)
-		return nil
 	}
 }
 
--- a/pkg/common/executor_test.go
+++ b/pkg/common/executor_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/assert"
 )
@@ -79,14 +80,19 @@ func TestNewParallelExecutor(t *testing.T) {
 	ctx := context.Background()
 
 	count := 0
+	maxCount := 0
 	emptyWorkflow := NewPipelineExecutor(func(ctx context.Context) error {
 		count++
+
+		time.Sleep(2 * time.Second)
+
 		return nil
 	})
 
-	err := NewParallelExecutor(2, emptyWorkflow, emptyWorkflow)(ctx)
-	assert.Equal(2, count)
+	err := NewParallelExecutor(2, emptyWorkflow, emptyWorkflow, emptyWorkflow)(ctx)
 
+	assert.Equal(2, count)
+	assert.Equal(2, maxCount)
 	assert.Nil(err)
 }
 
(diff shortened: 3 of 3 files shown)


ZauberNerd (approved):

pkg/runner/runner.go
```diff
@@ -169,13 +169,13 @@ func (runner *runnerImpl) NewPlanExecutor(plan *model.Plan) common.Executor {
...
 					if b == maxParallel {
-						pipeline = append(pipeline, common.NewParallelExecutor(stageExecutor...))
+						pipeline = append(pipeline, common.NewParallelExecutor(maxParallel, stageExecutor...))
 						stageExecutor = make([]common.Executor, 0)
 						b = 0
 					}
 				}
 			}
-			return common.NewPipelineExecutor(pipeline...)(ctx)
+			return common.NewParallelExecutor(runtime.NumCPU(), pipeline...)(ctx)
```
danhunsaker, line 178: Should probably be configurable via CLI switch (a la `make -j` and friends), and default to `runtime.NumCPU() - 1` to prevent complete system crashes. (I've had a few such crashes, myself, over the last week, so I'm particularly sensitive to them at the moment.)

danhunsaker (changes requested):

pkg/runner/runner.go
KnisterPeter: Yes, thought about making this configurable as well.
We are already overcomming here, since the limit to num-cpu applies only to parallel jobs.
We would spawn 4x that number if all jobs have matrices.

KnisterPeter: I don't think this could, by default, lead to a system crash.
If you imagine a computer with 4 CPUs, we will at most run 16 goroutines. Unless each of them are burning a CPU completely, this should be okay.
Even saving a CPU by reducing parallel tasks does not help then, since that would still be 12 goroutines.
But with 'max-parallel' on the task level it could be reduced to 1 per talk which would be totally fine.

So I think the configurable limit could be done in another PR.

catthehacker: There is no point to introduce additional complexity of limiting CPU used by act and making new flags. We already provide options to choose specific jobs/workflows.

Another thing is that this CPU limit is not that important at all as CPU scheduler will handle those containers just fine.
What you should be afraid of is what is running in the workflow (e.g. build Rust crate and you will see)

I've never seen a system OS "crash" due to CPU overcommitting.

cplee (approved):

ChristopherHX (changes requested):
Please add a test for NewParallelExecutor and test that not more than `parallel` executors are running.
max-parallel in strategy isn't working anymore after this change on my win11 machine.
Also it seems that all my 9 jobs runs in parallel, while I only have an 8 core cpu.

ChristopherHX:
<Details>
<summary>Here are the workflows I used to test this PR</summary>

```yaml
on: push
jobs:
  p1:
    runs-on: ubuntu-latest
    strategy:
      max-parallel: 1
      matrix:
        0:
        - 0
        - 1
        1:
        - 0
        - 1
        2:
        - 0
        - 1
    steps:
    - run: exit 0
  p2:
    runs-on: ubuntu-latest
    strategy:
      max-parallel: 1
      matrix:
        0:
        - 0
        - 1
        1:
        - 0
        - 1
        2:
        - 0
        - 1
    steps:
    - run: exit 0
  p3:
    runs-on: ubuntu-latest
    strategy:
      max-parallel: 1
      matrix:
        0:
        - 0
        - 1
        1:
        - 0
        - 1
        2:
        - 0
        - 1
    steps:
    - run: exit 0
  p4:
    runs-on: ubuntu-latest
    strategy:
      max-parallel: 1
      matrix:
        0:
        - 0
        - 1
        1:
        - 0
        - 1
        2:
        - 0
        - 1
    steps:
    - run: exit 0
  p5:
    runs-on: ubuntu-latest
    strategy:
      max-parallel: 1
      matrix:
        0:
        - 0
        - 1
        1:
        - 0
        - 1
        2:
        - 0
        - 1
    steps:
    - run: exit 0
  p6:
    runs-on: ubuntu-latest
    strategy:
      max-parallel: 1
      matrix:
        0:
        - 0
        - 1
        1:
        - 0
        - 1
        2:
        - 0
        - 1
    steps:
    - run: exit 0
  p7:
    runs-on: ubuntu-latest
    strategy:
      max-parallel: 1
      matrix:
        0:
        - 0
        - 1
        1:
        - 0
        - 1
        2:
        - 0
        - 1
    steps:
    - run: exit 0
  p8:
    runs-on: ubuntu-latest
    strategy:
      max-parallel: 1
      matrix:
        0:
        - 0
        - 1
        1:
        - 0
        - 1
        2:
        - 0
        - 1
    steps:
    - run: exit 0
  p9:
    runs-on: ubuntu-latest
    strategy:
      max-parallel: 1
      matrix:
        0:
        - 0
        - 1
        1:
        - 0
        - 1
        2:
        - 0
        - 1
    steps:
    - run: exit 0
```
```yaml
on: push
jobs:
  p1:
    runs-on: ubuntu-latest
    strategy:
      matrix:
        0:
        - 0
        - 1
        1:
        - 0
        - 1
        2:
        - 0
        - 1
    steps:
    - run: exit 0
  p2:
    runs-on: ubuntu-latest
    strategy:
      matrix:
        0:
        - 0
        - 1
        1:
        - 0
        - 1
        2:
        - 0
        - 1
    steps:
    - run: exit 0
  p3:
    runs-on: ubuntu-latest
    strategy:
      matrix:
        0:
        - 0
        - 1
        1:
        - 0
        - 1
        2:
        - 0
        - 1
    steps:
    - run: exit 0
  p4:
    runs-on: ubuntu-latest
    strategy:
      matrix:
        0:
        - 0
        - 1
        1:
        - 0
        - 1
        2:
        - 0
        - 1
    steps:
    - run: exit 0
  p5:
    runs-on: ubuntu-latest
    strategy:
      matrix:
        0:
        - 0
        - 1
        1:
        - 0
        - 1
        2:
        - 0
        - 1
    steps:
    - run: exit 0
  p6:
    runs-on: ubuntu-latest
    strategy:
      matrix:
        0:
        - 0
        - 1
        1:
        - 0
        - 1
        2:
        - 0
        - 1
    steps:
    - run: exit 0
  p7:
    runs-on: ubuntu-latest
    strategy:
      matrix:
        0:
        - 0
        - 1
        1:
        - 0
        - 1
        2:
        - 0
        - 1
    steps:
    - run: exit 0
  p8:
    runs-on: ubuntu-latest
    strategy:
      matrix:
        0:
        - 0
        - 1
        1:
        - 0
        - 1
        2:
        - 0
        - 1
    steps:
    - run: exit 0
  p9:
    runs-on: ubuntu-latest
    strategy:
      matrix:
        0:
        - 0
        - 1
        1:
        - 0
        - 1
        2:
        - 0
        - 1
    steps:
    - run: exit 0
```

</Details>

KnisterPeter:
@ChristopherHX
As stated in this PR we are overcomitting to the CPU. That means we have #cpu * matrix max-parallel goroutines running in parallel.
If you definitely want to make sure you are not running more than #cpu jobs in parallel, set `max-parallel` in all matrices to 1.

This is by design. Limiting the parallelism to a global maximum is way harder to implement.

ChristopherHX:
@KnisterPeter 
I limited max-parallel to 1 and it changed nothing.

KnisterPeter:
I will add a test for that.

KnisterPeter:
@ChristopherHX  Fixed it, please review and retest again.
The problem wasn't in the `ParallelExecutor` but in the way the pipeline was created in the main `Executor`.

KnisterPeter:
Okay, I was too fast. It worked locally, but it seems I have to inspect this a bit more.

KnisterPeter:
This should be fine now (I hope :smile:)

ChristopherHX (approved):
It works now as expected.

pkg/runner/runner.go
danhunsaker: Choosing specific workflows and jobs doesn't choose specific matrices, and anyway only running specific matrices wouldn't solve the larger issue of trying to run an entire workflow locally without overloading the entire host system in the process. As you pointed out, it's the jobs inside the containers that cause problems, and act doesn't have any reasonable way to control that. But it does have a reasonable way to control itself, and letting the user control how many jobs to run in parallel would let full workflow runs with multiple parallel builds/tests be feasible on hardware that isn't a data center like GHA itself.

KnisterPeter: It is perfectly possible to set `max-parallel` in your matrices jobs to `1` and act will run at most one job per CPU in your system. That should be fine for 99.99% of all cases.

catthehacker: You can chip in here: https://github.com/nektos/act/issues/114
Specifying CPU limit is waste of time and resources

ChristopherHX: For that usecase you could try my act clone https://github.com/ChristopherHX/runner.server. In my opinion it is more feature complete than nektos/act.
The `Runner.Client` or `gharun` executable is an act replacement
It has similar cli switches than `nektos/act`, but is based on actions/runner c#.
It has a `--parallel` switch for selecting how many actions/runner are spawned on your pc. ( defaults to 1 )
It has a `--matrix` switch for selecting a specfix matrix entry of a job, but then you won't be able to run all jobs of your workflow.
However it has a performance problem if you aren't on linux and use linux containers ( on windows you should call it from a windows subsystem for linux shell )

danhunsaker: > It is perfectly possible to set `max-parallel` in your matrices jobs to `1`

The idea is to let GHA run things in parallel while my local system can't. Setting `max-parallel` to anything defeats that purpose.

> You can chip in here: #114

Saw that, interacted with existing comments, will be useful to have when fixing specific issues. Doesn't really address the opposite issue with running the entire workflow, though.

> Adding option to specify CPU limit is waste of time and resources

Not trying to limit CPU, but rather parallel execution. This PR already does exactly that, as does its predecessor for matrices. The switch i was requesting was to manually override the default number used as the limiter, whatever that might be. Some folks will have workflows that don't need limiting at all, and others will have systems that weren't made in the last two years and need to limit things further than the default. I recognize that i mentioned a minus-one approach in my original review, but that's entirely separate from the switch I'd like to see, and honestly irrelevant to anything, as pointed out above in responses to that review.

> For that usecase you could try my act clone 

I'll have a look. I don't need the `--matrix` switch as often as the `--parallel` one, which is why I commented here and not over on 114, but having both is nice. I imagine I'll have to alter my workflows for the implementation differences, though, which is why I was trying to stick with act if i can, since it already powers the majority of my workflows in other repos. It's just when i have to build large numbers of packages in parallel that things start to go sideways, and the ability to have more control over how act spins those tasks up becomes important rather than overkill. I can't exactly change my tech stack between local and GHA without defeating the entire purpose of a local run.

KnisterPeter: > The idea is to let GHA run things in parallel while my local system can't. Setting max-parallel to anything defeats that purpose.

You can use expressions and the `env` context with the value of `ACT` to decide if you are running locally (where you have more limited concurrency) compared to GHA where you have a farm of computers.

Others that that, a more complex implementation with a global limit which could be obeyed is probably welcome.
This PR allows to have concurrency at all, which is not defined by matrices.

danhunsaker: Right. The matrices bit was a side note.

ZauberNerd (approved):

cplee (approved):

PR #1004, opened 2022-02-17
feat: split job steps into its own files/structs

KnisterPeter:
To make the steps more testable, we refactored the step context
into smaller parts. Each step kind is moved into its own struct and
file.
This does help keep structure and reduce interface size. This also
allows easier testing and mocking of each step.

This prepares to integrate action pre- and post-steps.


KnisterPeter:
@nektos/act-maintainers I would consider this ready for review now.
The main purpose of this PR is to split the `StepContext` into multiple smaller, and better testable, parts.

The steps are aggregated in a new `JobExecutor` which creates the steps with the introduced `stepFactory`. This allows separating the logic for each step kind into it's own file and restructure logic. We already provide tests for each step kind.

Also this already contains the interfaces to integrate pre- and post-action steps into the job workflow. Currently these are empty stubs.

ZauberNerd:
@nektos/act-maintainers we've separated the changes on this PR into two commits to ease reviewing.
* The first commit [`refactor: split step_context into separate files`](https://github.com/nektos/act/pull/1004/commits/993555eaca3e8f054edbbc1e9e907d2acb946a14) only moves functions from the `step_context.go` into separate files, but does otherwise no changes.
* The second commit [`refactor: introduce step factory and make steps testable`](https://github.com/nektos/act/pull/1004/commits/45cff59488eed401f7c944aa5e90394e1d327e10) removes the `step_context.go` entirely and introduces the step factory and step interfaces.

pkg/runner/job_executor.go
```diff
@@ -47,9 +63,10 @@ func newJobExecutor(info jobInfo) common.Executor {
 			}
 			return nil
 		})
-	}
 
-	steps = append(steps, func(ctx context.Context) error {
+		postSteps = append(postSteps, step.post())
```
ChristopherHX, line 67: You messed up the order of post steps? They are called in reverse order see my other comment.

pkg/runner/run_context.go
```diff
@@ -275,19 +275,30 @@ func (rc *RunContext) steps() []*model.Step {
...
+		// run multiple times and we might modify the instance
 		stepcopy := step
-		stepExec := rc.newStepExecutor(&stepcopy)
+
+		step, err := sf.newStep(&stepcopy, rc)
+		if err != nil {
+			return common.NewErrorExecutor(err)
+		}
+		stepExec := common.NewPipelineExecutor(step.pre(), step.main(), step.post())
+
```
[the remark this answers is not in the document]
KnisterPeter: Why exactly is that wrong? In this PR all `pre()` and `post()` functions are empty stubs.
So on code besides a call is executed.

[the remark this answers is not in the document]
KnisterPeter: You mean that the current approach does not work for composite actions?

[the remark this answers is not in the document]
ChristopherHX: > You mean that the current approach does not work for composite actions?

Yes it doesn't work for composite actions, better skip them like before instead of calling them in a random order.

> pre() and post() functions are empty stubs.

Yes in this PR it is **still** ok, but in the following PR ( post action support ) it becomes a problem.

> Why exactly is that wrong?

You immediately call the post action handler, after each step within composite and don't run them in the composite post handler.

The main handler of composite never calls the pre executor, only the pre handler of composite does this if it is an remote action.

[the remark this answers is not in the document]
KnisterPeter: Good points. I'll update the PR to integrate that with composites.

pkg/runner/step_action_remote.go
```diff
@@ -0,0 +1,153 @@
...
+}
+
+func (sar *stepActionRemote) pre() common.Executor {
+	return func(ctx context.Context) error {
+		return nil
+	}
+}
+
+var (
+	stepAtionRemoteNewCloneExecutor = common.NewGitCloneExecutor
```
ZauberNerd, line 35:
```suggestion
	stepActionRemoteNewCloneExecutor = common.NewGitCloneExecutor
```

pkg/runner/step_docker.go
```diff
@@ -0,0 +1,121 @@
...
+	return &sd.env
+}
+
+func (sd *stepDocker) runUsesContainer() common.Executor {
+	rc := sd.RunContext
+	step := sd.Step
+
+	return func(ctx context.Context) error {
+		image := strings.TrimPrefix(step.Uses, "docker://")
+		eval := sd.RunContext.NewExpressionEvaluator()
```
ZauberNerd, line 56: Shouldn't you use `rc` here?
```suggestion
		eval := rc.NewExpressionEvaluator()
```

pkg/runner/step_test.go
```diff
@@ -0,0 +1,276 @@
...
+	err := setupEnv(context.Background(), sm)
+	assert.Nil(t, err)
+
+	// These are commit or system specific
+	delete((env), "GITHUB_REF")
+	delete((env), "GITHUB_SHA")
+	delete((env), "GITHUB_WORKSPACE")
+	delete((env), "GITHUB_REPOSITORY")
+	delete((env), "GITHUB_REPOSITORY_OWNER")
+	delete((env), "GITHUB_ACTOR")
```
ZauberNerd, line 161: `GITHUB_ACTOR` is hardcoded to `nektos/act`, so doesn't need to be excluded here.

pkg/runner/action_test.go
```diff
@@ -123,11 +124,86 @@ runs:
 
 			closerMock.On("Close")
 
-			sc := &StepContext{}
-			action, err := sc.readAction(tt.step, "actionDir", "actionPath", readFile, writeFile)
+			action, err := readActionImpl(tt.step, "actionDir", "actionPath", readFile, writeFile)
 
 			assert.Nil(t, err)
 			assert.Equal(t, tt.expected, action)
```
ZauberNerd, line 130: Missing `closerMock.AssertExpectations(t)`

pkg/runner/job_executor_test.go
```diff
@@ -11,170 +11,281 @@ import (
...
+			jim.On("closeContainer").Return(func(ctx context.Context) error {
 				executorOrder = append(executorOrder, "closeContainer")
 				return nil
 			})
 
-			executor := newJobExecutor(jpm)
+			executor := newJobExecutor(jim, sfm, rc)
 			err := executor(ctx)
 			assert.Nil(t, err)
 			assert.Equal(t, tt.executedSteps, executorOrder)
```
ZauberNerd, line 291: Missing `jim.AssertExpectations(t)` and `sfm.AssertExpectations(t)`

pkg/runner/job_executor_test.go
```diff
@@ -11,170 +11,281 @@ import (
...
 						return nil
 					})
-				}(stepMock)
+
+					sm.On("post").Return(func(ctx context.Context) error {
+						if tt.postSteps[i] {
+							executorOrder = append(executorOrder, "post"+stepModel.ID)
+						}
+						return nil
+					})
```
ZauberNerd, line 265: Missing `sm.AssertExpectations(t)`

pkg/runner/step_test.go
KnisterPeter: I would still like to exclude it, since it could be overridden by env-vars.
So it is out of control of this test (kind of).

pkg/runner/job_executor.go
KnisterPeter: Done

pkg/runner/run_context.go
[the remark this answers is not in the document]
KnisterPeter: > Yes it doesn't work for composite actions, better skip them like before instead of calling them in a random order.

This is a bigger preparation for pre/post steps to execute. I don't see random order here, but we are calling all pre-steps before all main steps before all post steps in order of definition in the yaml file.
The order of post steps is corrected and inverted to the definition order (as is on GHA).

Since pre and post steps in this PR are only empty functions no harm is done.
The way to implement this in the next PR (which is draft and not yet ready) should be to call all pre-steps of a composite action in it's pre step. And all post steps of a composite action it it's post step.
Like this:

```yaml
jobs:
  job:
    steps:
      - id: step1
        uses: some/action@v1
      - id: step2
        uses: some-other/action@v1
```

If `some/action` and `some-other/action` are composite with both pre and post steps then we will get this execution order:

- step1 pre
  - call all pre steps which are defined by actions inside `some/action`
- step2 pre
  - call all pre steps which are defined by actions inside `some-other/action`
- step1 main
- step2 main
- step2 post
  - call all post steps (in reverse order) which are defined by actions inside `some-other/action`
- step1 post
  - call all post steps (in reverse order) which are defined by actions inside `some/action`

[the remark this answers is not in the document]
KnisterPeter: @ChristopherHX Do you see any error in what I described in my previous comment?

ChristopherHX (dismissed):
I'm ok with this kind of refactoring. `actions/cache` did work for me in the next PR based on this one.

pkg/runner/run_context.go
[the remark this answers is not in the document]
ChristopherHX: > I don't see random order here

This is all about composite, in job level it works.

> Since pre and post steps in this PR are only empty functions no harm is done.

I agree for this PR, but I still prefer to change that specfic line related to composite. Till you implemented it for composite actions correctly.
```suggestion
		stepExec := step.main()

```

No changes needed in this PR in my point of view.

pkg/runner/step_action_local.go
```diff
@@ -0,0 +1,85 @@
...
+func (sal *stepActionLocal) main() common.Executor {
+	sal.env = map[string]string{}
+
+	return runStepExecutor(sal, func(ctx context.Context) error {
+		actionDir := filepath.Join(sal.getRunContext().Config.Workdir, sal.Step.Uses)
+
+		localReader := func(ctx context.Context) actionYamlReader {
+			_, cpath := getContainerActionPaths(sal.Step, path.Join(actionDir, ""), sal.RunContext)
+			return func(filename string) (io.Reader, io.Closer, error) {
+				tars, err := sal.RunContext.JobContainer.GetContainerArchive(ctx, path.Join(cpath, filename))
```
ZauberNerd, line 41:
```suggestion
				tars, err := sal.getRunContext().JobContainer.GetContainerArchive(ctx, path.Join(cpath, filename))
```

pkg/runner/step_action_remote.go
```diff
@@ -0,0 +1,153 @@
...
+func (sar *stepActionRemote) main() common.Executor {
+	sar.env = map[string]string{}
+
+	return runStepExecutor(sar, func(ctx context.Context) error {
+		remoteAction := newRemoteAction(sar.Step.Uses)
+		if remoteAction == nil {
+			return fmt.Errorf("Expected format {org}/{repo}[/path]@ref. Actual '%s' Input string was not in a correct format", sar.Step.Uses)
+		}
+
+		remoteAction.URL = sar.RunContext.Config.GitHubInstance
```
ZauberNerd, line 47:
```suggestion
		remoteAction.URL = sar.getRunContext().Config.GitHubInstance
```

pkg/runner/step_action_remote.go
```diff
@@ -0,0 +1,153 @@
...
+
+		github := sar.RunContext.getGithubContext()
```
ZauberNerd, line 49:
```suggestion
		github := sar.getRunContext().getGithubContext()
```

pkg/runner/step_action_remote.go
```diff
@@ -0,0 +1,153 @@
...
+		if remoteAction.IsCheckout() && isLocalCheckout(github, sar.Step) {
+			common.Logger(ctx).Debugf("Skipping local actions/checkout because workdir was already copied")
+			return nil
+		}
+
+		actionDir := fmt.Sprintf("%s/%s", sar.RunContext.ActionCacheDir(), strings.ReplaceAll(sar.Step.Uses, "/", "-"))
```
ZauberNerd, line 55:
```suggestion
		actionDir := fmt.Sprintf("%s/%s", sar.getRunContext().ActionCacheDir(), strings.ReplaceAll(sar.Step.Uses, "/", "-"))
```

pkg/runner/step_run.go
```diff
@@ -0,0 +1,166 @@
...
+}
+
+func (sr *stepRun) setupShellCommandExecutor() common.Executor {
+	return func(ctx context.Context) error {
+		scriptName, script, err := sr.setupShellCommand()
+		if err != nil {
+			return err
+		}
+
+		return sr.RunContext.JobContainer.Copy(ActPath, &container.FileEntry{
```
ZauberNerd, line 64:
```suggestion
		return sr.getRunContext().JobContainer.Copy(ActPath, &container.FileEntry{
```

pkg/runner/step_run.go
```diff
@@ -0,0 +1,166 @@
...
+// so we return proper errors before any execution or spawning containers
+// it will error anyway with:
+// OCI runtime exec failed: exec failed: container_linux.go:380: starting container process caused: exec: "${{": executable file not found in $PATH: unknown
+func (sr *stepRun) setupShellCommand() (name, script string, err error) {
+	sr.setupShell()
+	sr.setupWorkingDirectory()
+
+	step := sr.Step
+
+	script = sr.RunContext.NewStepExpressionEvaluator(sr).Interpolate(step.Run)
```
ZauberNerd, line 91:
```suggestion
	script = sr.getRunContext().NewStepExpressionEvaluator(sr).Interpolate(step.Run)
```

ZauberNerd (commented):
There are still a few places where you access the `RunContext` directly on the step implementation instead of using the `getRunContext()` function of the step interface.
Should this be made the same even though it works at the moment without compiler/test errors?
* I suppose we could change the `RunContext` of the concrete step implementations to private struct members, we would be forced to always use `getRunContext()` but we probably also couldn't just assign it in the tests.
* Or we could leave it as is.
* Or we could make them all as `getRunContext()` to be consistent for now, but it isn't guaranteed that it will stay this way in the future.

KnisterPeter:
The `getRunContext()` function is there to be able to get the `RunContext` in dependent structs (from the step). Like in actions or in functions which don't have a call target.
Since the function just returns the `RunContext` it doesn't matter if you call the function or the struct member. The compiler should be able to inline the call as well.
Not sure if we should change that, as from my point of few, it doesn't matter much.
I could change it if you opt for it.

ZauberNerd:
> The `getRunContext()` function is there to be able to get the `RunContext` in dependent structs (from the step). Like in actions or in functions which don't have a call target. Since the function just returns the `RunContext` it doesn't matter if you call the function or the struct member. The compiler should be able to inline the call as well. Not sure if we should change that, as from my point of few, it doesn't matter much. I could change it if you opt for it.

Ok, then I'm fine with this either way.
For the this PR looks good, but since I'm one of the authors as well, I'd like the others (@cplee and @catthehacker) to review it.

ChristopherHX (dismissed):
Could you please avoid doing force-pushs? This is a big PR and this makes it harder for me to track changes between my reviews.

KnisterPeter:
Sure, I'll merge with master then.

Procorners (approved):

KnisterPeter:
@nektos/act-maintainers I'm about to give up on this. I've not seen a sign of 'the direction of this pr is okay' and I'm constantly merging/rebasing and resolving conflicts.
Unless we do agree that we want to follow this, I will no update this anymore.

ChristopherHX (approved):
We can resolve issues later, get this merged

cplee (approved):
thanks for this!!

Issue #1008, opened 2022-02-22
Issue: Error in if expression return code 0.


mvrshvl:
## System information
  - Operating System: Linux
  - Architecture: x64 (64-bit)
  - Docker version: 20.10.12
  - Docker image used in `act`: golang:1.16.5-alpine
  - `act` version: act version 0.2.25-41-gff13844-dirty

## Expected behaviour
The act returns a non-zero error code about an incorrectly described yaml file 

## Actual behaviour

The act returns an error log, but the exit code is 0.

Trace with nil error:

https://github.com/nektos/act/blob/e23223ad02ddeafe581df97c7e9462df28d8e104/pkg/runner/run_context.go#L278

https://github.com/nektos/act/blob/7dbf3fcb96e2d70b3436b718c1f4a9fe4b169a4f/pkg/common/executor.go#L153

https://github.com/nektos/act/blob/e23223ad02ddeafe581df97c7e9462df28d8e104/pkg/runner/run_context.go#L414

https://github.com/nektos/act/blob/7dbf3fcb96e2d70b3436b718c1f4a9fe4b169a4f/pkg/common/executor.go#L156

## Workflow and/or repository
```
jobs:
  lints:
    if: something
    runs-on: ubuntu-latest
    timeout-minutes: 60
    container: golang:1.16.5-alpine
    steps:
      - uses: actions/checkout@v1
```
## Steps to reproduce

Run a workflow with the same job as in the step `Workflow and/or repository`.

## `act` output

[GitHub Actions/lints]   ❌  Error in if: expression - lints

ZauberNerd:
A fix for this and some improvements in the log output in case of errors is included in: https://github.com/nektos/act/pull/1009

ecwootten:
Any news on when this will make it into a release?

PR #928, opened 2021-12-15
Refactor logging

catthehacker:
Fixes #195

- add `--no-color` flag
- handle more envvars for colour disabling
- Unicode Emojis are disabled for terminals not capable of colours and/or envvars/flag
- Moved logger from `runner` to `common`
- add check for survey to verify that terminal is capable of answering it
- change `--verbose` from `bool` to `count`, single `--verbose` enables `Debug` log level, 2 or more enable `Trace` log level
  - `logrus.FieldLogger` was replaced with `logrus.Ext1FieldLogger` since `logrus.FieldLogger` does not export `Trace` log level functions

Signed-off-by: hackercat <me@hackerc.at>

PR #1009, opened 2022-02-22
Expression evaluator fixes

ZauberNerd:
This PR contains a few fixes and improvements for the new expression evaluator:
* Adds more debug output about which expression is going to be evaluated
* Use exported `IsTruthy()` function from the `exprparser` package inside `EvalBool()`
* Fail act when an expression contains errors and cannot be evaluated

Fixes: #1008 
Ref: #971

### DIFF
--- a/pkg/runner/expression.go
+++ b/pkg/runner/expression.go
@@ -142,9 +142,6 @@ func (ee expressionEvaluator) evaluateScalarYamlNode(node *yaml.Node) error {
 		return nil
 	}
 	expr, _ := rewriteSubExpression(in, false)
-	if in != expr {
-		log.Debugf("expression '%s' rewritten to '%s'", in, expr)
-	}
 	res, err := ee.evaluate(expr, false)
 	if err != nil {
 		return err
@@ -215,10 +212,6 @@ func (ee expressionEvaluator) Interpolate(in string) string {
 	}
 
 	expr, _ := rewriteSubExpression(in, true)
-	if in != expr {
-		log.Debugf("expression '%s' rewritten to '%s'", in, expr)
-	}
-
 	evaluated, err := ee.evaluate(expr, false)
 	if err != nil {
 		log.Errorf("Unable to interpolate expression '%s': %s", expr, err)
@@ -236,9 +229,6 @@ func (ee expressionEvaluator) Interpolate(in string) string {
 // EvalBool evaluates an expression against given evaluator
 func EvalBool(evaluator ExpressionEvaluator, expr string) (bool, error) {
 	nextExpr, _ := rewriteSubExpression(expr, false)
-	if expr != nextExpr {
-		log.Debugf("expression '%s' rewritten to '%s'", expr, nextExpr)
-	}
 
 	evaluated, err := evaluator.evaluate(nextExpr, true)
 	if err != nil {
@@ -312,5 +302,9 @@ func rewriteSubExpression(in string, forceFormat bool) (string, error) {
 		return in, nil
 	}
 
-	return fmt.Sprintf("format('%s', %s)", strings.ReplaceAll(formatOut, "'", "''"), strings.Join(results, ", ")), nil
+	out := fmt.Sprintf("format('%s', %s)", strings.ReplaceAll(formatOut, "'", "''"), strings.Join(results, ", "))
+	if in != out {
+		log.Debugf("expression '%s' rewritten to '%s'", in, out)
+	}
+	return out, nil
 }
--- a/pkg/runner/expression.go
+++ b/pkg/runner/expression.go
@@ -128,7 +128,9 @@ type expressionEvaluator struct {
 }
 
 func (ee expressionEvaluator) evaluate(in string, isIfExpression bool) (interface{}, error) {
+	log.Debugf("evaluating expression '%s'", in)
 	evaluated, err := ee.interpreter.Evaluate(in, isIfExpression)
+	log.Debugf("expression '%s' evaluated to '%t'", in, evaluated)
 	return evaluated, err
 }
 
@@ -224,8 +226,6 @@ func (ee expressionEvaluator) Interpolate(in string) string {
 		return ""
 	}
 
-	log.Debugf("expression '%s' evaluated to '%s'", expr, evaluated)
-
 	value, ok := evaluated.(string)
 	if !ok {
 		panic(fmt.Sprintf("Expression %s did not evaluate to a string", expr))
@@ -262,10 +262,8 @@ func EvalBool(evaluator ExpressionEvaluator, expr string) (bool, error) {
 			result = t != 0
 		}
 	default:
-		return false, fmt.Errorf("Unable to map return type to boolean for '%s'", expr)
+		return false, fmt.Errorf("Unable to map type '%t' to boolean for '%s'", evaluated, expr)
 	}
-
-	log.Debugf("expression '%s' evaluated to '%t'", nextExpr, result)
 
 	return result, nil
 }
--- a/pkg/runner/expression.go
+++ b/pkg/runner/expression.go
@@ -2,7 +2,6 @@ package runner
 
 import (
 	"fmt"
-	"math"
 	"regexp"
 	"strings"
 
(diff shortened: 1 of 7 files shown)


ZauberNerd:
@ChristopherHX do you remember why you added this debug log in #971 https://github.com/nektos/act/commit/7dbf3fcb96e2d70b3436b718c1f4a9fe4b169a4f#diff-fa62a470d83900f015861d603c59c5e371354fb0cdf36089a274d40ee409029eR139?
Since the error is being returned it will be printed eventually.

ChristopherHX:
I restored the function 1:1 as it was prior running steps after job failure

https://github.com/nektos/act/blob/v0.2.25/pkg/common/executor.go#L139

I'm not the original author of this debug print
> Since the error is being returned it will be printed eventually.

Yes it will **eventually** get printed. I think it good for debugging to always print any error of the then executor.

ZauberNerd:
> Yes it will eventually get printed. I think it good for debugging to always print any error of the then executor.

In my opinion debug logging should be used to provide more information that otherwise isn't visible in order to be able to understand what is happening.
In this case the error is already being printed, so logging the error into the debug log doesn't make much sense.

Below is the test run of https://github.com/nektos/act/pull/971/files#diff-4526efc54b2af950bff29a71c13adadb9f22132374291f264c843cc3cfd51d21

```
$ ./dist/local/act -W ./pkg/runner/testdata/non-existent-action/push.yml 
[non-existent-action/nopanic] 🚀  Start image=ghcr.io/catthehacker/ubuntu:act-latest
[non-existent-action/nopanic]   🐳  docker pull image=ghcr.io/catthehacker/ubuntu:act-latest platform= username= forcePull=false
[non-existent-action/nopanic]   🐳  docker create image=ghcr.io/catthehacker/ubuntu:act-latest platform= entrypoint=["/usr/bin/tail" "-f" "/dev/null"] cmd=[]
[non-existent-action/nopanic]   🐳  docker run image=ghcr.io/catthehacker/ubuntu:act-latest platform= entrypoint=["/usr/bin/tail" "-f" "/dev/null"] cmd=[]
[non-existent-action/nopanic]   🐳  docker exec cmd=[mkdir -m 0777 -p /var/run/act] user=root workdir=
[non-existent-action/nopanic] ⭐  Run ./path/to/non-existent-action
[non-existent-action/nopanic]   ❌  Failure - ./path/to/non-existent-action
[non-existent-action/nopanic] file does not exist
Error: Job 'nopanic' failed
```

KnisterPeter (dismissed):

catthehacker:
> In my opinion debug logging should be used to provide more information that otherwise isn't visible in order to be able to understand what is happening. In this case the error is already being printed, so logging the error into the debug log doesn't make much sense.

you could make it a trace instead debug

EldarKurbanov (approved):

mvrshvl (approved):

mvrshvl (approved):

ZauberNerd:
@ChristopherHX @catthehacker so what's the status here? Should I:
* restore the old behaviour and log the error as debug?
* log the error as trace?
* leave it as is?

In case you'd like to keep a log in the `Then` executor, I'd probably change the log to include a prefix, such as `Error in Then executor: %+v`` or similar, because @KnisterPeter and me had a hard time figuring out where the log came from.

catthehacker:
I don't really have any preference, I'll leave it up to you
Just wanted to give another option due to https://github.com/nektos/act/pull/928

pkg/runner/expression.go
```diff
@@ -262,7 +266,7 @@ func EvalBool(evaluator ExpressionEvaluator, expr string) (bool, error) {
 			result = t != 0
 		}
 	default:
-		return false, fmt.Errorf("Unable to map return type to boolean for '%s'", expr)
+		return false, fmt.Errorf("Unable to map type '%t' to boolean for '%s'", evaluated, expr)
```
ChristopherHX, line 269: I thought the default case is impossible to reach for a conforming evaluator.
Any Object != null => true, except empty string
- `if: github.event.pull_request` ( true if PR )
- `if: fromJSON('{}')` ( always true )

<Details>
<Summary>This weird workflow only fails in act, the last step is expected to run</Summary>

```yaml
on:
- push
- pull_request
jobs:
  m:
    runs-on:
    - ubuntu-latest
    steps:
    - run: |
                echo PR!
      shell: bash
      if: |
                github.event.pull_request
    - run: |
                echo Push!
      shell: bash
      if: |
                !github.event.pull_request
    - run: |
                echo empty obj!
      shell: bash
      if: |
                fromJSON('{}')
```

</Details>

ChristopherHX (approved):
Works: I'm ok with removing the debug print. You won't need to address my review comment, which is just an example ( and a bug in act ) why I prefer to use actions/runner locally.

In a later change, we might want to connect this error to an actual job.
```
Error:   ❌  Error in if-expression: "if: fromJSON('{}')||github.event.pull_request || !github.event.pull_request
" (Unable to map type 'map[]' to boolean for 'fromJSON('{}')||github.event.pull_request || !github.event.pull_request
')
```

pkg/runner/expression.go
ZauberNerd: Good catch, I'll update the function to handle maps.

KnisterPeter (approved):
Nice

cplee:
@ZauberNerd - is this a breaking change? looks like the tests had to be updated from `"` -> `'`...any concern with that having impact on users?

ZauberNerd:
> @ZauberNerd - is this a breaking change? looks like the tests had to be updated from `"` -> `'`...any concern with that having impact on users?

These workflows are invalid according to GitHub and strings can only use single quotes: https://docs.github.com/en/actions/learn-github-actions/expressions#literals

Here's a workflow run of the "issue-597" workflow: https://github.com/ZauberNerd/act-test/actions/runs/1889956547
```
[Invalid workflow file: .github/workflows/test.yml#L17](https://github.com/ZauberNerd/act-test/actions/runs/1889956547/workflow)
The workflow is not valid. .github/workflows/test.yml (Line: 17, Col: 13): Unexpected symbol: '"endsWith'. Located at position 1 within expression: "endsWith('Hello world', 'ld')"
```

Also, it doesn't look like they worked on master either, just the error while parsing the `if` condition didn't lead to a non-zero exit-code:
```
$ ./dist/local/act -W ./pkg/runner/testdata/issue-597/spelling.yaml 
[issue-597/my_first_job] 🚀  Start image=ghcr.io/catthehacker/ubuntu:act-latest
[issue-597/my_first_job]   🐳  docker pull image=ghcr.io/catthehacker/ubuntu:act-latest platform= username= forcePull=false
[issue-597/my_first_job]   🐳  docker create image=ghcr.io/catthehacker/ubuntu:act-latest platform= entrypoint=["/usr/bin/tail" "-f" "/dev/null"] cmd=[]
[issue-597/my_first_job]   🐳  docker run image=ghcr.io/catthehacker/ubuntu:act-latest platform= entrypoint=["/usr/bin/tail" "-f" "/dev/null"] cmd=[]
[issue-597/my_first_job]   🐳  docker exec cmd=[mkdir -m 0777 -p /var/run/act] user=root workdir=
[issue-597/my_first_job]   ❌  Error in if: expression - My first true step
```

With these changes it looks like this:
```
$ ~/act/dist/local/act -W ./issue-597/spelling.yaml 
[issue-597/my_first_job] 🚀  Start image=ghcr.io/catthehacker/ubuntu:act-latest
[issue-597/my_first_job]   🐳  docker pull image=ghcr.io/catthehacker/ubuntu:act-latest platform= username= forcePull=false
[issue-597/my_first_job]   🐳  docker create image=ghcr.io/catthehacker/ubuntu:act-latest platform= entrypoint=["/usr/bin/tail" "-f" "/dev/null"] cmd=[]
[issue-597/my_first_job]   🐳  docker run image=ghcr.io/catthehacker/ubuntu:act-latest platform= entrypoint=["/usr/bin/tail" "-f" "/dev/null"] cmd=[]
[issue-597/my_first_job]   🐳  docker exec cmd=[mkdir -m 0777 -p /var/run/act] user=root workdir=
[issue-597/my_first_job] ⭐  Run My first true step
[issue-597/my_first_job]   ☁  git clone 'https://github.com/actions/hello-world-javascript-action' # ref=main
[issue-597/my_first_job]   🐳  docker cp src=/home/ubuntu/.cache/act/actions-hello-world-javascript-action@main/ dst=/var/run/act/actions/actions-hello-world-javascript-action@main/
[issue-597/my_first_job]   🐳  docker exec cmd=[mkdir -p /var/run/act/actions/actions-hello-world-javascript-action@main/] user= workdir=
[issue-597/my_first_job]   🐳  docker exec cmd=[node /var/run/act/actions/actions-hello-world-javascript-action@main/index.js] user= workdir=
| Hello Renst the Octocat!
[issue-597/my_first_job]   ⚙  ::set-output:: time=22:26:32 GMT+0000 (GMT)
| The event payload: {}
[issue-597/my_first_job]   ✅  Success - My first true step
```

ChristopherHX (approved):
I also wouldn't classify the test workflow corrections as a breaking change, since it never used to work on github. I even blindly duplicated the invalid if in the composite test, in the past assuming these were valid.
_I would say deleting container on workflow failure was a breaking change introduced after 0.2.25, no tests no ci failure._

cplee (approved):

cplee:
Thanks @ZauberNerd  - solid answer :)

PR #898, opened 2021-11-22
feat: allow existing logger from context

ZauberNerd:
We should reuse an existing context logger if in test context.
This will allow test to setup act with a null logger to assert
log messages.

PR #861, opened 2021-11-02
Implement support for pre-/post-actions

ZauberNerd:
This PR needs https://github.com/nektos/act/pull/898 and https://github.com/nektos/act/pull/901 to be merged first.

Implemented here are:
- `pre`, `pre-if` for remote JavaScript actions
- `post`, `post-if` for local and remote JavaScript actions

PR #859, opened 2021-11-02
feat: allow existing logger from context

KnisterPeter:
We should reuse an existing context logger if available.
This will allow test to setup act with a null logger to assert log messages.

PR #1013, opened 2022-02-24
fix: remove unused TestContext

robertkowalski:
The function `WithTestContext` is never called in the project, and
there is just one usage of `TestContext`, inside one of our loggers


ZauberNerd:
A bit of context: We've introduced the `TestContext` here: https://github.com/nektos/act/pull/859, which then got reverted: https://github.com/nektos/act/pull/890 and rewritten: https://github.com/nektos/act/pull/898
It was initially introduced to allow assertions on log output in our first pre/post PR: https://github.com/nektos/act/pull/861/files#diff-1f71c031405f1ee099e0d3fc42613897dcb824213ab0a09321420b4e9ecb0cddR294-R381

KnisterPeter (approved):

Issue #706, opened 2021-05-25
Enhancement: Use act as a self-hosted runner for github actions


ChristopherHX:
[Some people](https://github.com/actions/runner/issues/243) want to have more platform support for self-hosted runners than dotnet provides. What about reusing the runner of act to create a self-hosted runner written in go?

I have written a proof of concept [prototype](https://github.com/ChristopherHX/github-actions-act-runner) which (partially) implements github's / microsoft's proprietary azure pipelines protocol and executes the job via act.

Does this fit into the goals of act?
I guess this breaks
> "Think globally, act locally"

from the ReadMe.

I thinks this improves compatibility for workflows tested against act, because less lines of the workflow are interpreted by github.

My problems using act as library are:
- No way to disable localcheckout
- No way to provide the repository owner and name without a checkout
- No way to use an existing github context (needs depends on an exitsting PR)
- Cannot fully instantiate a runcontext from scratch (outside of the act module), I only want to run one job at a time

What works
- translating steps from githubs job request into acts step model
- live logs on github.com
- logs foreach step are uploaded to github.com

Sample Test Run of my prototype:
![Webaufnahme_25-5-2021_22519_github com](https://user-images.githubusercontent.com/44845461/119569364-f22e1e80-bdae-11eb-8851-4c8cbdf8e3f3.jpeg)

_My prototype currently doesn't have a cli parser and needs small code changes to be registered against github or my own [actions server](https://github.com/ChristopherHX/runner.server)_

igagis:
I'm trying @ChristopherHX 's runner at the moment, it looks very promising for my needs.

catthehacker:
I don't see it happening in a near future. I assigned us to this but it's not something that is planned to be implemented in `act`.
It would be nice to see if there is any demand for it but `act` is mostly useful to *eliminate* the need to trigger GitHub Actions.

I've also explained [here](https://github.com/cli/cli/issues/3585#issuecomment-847849359) why it's not recommended *for now* to do such things.
The ultimate priority for now is to match full compliance with GitHub Actions workflow behaviour, since we still have a lot of issues with that (e.g.: `env:` not properly parsed).

igagis:
> It would be nice to see if there is any demand for it but act is mostly useful to eliminate the need to trigger GitHub Actions.

I'm currently using @ChristopherHX 's runner based on `act` due to some bug in official github runner, and it doesn't seem like they will fix it anytime soon, see https://github.com/actions/runner/issues/1168.

[...]

PR #1015, opened 2022-02-25
Feature: Integrate github-act-runner into act

ChristopherHX:
This PR alone give a low self-hosted runner expierence. See https://github.com/nektos/act/issues/706#issuecomment-1047058761 for required backports of my fork. At the time of writing it is better to use my seperate project https://github.com/christopherhx/github-act-runner.

## Known issues
- the existence of an `.actrc` file leads to unknown flag errors
- you have to specify the `container` property of a job, since native act has no no-container mode.

## Added subcommands
- `configure` setup your connection to github self-hosted runner endpoint.
- `run` listen and run jobs requests of github actions
  - live logs are streamed into the github ui
- `remove` reverts configure and remove the agent from github

More details here: https://github.com/christopherhx/github-act-runner

Resolves https://github.com/nektos/act/issues/706

### DIFF
--- a/cmd/agent.go
+++ b/cmd/agent.go
@@ -1,0 +1,66 @@
+package cmd
+
+import (
+	"github.com/nektos/act/pkg/agent"
+	"github.com/spf13/cobra"
+	"os"
+	"runtime"
+)
+
+func addGitHubAgentCommand(rootCmd *cobra.Command) {
+	config := &agent.ConfigureRunner{}
+	run := &agent.RunRunner{}
+	remove := &agent.RemoveRunner{}
+	var cmdConfigure = &cobra.Command{
+		Use:   "configure",
+		Short: "Configure your self-hosted runner",
+		Args:  cobra.MaximumNArgs(0),
+		Run: func(cmd *cobra.Command, args []string) {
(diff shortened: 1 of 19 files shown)


ZauberNerd:
I haven't read through this PR in detail, but a few things I've noticed:
* act doesn't have upper-case in file names.
* Maybe we should first try to fix: https://github.com/nektos/act/blob/master/cmd/root.go#L73 which leads to act not allowing other commands than the root command (this also break bash completion, which cobra should support). We could probably switch to https://github.com/spf13/viper which would integrate well with cobra.

catthehacker:
> * We could probably switch to [spf13/viper](https://github.com/spf13/viper) which would integrate well with cobra.

https://github.com/nektos/act/pull/479

ChristopherHX:
> act doesn't have upper-case in file names.

You are right..., probably some go best practice. I converted the protocol package to lowercase filename + underscores.

> Maybe we should first try to fix: https://github.com/nektos/act/blob/master/cmd/root.go#L73

Yes we need to do something about this, since an error about an `-P ubuntu-latest=...` in `.actrc` has confused me a bit.

Still a lot of work to be done, before the linter is green...

PR #1026, opened 2022-03-03
feat: add json logger output

KnisterPeter:
This will allow to format log output as json.
This is helpful in cases where act is not executed on a 'local' machine.

### DIFF
--- a/pkg/runner/logger.go
+++ b/pkg/runner/logger.go
@@ -41,11 +41,19 @@ func init() {
 func WithJobLogger(ctx context.Context, jobName string, secrets map[string]string, insecureSecrets bool, masks *[]string) context.Context {
 	mux.Lock()
 	defer mux.Unlock()
-	formatter := new(stepLogFormatter)
-	formatter.color = colors[nextColor%len(colors)]
-	formatter.secrets = secrets
-	formatter.insecureSecrets = insecureSecrets
-	formatter.masks = masks
+
+	var formatter logrus.Formatter
+	if isJSONFormatter(ctx) {
+		formatter = &jobLogJSONFormatter{
+			formatter: &logrus.JSONFormatter{},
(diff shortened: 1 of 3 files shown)


pkg/runner/logger.go
```diff
@@ -64,30 +72,51 @@ func WithJobLogger(ctx context.Context, jobName string, secrets map[string]strin
...
+func isJSONFormatter(ctx context.Context) bool {
+	if ctxLogger := common.Logger(ctx); ctxLogger != nil {
+		if logger, ok := ctxLogger.(*logrus.Logger); ok {
+			_, ok := logger.Formatter.(*logrus.JSONFormatter)
+			return ok
+		}
+	}
+
+	_, ok := logrus.StandardLogger().Formatter.(*logrus.JSONFormatter)
+	return ok
```
ChristopherHX, line 84: This reads like a hack...

ChristopherHX (dismissed):
Nice to have, it works :)

pkg/runner/logger.go
KnisterPeter: Any idea how to check what logger the context logger has?

ChristopherHX: I would add a parameter to WithJobLogger, then pass the whole runcontext.Config object or just the jsonLogger field.
We can also keep it like this, I have approved your changes already.

ChristopherHX (approved):
Note: I wondered why I see a `Error: <errmsg>` line after a failure, which ignored the logrus logger / json formatter. It seems that cobra cli prints it if we return the error back to the library, here the callback which causes the printed line when error != nil https://github.com/nektos/act/blob/3db3c737230adc457e08a8d60b34458977d01e5c/cmd/root.go#L158

ZauberNerd (approved):

cplee (approved):

PR #1027, opened 2022-03-03
feat: add step name to logger field

KnisterPeter:
This change does add the step name to the logger fields. This does
not change the output for our users, but for the json logger, it
does make each step output traceable.

### DIFF
--- a/pkg/runner/logger.go
+++ b/pkg/runner/logger.go
@@ -64,6 +64,27 @@ func WithJobLogger(ctx context.Context, jobName string, secrets map[string]strin
 	return common.WithLogger(ctx, rtn)
 }
 
+func withStepLogger(ctx context.Context, stepName string) context.Context {
+	mux.Lock()
+	defer mux.Unlock()
+
+	stepLogger := logrus.New()
+
+	parentLogger := common.Logger(ctx)
+
+	if logger, ok := parentLogger.(*logrus.Logger); ok {
+		stepLogger.SetFormatter(logger.Formatter)
+		stepLogger.SetOutput(logger.Out)
+		stepLogger.SetLevel(logger.GetLevel())
+	} else {
+		stepLogger.SetOutput(os.Stdout)
+		stepLogger.SetLevel(logrus.GetLevel())
+	}
+
+	rtn := stepLogger.WithFields(logrus.Fields{"step": stepName})
+	return common.WithLogger(ctx, rtn)
+}
+
 type stepLogFormatter struct {
 	color           int
 	secrets         map[string]string
--- a/pkg/runner/job_executor.go
+++ b/pkg/runner/job_executor.go
@@ -37,15 +37,18 @@ func newJobExecutor(info jobInfo) common.Executor {
 		}
 		stepExec := info.newStepExecutor(step)
 		steps = append(steps, func(ctx context.Context) error {
-			err := stepExec(ctx)
-			if err != nil {
-				common.Logger(ctx).Errorf("%v", err)
-				common.SetJobError(ctx, err)
-			} else if ctx.Err() != nil {
-				common.Logger(ctx).Errorf("%v", ctx.Err())
-				common.SetJobError(ctx, ctx.Err())
-			}
-			return nil
+			stepName := step.String()
+			return (func(ctx context.Context) error {
+				err := stepExec(ctx)
+				if err != nil {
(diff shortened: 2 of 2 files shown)


pkg/runner/logger.go
```diff
@@ -64,6 +64,27 @@ func WithJobLogger(ctx context.Context, jobName string, secrets map[string]strin
...
+	if logger, ok := parentLogger.(*logrus.Logger); ok {
+		stepLogger.SetFormatter(logger.Formatter)
+		stepLogger.SetOutput(logger.Out)
+		stepLogger.SetLevel(logger.GetLevel())
+	} else {
+		stepLogger.SetOutput(os.Stdout)
+		stepLogger.SetLevel(logrus.GetLevel())
+	}
+
+	rtn := stepLogger.WithFields(logrus.Fields{"step": stepName})
```
ChristopherHX, line 84: > This does not change the output for our users

I'm afraid, this change forces output to stdout even if my usecase for nektos/act wants redirection.
During debugging I always branched into the second case.

The following change does work for me without side effects
```suggestion
	parentLogger := common.Logger(ctx)
	rtn := parentLogger.WithFields(logrus.Fields{"step": stepName})
```

ChristopherHX (changes requested):
This doesn't seem to work for me.
Also every step no longer prints via the job logger.

pkg/runner/logger.go
KnisterPeter: That would add the step to the parent logger which is logically wrong. When a step is done, no logger should have this field anymore. That's why I created a new logger as a child.

KnisterPeter: Can you please be a bit more specific on your test case? The default logrus logger outputs to stdout, if there is no logger in the context, that should be fine.

KnisterPeter:
> This doesn't seem to work for me.
> Also every step no longer prints via the job logger.

Right, the step should log through the step-logger.
Do you use a different job-logger?

ChristopherHX:
> Do you use a different job-logger?

Yes I use a custom one, since I send the log to github actions service no logging to stdout.

pkg/runner/logger.go
ChristopherHX: > The default logrus logger outputs to stdout

No I changed it in my code see https://github.com/nektos/act/pull/1015

> That would add the step to the parent logger which is logically wrong

I need to check that later.

KnisterPeter:
So, the problem is that a different logger is used. Can't you set the output on the root logger? That should then be used by all context loggers and redirect as you wish

ChristopherHX:
> Can't you set the output on the root logger?

No, this PR overrides everything I set:
```
stepLogger.SetOutput(os.Stdout)
```

**The other if case is always skipped during debugging of this PR**

KnisterPeter:
Since I don't know your code, I cannot even think of another way to do it. Can you reference your code lines here?

pkg/runner/logger.go
ChristopherHX: > That would add the step to the parent logger which is logically wrong

**No it doesn't** parentLogger.WithFields doesn't alter parentLogger.
Also common.WithLogger also doesn't alter the previous context.

Why do you think it does alter  the parent?

ChristopherHX:
> Can you reference your code lines here?

https://github.com/ChristopherHX/act/blob/cd5e0e29fd725880b4cf58f315f6c791450bf266/pkg/agent/main.go#L1570-L1587

pkg/runner/logger.go
KnisterPeter: I think my misunderstanding was that changing the logger is mutating, but after reading the logrus source I noticed it isn't.
So I don't have to change the logger at all, but just at fields and put that enrty into the context.
I'll update that and it should work for you as well, right?

ChristopherHX: > I'll update that and it should work for you as well, right?

Yes

KnisterPeter: Updated it, I hope it does work now.

ChristopherHX (approved):
All fine now, I can confirm that you no longer notice the change without a json logger or a debugger.

ZauberNerd (approved):

cplee (dismissed):

ZauberNerd (approved):

PR #1030, opened 2022-03-09
fix: coerce booleans to numbers for comparison in exprparser

ZauberNerd:


KnisterPeter (approved):

ZauberNerd:
@cplee here is an example workflow that fails:

```
~/act $ cat test.yml 
on: push
jobs:
  job-1:
    runs-on: ubuntu-latest
    steps:
      - run: echo "hello"
        if: ${{ true == true }}
~/act $ ./dist/local/act -W ./test.yml 
[test.yml/job-1] 🚀  Start image=ghcr.io/catthehacker/ubuntu:act-latest
[test.yml/job-1]   🐳  docker pull image=ghcr.io/catthehacker/ubuntu:act-latest platform= username= forcePull=false
[test.yml/job-1]   🐳  docker create image=ghcr.io/catthehacker/ubuntu:act-latest platform= entrypoint=["/usr/bin/tail" "-f" "/dev/null"] cmd=[]
[test.yml/job-1]   🐳  docker run image=ghcr.io/catthehacker/ubuntu:act-latest platform= entrypoint=["/usr/bin/tail" "-f" "/dev/null"] cmd=[]
[test.yml/job-1]   🐳  docker exec cmd=[mkdir -m 0777 -p /var/run/act] user=root workdir=
[test.yml/job-1]   ❌  Error in if-expression: "if: ${{ true == true }}" (TODO: evaluateCompare not implemented reflect.Value)
Error: Job 'job-1' failed
```

The problem was, that we only coerced to numbers when the inputs are differet: https://github.com/nektos/act/blob/b681c31bee5790628535762d2df796a71990bfd4/pkg/exprparser/interpreter.go#L298-L305 but we forgot to implement comparison for booleans: https://github.com/nektos/act/blob/b681c31bee5790628535762d2df796a71990bfd4/pkg/exprparser/interpreter.go#L307-L329

cplee (approved):

ZauberNerd:
Ah, yes. I added those tests, to ensure that it doesn't fail for other types of comparison. There is also one test for `!=` which is a bit more useful than `>` or `<` ;)

PR #1049, opened 2022-03-16
feat: add flag to always run the checkout action

robertkowalski:
act has a feature that skips the checkout action to do a remote
checkout when a local checkout exists. in some cases, e.g. when
running act in a CI, you always want to clone the repository.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -71,6 +71,7 @@ func Execute(ctx context.Context, version string) {
 	rootCmd.PersistentFlags().StringVarP(&input.githubInstance, "github-instance", "", "github.com", "GitHub instance to use. Don't use this if you are not using GitHub Enterprise Server.")
 	rootCmd.PersistentFlags().StringVarP(&input.artifactServerPath, "artifact-server-path", "", "", "Defines the path where the artifact server stores uploads and retrieves downloads from. If not specified the artifact server will not start.")
 	rootCmd.PersistentFlags().StringVarP(&input.artifactServerPort, "artifact-server-port", "", "34567", "Defines the port where the artifact server listens (will only bind to localhost).")
+	rootCmd.PersistentFlags().BoolVarP(&input.noSkipCheckout, "no-skip-checkout", "", false, "Do not skip actions/checkout")
 	rootCmd.SetArgs(args())
 
 	if err := rootCmd.Execute(); err != nil {
@@ -287,6 +288,7 @@ func newRunCommand(ctx context.Context, input *Input) func(*cobra.Command, []str
 			AutoRemove:            input.autoRemove,
 			ArtifactServerPath:    input.artifactServerPath,
 			ArtifactServerPort:    input.artifactServerPort,
+			NoSkipCheckout:        input.noSkipCheckout,
 		}
 		r, err := runner.New(config)
 		if err != nil {
--- a/pkg/runner/step_context.go
+++ b/pkg/runner/step_context.go
@@ -92,7 +92,7 @@ func (sc *StepContext) Executor(ctx context.Context) common.Executor {
 		remoteAction.URL = rc.Config.GitHubInstance
 
 		github := rc.getGithubContext()
-		if remoteAction.IsCheckout() && isLocalCheckout(github, step) {
+		if remoteAction.IsCheckout() && isLocalCheckout(github, step) && !rc.Config.NoSkipCheckout {
 			return func(ctx context.Context) error {
 				common.Logger(ctx).Debugf("Skipping local actions/checkout because workdir was already copied")
 				return nil
--- a/cmd/input.go
+++ b/cmd/input.go
@@ -39,6 +39,7 @@ type Input struct {
 	artifactServerPath    string
 	artifactServerPort    string
 	jsonLogger            bool
+	noSkipCheckout        bool
 }
 
 func (i *Input) resolve(path string) string {
--- a/pkg/runner/run_context.go
+++ b/pkg/runner/run_context.go
@@ -708,6 +708,10 @@ func setActionRuntimeVars(rc *RunContext, env map[string]string) {
 }
 
 func (rc *RunContext) localCheckoutPath() (string, bool) {
+	if rc.Config.NoSkipCheckout {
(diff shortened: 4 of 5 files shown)


robertkowalski:
@ChristopherHX heard through the grapevine's that you also would like to have such a feature

pkg/runner/step_context.go
```diff
@@ -91,6 +91,13 @@ func (sc *StepContext) Executor(ctx context.Context) common.Executor {
 
 		remoteAction.URL = rc.Config.GitHubInstance
 
+		if remoteAction.IsCheckout() && rc.Config.DisableLocalCheckout {
```
ZauberNerd, line 94: That's not correct. What I meant, when I talked with you about this was:
https://github.com/nektos/act/blob/b7f0a7de234ef2c3065a41e4a5d0b59d0f128828/pkg/runner/step_context.go#L103 returns a `nil` executor instead of actually executing the `actions/checkout` in act.
So I meant a flag to disable this "skipping" behavior of act, so that when the flag is set, act will always execute `actions/checkout` when it is present in a workflow just like any other action.

ZauberNerd (changes requested):

pkg/runner/step_context.go
ZauberNerd: The default should continue to be that act copies / bind-mounts the workdir into the container, so that it is easy to develop and debug actions and workflows locally.
But in some cases, you might want to actually run the `actions/checkout` action instead of letting act handle the copy/mount.

ChristopherHX: I agree with @ZauberNerd. I have an implementation for forceremotecheckout, but my backports got stuck a while ago https://github.com/ChristopherHX/act/blob/54588b02c3e0cef063635341c0163553fe571dd5/pkg/runner/runner.go#L54

FYI My backport progress is blocked by https://github.com/nektos/act/pull/1004

cmd/root.go
```diff
@@ -71,6 +71,7 @@ func Execute(ctx context.Context, version string) {
 	rootCmd.PersistentFlags().StringVarP(&input.githubInstance, "github-instance", "", "github.com", "GitHub instance to use. Don't use this if you are not using GitHub Enterprise Server.")
 	rootCmd.PersistentFlags().StringVarP(&input.artifactServerPath, "artifact-server-path", "", "", "Defines the path where the artifact server stores uploads and retrieves downloads from. If not specified the artifact server will not start.")
 	rootCmd.PersistentFlags().StringVarP(&input.artifactServerPort, "artifact-server-port", "", "34567", "Defines the port where the artifact server listens (will only bind to localhost).")
+	rootCmd.PersistentFlags().BoolVarP(&input.disableLocalCheckout, "disable-local-checkout", "", false, "disables the local checkout for actions")
```
catthehacker, line 74:
```suggestion
	rootCmd.PersistentFlags().BoolVarP(&input.disableLocalCheckout, "no-local-checkout", "", false, "disables the local checkout for actions")
```

pkg/runner/step_context.go
robertkowalski: oh i misunderstood @ZauberNerd

robertkowalski:
i think the naming from @ChristopherHX is better -- `force-remote-checkout` -- maybe use this instead?

cmd/root.go
```diff
@@ -71,6 +71,7 @@ func Execute(ctx context.Context, version string) {
 	rootCmd.PersistentFlags().StringVarP(&input.githubInstance, "github-instance", "", "github.com", "GitHub instance to use. Don't use this if you are not using GitHub Enterprise Server.")
 	rootCmd.PersistentFlags().StringVarP(&input.artifactServerPath, "artifact-server-path", "", "", "Defines the path where the artifact server stores uploads and retrieves downloads from. If not specified the artifact server will not start.")
 	rootCmd.PersistentFlags().StringVarP(&input.artifactServerPort, "artifact-server-port", "", "34567", "Defines the port where the artifact server listens (will only bind to localhost).")
+	rootCmd.PersistentFlags().BoolVarP(&input.alwaysRunCheckoutAction, "always-run-checkout-action", "", false, "always run the checkout action")
```
ZauberNerd, line 74: How about `--no-skip-checkout`? Does that make it better?

ZauberNerd (commented):
Except for the long flag name I think this PR looks fine now 🎉

cmd/root.go
catthehacker: Either `--no-skip-checkout` or `--force-remote-checkout` is fine

catthehacker: Flag description should start with capitalised letter and if it could be rephrased a bit, e.g. `Checkout actions remote repository even if it's cached`

catthehacker: Eventually we could introduce flag `--invalidate-actions-cache` which removes cached actions thus making act check them out

ZauberNerd: Hm, but this flag isn't only about other actions, but also about the repository. In our case for example, we're using the full payload that GitHub sends and want act to run the `actions/checkout` to check out the ref specified in the event payload instead of using the local working directory.
So I think something like `Do not skip actions/checkout` as message would be better

catthehacker: ah, I haven't actually looked into diff and from all said before I assumed it's about actions checkouts not `actions/checkout`

catthehacker: Disregard the part about actions checkout then, capitalisation and flag name still applies

ChristopherHX: I actually used `--remote-checkout` as cli flag name outside of act, with `Do not inject localcheckout into your workflows, always use the original actions/checkout` as description. _In this case I actually added an additional step before all actions/checkout steps of all jobs, without modifing the `actions/runner`._

I'm also ok with `--force-remote-checkout` or `--no-skip-checkout`.

robertkowalski: its `--no-skip-checkout` now! :)

ZauberNerd (approved):

KnisterPeter (approved):

catthehacker (approved):

PR #1048, opened 2022-03-16
governance: new issue templates

catthehacker:

Issue #851, opened 2021-10-29
Issue: Invalid mount path error when running act


MattStaffordOmny:
## System information

- Operating System: Windows 10
- Architecture: x64
- Apple M1:
- Docker version: 20.10.8, build 3967b7d
- Docker image used in `act`: ubuntu-latest=catthehacker/ubuntu:act-latest
- `act` version: 0.2.24

## Expected behaviour

act should run

## Actual behaviour

act fails to run and returns the following error:

```
Error: Error response from daemon: invalid mount config for type "volume": invalid mount path: '/toolcache'
```

## Workflow and/or repository

<details>
  <summary>Log</summary>

```none
time="2021-10-29T14:53:29+11:00" level=debug msg="using github ref: refs/heads/master"
[GitHub Actions Demo/Explore-GitHub-Actions]   🐳  docker pull catthehacker/ubuntu:act-latest
time="2021-10-29T14:53:29+11:00" level=debug msg="Image exists? false"
[GitHub Actions Demo/Explore-GitHub-Actions] pulling image 'docker.io/catthehacker/ubuntu:act-latest' ()
[GitHub Actions Demo/Explore-GitHub-Actions] Pulling from catthehacker/ubuntu :: act-latest
[GitHub Actions Demo/Explore-GitHub-Actions] no matching manifest for windows/amd64 10.0.19043 in the manifest list entries
[GitHub Actions Demo/Explore-GitHub-Actions]   🐳  docker create image=catthehacker/ubuntu:act-latest platform= entrypoint=["/usr/bin/tail" "-f" "/dev/null"] cmd=[]
[...]

PR #1056, opened 2022-03-17
feat: add bug-report flag

catthehacker:
image


pkg/container/docker_run.go
```diff
@@ -232,6 +229,21 @@ func GetDockerClient(ctx context.Context) (*client.Client, error) {
 	return cli, err
 }
 
+func GetHostInfo(ctx context.Context) (info types.Info, err error) {
+	var cli *client.Client
+	cli, err = GetDockerClient(ctx)
+	if err != nil {
+		return info, err
+	}
```
ChristopherHX, line 237: Please close the docker connection, after using it.
```suggestion
	cli, err = GetDockerClient(ctx)
	if err != nil {
		return info, err
	}
	defer cli.Close()
```

cmd/root.go
```diff
@@ -111,6 +117,60 @@ func args() []string {
...
+	report += fmt.Sprintf("\tCgroup version:\t%s\n", info.CgroupVersion)
+	report += fmt.Sprintf("\tEngine runtime:\t%s\n", info.DefaultRuntime)
+
+	report += fmt.Sprintf("\tOS:\t\t%s\n", info.OperatingSystem)
+	report += fmt.Sprintf("\tOS type:\t%s\n", info.OSType)
+	report += fmt.Sprintf("\tOS version:\t%s\n", info.OSVersion)
+	report += fmt.Sprintf("\tOS arch:\t%s\n", info.Architecture)
+	report += fmt.Sprintf("\tOS kernel:\t%s\n", info.KernelVersion)
+	report += fmt.Sprintf("\tOS CPU:\t\t%d\n", info.NCPU)
+	report += fmt.Sprintf("\tOS memory:\t%d\n", info.MemTotal)
```
ZauberNerd, line 163: Just a small nitpick, but can we format the bytes human readable?

catthehacker: sure, I have yet to still get rid of `\t`s and `\n`s 😸

cmd/root.go
```diff
@@ -111,6 +117,60 @@ func args() []string {
 	return args
 }
 
+func bugReport(ctx context.Context) error {
+	var containerDaemonPaths = []string{
+		"/var/run/docker.sock",
+		"/var/run/podman/podman.sock",
+		"$HOME/.colima/docker.sock",
```
ZauberNerd, line 124: I guess this could also be: `$XDG_RUNTIME_DIR/docker.sock` in case of rootless docker, right? https://docs.docker.com/engine/security/rootless/#daemon

ZauberNerd: But nevertheless, rootless docker also needs `DOCKER_HOST` to be set

Procorners (approved):

Procorners (approved):

cmd/root.go
catthehacker: done

pkg/runner/runner.go
```diff
@@ -170,7 +172,15 @@ func (runner *runnerImpl) NewPlanExecutor(plan *model.Plan) common.Executor {
 				}
 				pipeline = append(pipeline, common.NewParallelExecutor(maxParallel, stageExecutor...))
 			}
-			return common.NewParallelExecutor(runtime.NumCPU(), pipeline...)(ctx)
+			var ncpu int
+			info, err := container.GetHostInfo(ctx)
+			if err == nil {
```
ChristopherHX, line 177: This doesn't seem to be right
```suggestion
			if err != nil {
```

catthehacker: right, typo

pkg/runner/runner.go
```diff
@@ -170,7 +172,15 @@ func (runner *runnerImpl) NewPlanExecutor(plan *model.Plan) common.Executor {
 				}
 				pipeline = append(pipeline, common.NewParallelExecutor(maxParallel, stageExecutor...))
 			}
-			return common.NewParallelExecutor(runtime.NumCPU(), pipeline...)(ctx)
+			var ncpu int
+			info, err := container.GetHostInfo(ctx)
+			if err != nil {
+				log.Errorf("failed to obtain container engine info: %s", err)
+				ncpu = 1 // sane default?
```
ZauberNerd, line 179: Or `runtime.NumCPU()`?
What's the benefit of asking the docker client for the number of CPUs? Does it take into account resource constraints via cgroups or the scheduler?

cmd/root.go
```diff
@@ -111,6 +115,67 @@ func args() []string {
...
+	sprintf := func(key, val string) string {
+		return fmt.Sprintf("%s:\t\t%s\n", key, val)
+	}
+
+	report := sprintf("act version", version)
+	report += sprintf("GOOS", runtime.GOOS)
+	report += sprintf("GOARCH", runtime.GOARCH)
+	report += sprintf("NumCPU", fmt.Sprint(runtime.NumCPU()))
+
+	if os.Getenv("DOCKER_HOST") == "" {
```
ZauberNerd, line 135: We're printing all found sockets in case `DOCKER_HOST` is not set, but shouldn't we also print the value of `DOCKER_HOST` in case it is set?

pkg/runner/runner.go
ChristopherHX: > Or runtime.NumCPU()?

Once I contributed my hostexecutor, I would prefer:
- `runtime.NumCPU()` here, since I don't want to start the docker daemon to test something on the host.
- `failed to obtain container engine info: %s` should become a debug log, since docker would be optional

Until then, I'm fine with `1` as default and an error message.

> What's the benefit of asking the docker client for the number of CPUs?

A remote docker server might have a different amount of CPUs, but whatever. I never actually used a remote docker server.

catthehacker: > What's the benefit of asking the docker client for the number of CPUs?

Docker host doesn't have to be on the same logical machine as `act`. Docker Desktop on macOS and Windows both run in VMs which can have different CPU count.

cmd/root.go
catthehacker: I can't think of any benefit it will provide us or the user, if `DOCKER_HOST` is set then all is fine

ZauberNerd (approved):

catthehacker:
Accidentally removed `MB` from OS RAM info, do we need that? 🤖

ChristopherHX (dismissed):
I don't need the unit for the amount of ram. `Sockets found` is empty on windows, but that's no problem.

catthehacker:
> I don't need the unit for the amount of ram. `Sockets found` is empty on windows, but that's no problem.

Added Docker engine pipe location for Windows (+ missing `MB`)

catthehacker:
> We might want to consider to print an error if we detect `OS type: windows` and tell the user how to switch docker desktop to linux container.

I haven't seen anyone have issues with that so for now I'd rather omit it

catthehacker:
Windows:
```
act version:
GOOS:                   windows
GOARCH:                 amd64
NumCPU:                 24
Docker host:            ssh://[REDACTED]
Sockets found:
        \\.\pipe\docker_engine
Docker Engine:
        Engine version:        20.10.9
        Engine runtime:        runc
        Cgroup version:        1
        OS:                    Alpine Linux v3.15
        OS type:               linux
        OS version:            3.15.0
        OS arch:               x86_64
        OS kernel:             5.10.59-0-lts
        OS CPU:                4
        OS memory:             16017 MB
        Security options:
                name=seccomp,profile=default
```

ChristopherHX:
>> We might want to consider to print an error if we detect OS type: windows and tell the user how to switch docker desktop to linux container.
>
>I haven't seen anyone have issues with that so for now

I saw it once
https://github.com/nektos/act/issues/851

catthehacker:
> I saw it once #851

I rather prefer to fix it via better documentation. If only a single person raised this issue so far, it doesn't make sense to clutter it in this output.
After all main goal for `--bug-report` is to provide easy to copy information for #1048

ChristopherHX (dismissed):

KnisterPeter (approved):

ZauberNerd (approved):

ChristopherHX (approved):

PR #1060, opened 2022-03-20
feat: podman CI

catthehacker:


ntindle:
is that failure in ci expected?

XVilka:
Are there any plans to revive this? Logs are expired and it's now unclear what exactly was broken/didn't pass.

PR #1066, opened 2022-03-21
Fix: panic: reflect: slice index out of range

ChristopherHX:
panics are very bad


catthehacker (dismissed):

pkg/exprparser/interpreter.go
```diff
@@ -170,6 +170,9 @@ func (impl *interperterImpl) evaluateIndexAccess(indexAccessNode *actionlint.Ind
 	case reflect.Int:
 		switch leftValue.Kind() {
 		case reflect.Slice:
+			if rightValue.Int() >= int64(leftValue.Len()) {
+				return nil, nil
+			}
```
KnisterPeter, line 175: Please add a test case for this

KnisterPeter (changes requested):

ZauberNerd (approved):

pkg/exprparser/interpreter.go
ZauberNerd: Sure, sounds like a good idea. Whatever you can do to help us get closer to GHA in terms of compatibility would be awesome 🚀 
But let's do that then in a separate PR.

catthehacker (approved):

KnisterPeter (approved):

PR #1070, opened 2022-03-23
fix: set default version string

catthehacker:
for when people don't specify version via -ldflags


KnisterPeter (approved):

ChristopherHX:
> Manually bump after tagging next release

Add this to the workflow, which creates the new release on push a tag?
That workflow has the previous tag `${{github.ref_name}}`, which could be incremented by a script then automatically create a PR with github actions.

ChristopherHX (approved):

ZauberNerd (approved):

PR #1074, opened 2022-03-24
fix: copy ignored tracked files

### LINKED ISSUE #912: Issue: actions/checkout doesn't copy git tracked but gitignored files
The general behavior of `.gitignore` doesn't prevent files that are tracked from being checked out. It just prevents from from being updated or added.

I'd expect that act not copy over untracked files matched by `.gitignore`, but I'd expect it to include tracked files that are matched by `.gitignore`.

## System information

- Operating System: Linux & macOS
- Architecture: x64
- Apple M1: no
- Docker version: on macOS: 20.10.8 (not sure about the version used by GHA)
- Docker image used in `act`: catthehacker/ubuntu:act-latest
- `act` version: 0.2.25

## Expected behaviour

Files listed in `.gitignore` should be present in the checked out repository

## Actual behaviour

Files listed in `.gitignore` are not present

## Workflow and/or repository

https://github.com/jsoref/ignored/commit/bfbee7f061955141522839cbb6abf690678adfb0
https://github.com/jsoref/ignored/runs/4398686247?check_suite_focus=true

https://github.com/jsoref/ignored/commit/5e20813c26869ea46a07a4ec03481c911377979c
https://github.com/jsoref/ignored/runs/4398859601?check_suite_focus=true

## Steps to reproduce

1. Push https://github.com/jsoref/ignored/commit/bfbee7f061955141522839cbb6abf690678adfb0 to a GitHub repo
2. Check the action log
3. Push https://github.com/jsoref/ignored/commit/5e20813c26869ea46a07a4ec03481c911377979c to the GitHub repo
4. Check the action log

## `act` output

<details>
  <summary>Log</summary>

```none
[What do I have/Introspection] Skipping local actions/checkout because workdir was already copied
```

</details>

### LINKED ISSUE #1002: Issue: cp: cannot stat 'js/config.dist.ts': No such file or directory
## System information

    - Operating System: Arch Linux
    - Architecture:x64 (64-bit)
    - Apple M1: no
    - Docker version: 20.10.12
    - `act` version: 0.2.25

## Expected behaviour

The workflow, which works on GitHub actions, should work locally with act.

## Actual behaviour
It throws an error  `cp: cannot stat 'js/config.dist.ts': No such file or directory`.
However that file exists at `repo/js/config.dist.ts`.

## Workflow and/or repository

<https://raw.githubusercontent.com/snikproject/snik-graph/da80e8c2ab07a5bf68fb0575706ef4ad5d23a27a/.github/workflows/build.yml>

<details>
  <summary>workflow</summary>

```yml
name: build

on:
  workflow_dispatch:
  push:
    branches:
      - master

jobs:
  build: 
    runs-on: ubuntu-latest
    steps:
      - name: Checkout code
        uses: actions/checkout@v2

      - uses: actions/setup-node@v2

      - name: Cache node modules
        uses: actions/cache@v2
        with:
          path: ~/node_modules
          key: cache-${{ hashFiles('**/package-lock.json') }}

      - name: Install
        run: |
          cp js/config.dist.ts js/config.ts
          npm install --ignore-scripts --production --no-audit

      - name: Test
        run: |
          npm run test

      - name: Build
        run: |
          npm run build
          sed -i "s|/assets|./assets|" dist/index.html
          cp -r html dist
          cp -r img dist

      - name: Doc
        run: |
          npm run doc
          cp -r doc dist 

      - name: Deploy to GitHub Pages
        uses: JamesIves/github-pages-deploy-action@4.1.5
        with:
          branch: gh-pages
          folder: dist
```

</details>

## Steps to reproduce

      1. `git clone git@github.com:snikproject/snik-graph.git`
      2. Enter cloned repo directory
      3. git checkout 6029745341016d37a87a68c82e2bd48262364ad6
      4. Run `act .github/workflows/build.yml -j build`

## The act log is in the comment because of a GitHub content length error

ChristopherHX:
Resolves #912
Resolves #1002
Closes #949

What I have tested:
- copy without git directory in current folder => old behavior
- copy with git directory, now includes all tracked files even those which are ignored
- first level of submodules
- you can now cancel the copydir by pressing ctrl+c

~~I'm not shure how to write tests for this change.~~


ChristopherHX:
@jsoref You might want to check if this fits you usecase.

`act -w` should be changed in a later PR.

pkg/container/docker_run.go
```diff
@@ -599,69 +604,98 @@ func (cr *containerReference) copyDir(dstPath string, srcPath string, useGitIgno
 			ignorer = gitignore.NewMatcher(ps)
 		}
 
-		err = filepath.Walk(srcPath, func(file string, fi os.FileInfo, err error) error {
-			if err != nil {
-				return err
+		var getCallback func(submodulePath []string) func(file string, fi fs.FileInfo, err error) error
+		getCallback = func(submodulePath []string) func(file string, fi os.FileInfo, err error) error {
```
ZauberNerd, line 608: This function does not use `cr *containerReference`, so for readability and testability you could extract it into its own function outside of `copyDir`.
And for the test maybe it is worth looking into https://pkg.go.dev/testing/fstest to create an in-memory FS. We're doing something like that for the artifact server tests: https://github.com/nektos/act/blob/master/pkg/artifacts/server_test.go#L24-L52

ChristopherHX: Good idea to extract it in it's own function, also may extract the tar usage into an interface to be reusable for a non docker copydir. ( run on host )

But how would I checkout a git repository in fstest? Only billyfs works with go-git if I'm not mistaken.

ZauberNerd: Ah, I haven't thought about the "how" too much but just remembered that fs stuff can be mocked with an in-memory fs.
Maybe it'd be worth to look into go-git to see how they test things?
Or if a mocked fs doesn't work you could probably create a temp directory and test in there?

Btw, with this change it is possible to run the `checks.yml` test to test act itself, which previously failed because of `.env` and `.secrets` files 🎉 
```
$ ./dist/local/act -W ./.github/workflows/checks.yml pull_request -j test-linux
```

pkg/container/docker_run.go
```diff
@@ -579,8 +578,8 @@ func (cr *containerReference) copyDir(dstPath string, srcPath string, useGitIgno
 			return err
 		}
 		log.Debugf("Writing tarball %s from %s", tarFile.Name(), srcPath)
-		defer tarFile.Close()
 		defer os.Remove(tarFile.Name())
+		defer tarFile.Close()
```
ChristopherHX, line 582: We need to close the tarFile before removing it, since we use defer here the order were wrong before my change

pkg/container/docker_run.go
ChristopherHX: > Btw, with this change it is possible to run the checks.yml test to test act itself

No idea if this still works :), after adding filesystem and output interfaces. _I don't want to test that myself._

I added an in memory test with go-billy and go-git.
Covers the two fixed bugs, no submodules.

pkg/container/docker_run.go
KnisterPeter: I think it would be more readable to create a defered inline func like this:

```go
defer (func () {
  tarFile.Close()
  os.Remove(tarFile.Name())
})()
```

catthehacker:
```suggestion
		defer func(tarFile *os.File) {
			name := tarFile.Name()
			err := tarFile.Close()
			if err != nil {
				logger.Error(err)
			}
			err = os.Remove(name)
			if err != nil {
				logger.Error(err)
			}
		}(tarFile)
```

ZauberNerd (commented):
Overall pretty good. I'll approve once the requested changes by @KnisterPeter / @catthehacker are committed.

pkg/container/file_collector_test.go
```diff
@@ -0,0 +1,117 @@
...
+			}
+		}
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (mfs *memoryFs) Walk(root string, fn filepath.WalkFunc) error {
```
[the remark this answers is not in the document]
ChristopherHX: If I use lowercase here I get:
```
cannot use &(memoryFs literal) (value of type *memoryFs) as fileCollectorFs value in struct literal: missing method Open
```
Then I would need to change the interface functions to lowercase too, but the type itself is unexported.

Should I use lowercase in the interface functions of fileCollectorFs?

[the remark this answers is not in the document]
ZauberNerd: Oh, true. They are all method receivers, so they aren't actually exported functions in this file (I'm still not 100% fluent with Golang 😄).
Yes, then it's fine, I guess.

pkg/container/docker_run.go
```diff
@@ -579,8 +578,8 @@ func (cr *containerReference) copyDir(dstPath string, srcPath string, useGitIgno
 			return err
 		}
 		log.Debugf("Writing tarball %s from %s", tarFile.Name(), srcPath)
-		defer tarFile.Close()
 		defer os.Remove(tarFile.Name())
```
ChristopherHX, line 581:
```suggestion
```

pkg/container/docker_run.go
ChristopherHX: Thanks, printing the error is better. I was confused due to a lot of non deleted tmp files created by act, because windows keeps them for a long time.

ZauberNerd (approved):

PR #1087, opened 2022-03-29
ci: bump go to 1.18

catthehacker:


cplee (approved):
Time to add some generics 😎

catthehacker:
@cplee if we want to add generics then we will need to bump `go.mod` as well

PR #1095, opened 2022-03-30
tests(runner): move arm64 test to own func

catthehacker:
Prevents random failing of all other runs when qemu binfmt is not set up
but also clean up since we need `ContainerArchitecture` in single test only

Signed-off-by: Ryan <me@hackerc.at>


catthehacker:
So I kind of went on a refactoring rampage again...

ZauberNerd (approved):

cplee (approved):

Issue #1098, opened 2022-04-02
Enhancement: Overwrite workflow env vars with passed in env vars


liamphmurphy:
## Describe feature
Say we have a workflow that at either the job or steps layer has the following variable declaration:
```
foo: bar
```

At the time of writing, passing in an environment variable such as `act -j job --env foo=another-value` will not overwrite `foo`, and the value will be `bar` when the workflow runs in `act`.

The use case here is that I can overwrite the default behavior of the workflow in some way; such as where the workflow should find config files, for example. This could allow for testing without having to change the actual workflow first.

I came across this when I wanted to change the path to output some TLS certs that my app uses from what my workflow usually does.

Let me know if I missed something and it does actually work; I'm happy to go ahead and make the PR as well.

liamphmurphy:
Closing since #1100 was merged in

PR #1100, opened 2022-04-02
Allow passed in env vars to supersede ones declared in the workflow

liamphmurphy:
Simple approach to allowing for #1098 

Since https://github.com/nektos/act/blob/b3bd268e4585b32673635685d68bf86c5b2d0bfa/pkg/runner/run_context.go#L78 specifies that the `rc.Config.env` is the first map to iterate through (which is what contains any variables passed into act from the CLI), this simple duplicate check works for not letting the values of an env var in the workflow supersede what is passed into the CLI or env-files.

Added a few tests as well, please let me know if anything should be adjusted.


KnisterPeter:
I think changing the order of parameters from
```golang
rc.Env = mergeMaps(rc.Config.Env, rc.Run.Workflow.Env, rc.Run.Job().Environment())
```
 
to

```golang
rc.Env = mergeMaps(rc.Run.Workflow.Env, rc.Run.Job().Environment(), rc.Config.Env)
```

would be simpler, wouldn't it? Can you try that instead of testing for existence?

liamphmurphy:
> I think changing the order of parameters from
> 
> ```go
> rc.Env = mergeMaps(rc.Config.Env, rc.Run.Workflow.Env, rc.Run.Job().Environment())
> ```
> 
> to
> 
> ```go
> rc.Env = mergeMaps(rc.Run.Workflow.Env, rc.Run.Job().Environment(), rc.Config.Env)
> ```
> 
> would be simpler, wouldn't it? Can you try that instead of testing for existence?

Yeah that's a better way, idk why I didn't think of that, will push that up. Thanks!

catthehacker (approved):

cplee (approved):
Thanks!

PR #1101, opened 2022-04-02
implement volume mount for container job

lesomnus:
It was assumed that 'workflow.yml' was parsed and validated correctly in advance.

### DIFF
--- a/pkg/runner/run_context.go
+++ b/pkg/runner/run_context.go
@@ -106,10 +106,10 @@ func (rc *RunContext) GetBindsAndMounts() ([]string, map[string]string) {
 		if container := job.Container(); container != nil {
 			for _, v := range container.Volumes {
 				if !strings.Contains(v, ":") || filepath.IsAbs(v) {
-					// Bind anonymous volume or host file
+					// Bind anonymous volume or host file.
 					binds = append(binds, v)
-					continue
 				} else {
+					// Mount existing volume.
 					paths := strings.SplitN(v, ":", 2)
 					mounts[paths[0]] = paths[1]
 				}
--- a/pkg/runner/run_context.go
+++ b/pkg/runner/run_context.go
@@ -102,6 +102,17 @@ func (rc *RunContext) GetBindsAndMounts() ([]string, map[string]string) {
 		name + "-env":   ActPath,
 	}
 
+	for _, v := range rc.Run.Job().Container().Volumes {
+		if !strings.Contains(v, ":") || filepath.IsAbs(v) {
+			// Bind anonymous volume or host file
+			binds = append(binds, v)
+			continue
+		}
+
+		paths := strings.Split(v, ":")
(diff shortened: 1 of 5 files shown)


pkg/runner/run_context.go
```diff
@@ -102,6 +102,17 @@ func (rc *RunContext) GetBindsAndMounts() ([]string, map[string]string) {
...
 
+	for _, v := range rc.Run.Job().Container().Volumes {
+		if !strings.Contains(v, ":") || filepath.IsAbs(v) {
+			// Bind anonymous volume or host file
+			binds = append(binds, v)
+			continue
+		}
+
+		paths := strings.Split(v, ":")
+		mounts[paths[0]] = paths[1]
```
catthehacker, line 113:
```suggestion
		paths := strings.SplitN(v, ":", 2)
		mounts[paths[0]] = paths[1]
```

pkg/runner/run_context.go
```diff
@@ -102,6 +102,17 @@ func (rc *RunContext) GetBindsAndMounts() ([]string, map[string]string) {
 		name + "-env":   ActPath,
 	}
 
+	for _, v := range rc.Run.Job().Container().Volumes {
```
catthehacker, line 105: `Container()` can be `nil` and panic here, same for `Job()`

catthehacker (changes requested):
needs tests as well

pkg/runner/run_context.go
lesomnus: I think `Container()` can be `nil` if workflow doesn't use a container, but how `Job()` can be `nil`?
Are two if statements with `...() != nil` sufficient?

lesomnus: fixed on aec56fc5306f1374e9a87aa5f2b4db34be9eac41

pkg/runner/run_context.go
```diff
@@ -102,6 +102,21 @@ func (rc *RunContext) GetBindsAndMounts() ([]string, map[string]string) {
...
+		if container := job.Container(); container != nil {
+			for _, v := range container.Volumes {
+				if !strings.Contains(v, ":") || filepath.IsAbs(v) {
+					// Bind anonymous volume or host file
+					binds = append(binds, v)
+					continue
+				}
+
+				paths := strings.SplitN(v, ":", 2)
+				mounts[paths[0]] = paths[1]
```
KnisterPeter, line 115: I would prefer to use `if-else` instead of `if continue`. That would make the loop easier to follow since it has no goto-like jumps.

catthehacker:
Just need some test workflow in `pkg/runner/testdata` and add it to tests in `pkg/model/workflow_test.go` and `pkg/runner/runner_test.go`

catthehacker (approved):

KnisterPeter (approved):

cplee (approved):

PR #1111, opened 2022-04-06
chore: fold integration test outputs

KnisterPeter:
To make the test logs more readable, we should fold the
output per test into a group.


KnisterPeter:
@nektos/act-maintainers There are still test outputs not folded. But it's already way better. I've folded the big integration test outputs.

ChristopherHX (approved):
You try to read the log in the GitHub Actions UI? I just search for `FAIL:`. It would be really fancy if the logs of the failed tests are attached as a markdown document, soon on github: https://github.com/github/roadmap/issues/470.

KnisterPeter:
> You try to read the log in the GitHub Actions UI

Yes that's what I did. I think the logs are better to read like this

catthehacker:
> You try to read the log in the GitHub Actions UI? I just search for `FAIL:`. It would be really fancy if the logs of the failed tests are attached as a markdown document, soon on github: [github/roadmap#470](https://github.com/github/roadmap/issues/470).

I do too but sheer amount of logs we have takes time to load :)

catthehacker:
We can also add `error`/`debug` GHA output for some specific tests (https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-error-message)

cplee (approved):

PR #1128, opened 2022-04-15
fix: remove composite restrictions

catthehacker:
As noted by @ChristopherHX on Gitter, composite actions don't need to be restricted anymore
https://github.com/actions/runner/pull/1763


pkg/model/workflow.go
```diff
@@ -427,17 +412,14 @@ func (s *Step) Type() StepType {
...
+func (s *Step) Validate() error {
+	if s.Type() != StepTypeRun {
 		return fmt.Errorf("(StepID: %s): Unexpected value 'uses'", s.String())
 	} else if s.Type() == StepTypeRun && s.Shell == "" {
 		return fmt.Errorf("(StepID: %s): Required property is missing: 'shell'", s.String())
-	} else if !s.If.IsZero() && !config.AllowCompositeIf {
+	} else if !s.If.IsZero() {
 		return fmt.Errorf("(StepID: %s): Property is not available: 'if'", s.String())
-	} else if s.ContinueOnError && !config.AllowCompositeContinueOnError {
+	} else if s.ContinueOnError {
```
ChristopherHX, line 422: You removed any references to this function. I would either still call the function or remove it.
If we want to call it, then we should remove the following error cases
- continue-on-error
-  `s.Type() != StepTypeRun`
- if

Currently there is an [open PR for actions/runner](https://github.com/actions/runner/pull/1803) to drop the non empty shell requirement and adding defaults to composite steps

ChristopherHX, line 422: You removed any references to this function. I would either still call the function or remove it.
If we want to call it, then we should remove the following error cases
- continue-on-error
-  `s.Type() != StepTypeRun`
- if

Currently there is an [open PR for actions/runner](https://github.com/actions/runner/pull/1803) to drop the non empty shell requirement and adding defaults to composite steps

cplee (approved):

PR #1135, opened 2022-04-19
use new paths from goreleaser

cplee:


ChristopherHX:
@cplee You forget the path in https://github.com/nektos/act/blob/master/.github/actions/choco/entrypoint.sh

cplee:
Good catch @ChristopherHX - thank you 🙌

ChristopherHX (approved):

Issue #970, opened 2022-01-27
Issue: default environment variables doesn't follow recent documents


ethinx:
## System information
    - Operating System: macOS
    - Architecture: x64 (64-bit)
    - Apple M1: no
    - Docker version: 20.10.12
    - Docker image used in `act`: < can be omitted if it's included in log >
    - `act` version: 0.2.25

## Expected behaviour
There is a github env called `GITHUB_REF_NAME`

## Actual behaviour
`env.GITHUB_REF_NAME` is not set

check https://github.com/nektos/act/blob/master/pkg/runner/run_context.go, it seems that the supported env in `act` is not the same to the recent documentation: https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables

## Workflow and/or repository
Get  `${{ env.GITHUB_REF_NAME }}` in steps

## Steps to reproduce
Get  `${{ env.GITHUB_REF_NAME }}` in steps

thiskevinwang:
+1 — Would be nice to have `echo ${GITHUB_REF_NAME}` behave similar to `echo ${GITHUB_REF}`

Thanks for this great tool by the way 🙏

ChristopherHX:
_On my low priority backlog_

hashibot-web:
@ChristopherHX I'd be happy to open a PR for this.

Is it pretty much adding an additional line here?
https://github.com/nektos/act/blob/3f2469bf5385b90da5f6a5d9922ccf9c9d7274a6/pkg/runner/run_context.go#L626

```diff
  env["GITHUB_REF"] = github.Ref
+ env["GITHUB_REF_NAME"] = github.RefName
```

thiskevinwang:
@ChristopherHX I'd be happy to open a PR for this.

Is it pretty much adding an additional line here?
https://github.com/nektos/act/blob/3f2469bf5385b90da5f6a5d9922ccf9c9d7274a6/pkg/runner/run_context.go#L626

```diff
  env["GITHUB_REF"] = github.Ref
+ env["GITHUB_REF_NAME"] = github.RefName
```

ChristopherHX:
This also needs to be added there
https://github.com/nektos/act/blob/3f2469bf5385b90da5f6a5d9922ccf9c9d7274a6/pkg/model/github_context.go#L20
Populated here based on github.ref without `refs/heads/` or `refs/tags/`.
[...]

PR #1142, opened 2022-04-28
feat: support `GITHUB_REF_NAME` & `GITHUB_REF_TYPE`

thiskevinwang:
Closes https://github.com/nektos/act/issues/970


KnisterPeter:
@thiskevinwang There is a test failing which checks if the env-vars are setup correctly. Your new vars have to be added there with expected values.

thiskevinwang:
@KnisterPeter should be fixed 🤞 , waiting for CI now...

💭 It took me a bit to find the broken test. Is there an easy way to do this?
I ended up running `make test | tee testoutput.txt` and groking the output file a little bit... and eventually grep'ing for `--- FAIL` which found it.

thiskevinwang:
> > thought_balloon It took me a bit to find the broken test. Is there an easy way to do this?
> > I ended up running `make test | tee testoutput.txt` and groking the output file a little bit... and eventually grep'ing for `--- FAIL` which found it.
> 
> Search for `FAIL: ` on GitHub CI log (when I'm lazy and pushing to PR) image or inside GoLand IDE (it hides successful tests) image image
> 
> I think VSCode test UI is still lacking since last time I tried it as well as `gopls`, so no idea about free alternatives

Thanks @catthehacker 

(Why was this comment marked as off topic 🤔 ???)

ChristopherHX:
I would delete GITHUB_REF_NAME and GITHUB_REF_TYPE from the test env, like done for GITHUB_REF itself
https://github.com/nektos/act/blob/578ce6d803b5934f992fc2b2432bb3602a4b3e30/pkg/runner/step_test.go#L156

I only added ACT_OWNER to the workflow, so the tests doesn't fail in my fork due to env dependent tests. If you ask me Tests should set github.ref to a known value, then we can check if github.ref_name and github.ref_type matches.

ChristopherHX (approved):

KnisterPeter (approved):
Great

PR #1145, opened 2022-04-29
fix: panic if a step in a job is nil

### LINKED ISSUE #1144: Issue: Improperly formatted workflow yaml file causes "panic: runtime error: invalid memory address or nil pointer dereference"
## System information

- Operating System: macOS
- Architecture: 64bit
- Apple M1: no
- Docker version: 20.10.14
- Docker image used in `act`: nektos/act 
- `act` version: 0.2.26

## Expected behaviour

Yaml parsing syntax error or otherwise error specifying that yaml file cannot be run as a github workflow.

## Actual behaviour

Cryptic go error is given.

## Workflow and/or repository

<details>
  <summary>workflow</summary>

```
name: bug
on:
  pull_request: [synchronize]

jobs:
  bug:
    runs-on: ubuntu-18.04
    steps:
      -
      - name: step
        run: echo "bug"
```

</details>

## Steps to reproduce

`act pull_request -W .github/workflows/bug.yml -v -P nektos/act`

## `act` output

<details>
  <summary>Log</summary>

```
❯ act pull_request -W .github/workflows/bug.yml -v -P nektos/act
DEBU[0000] Loading environment from /Users/vito/dev/redsetter/.env
DEBU[0000] Loading secrets from /Users/vito/dev/redsetter/.secrets
DEBU[0000] Loading workflow '/Users/vito/dev/redsetter/.github/workflows/bug.yml'
DEBU[0000] Reading workflow '/Users/vito/dev/redsetter/.github/workflows/bug.yml'
DEBU[0000] Planning event: pull_request
DEBU[0000] Loading slug from git directory '/Users/vito/dev/redsetter/.git'
DEBU[0000] Found revision: f2a25115b51440edb9a44c6882044454c1b90f5f
DEBU[0000] evaluating expression 'success()'
DEBU[0000] expression 'success()' evaluated to 'true'
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x16b6fd8]

goroutine 9 [running]:
github.com/nektos/act/pkg/runner.newJobExecutor({0x1986d78?, 0xc0003bdc20})
        /private/tmp/act-20220322-14079-18pjzcc/act-0.2.26/pkg/runner/job_executor.go:35 +0x238
github.com/nektos/act/pkg/runner.(*RunContext).Executor.func1({0x19841f8, 0xc0000a3680})
        /private/tmp/act-20220322-14079-18pjzcc/act-0.2.26/pkg/runner/run_context.go:290 +0x55
github.com/nektos/act/pkg/common.Executor.Finally.func1({0x19841f8, 0xc0000a3680})
        /private/tmp/act-20220322-14079-18pjzcc/act-0.2.26/pkg/common/executor.go:177 +0x34
github.com/nektos/act/pkg/runner.(*runnerImpl).NewPlanExecutor.func1.1({0x19841f8, 0xc00041e5d0})
        /private/tmp/act-20220322-14079-18pjzcc/act-0.2.26/pkg/runner/runner.go:171 +0x2f2
github.com/nektos/act/pkg/common.NewParallelExecutor.func1.1(0x0?, 0x0?)
        /private/tmp/act-20220322-14079-18pjzcc/act-0.2.26/pkg/common/executor.go:102 +0x5c
created by github.com/nektos/act/pkg/common.NewParallelExecutor.func1
        /private/tmp/act-20220322-14079-18pjzcc/act-0.2.26/pkg/common/executor.go:100 +0x89
```

</details>

ChristopherHX:
__We need to be careful__
Closes #1144

```yaml
on: push
jobs:
  test:
    runs-on: ubuntu-latest
    steps:
    -
    - run: exit 0
```

```
go run . -W .\w.yml -P ubuntu-latest=ubuntu:latest --no-skip-checkout
Error: invalid Step 0: missing run or uses key
exit status 1
```

### DIFF
--- a/pkg/runner/job_executor.go
+++ b/pkg/runner/job_executor.go
@@ -39,6 +39,12 @@ func newJobExecutor(info jobInfo, sf stepFactory, rc *RunContext) common.Executo
 	preSteps = append(preSteps, info.startContainer())
 
 	for i, stepModel := range infoSteps {
+		if stepModel == nil {
+			return func(ctx context.Context) error {
+				common.Logger(ctx).Errorf("Invalid Step %v: missing run or uses key", i)
+				return fmt.Errorf("Invalid Step %v: missing run or uses key", i)
+			}
+		}
 		if stepModel.ID == "" {
 			stepModel.ID = fmt.Sprintf("%d", i)
 		}
--- a/pkg/runner/job_executor.go
+++ b/pkg/runner/job_executor.go
@@ -41,8 +41,7 @@ func newJobExecutor(info jobInfo, sf stepFactory, rc *RunContext) common.Executo
 	for i, stepModel := range infoSteps {
 		if stepModel == nil {
 			return func(ctx context.Context) error {
-				common.Logger(ctx).Errorf("Invalid Step %v: missing run or uses key", i)
-				return fmt.Errorf("Invalid Step %v: missing run or uses key", i)
+				return fmt.Errorf("invalid Step %v: missing run or uses key", i)
 			}
 		}
 		if stepModel.ID == "" {
--- a/pkg/runner/job_executor_test.go
+++ b/pkg/runner/job_executor_test.go
@@ -21,6 +21,7 @@ func TestJobExecutor(t *testing.T) {
 		{workdir, "uses-docker-url", "push", "", platforms},
 		{workdir, "uses-github-full-sha", "push", "", platforms},
 		{workdir, "uses-github-short-sha", "push", "Unable to resolve action `actions/hello-world-docker-action@b136eb8`, the provided ref `b136eb8` is the shortened version of a commit SHA, which is not supported. Please use the full commit SHA `b136eb8894c5cb1dd5807da824be97ccdf9b5423` instead", platforms},
+		{workdir, "job-nil-step", "push", "invalid Step 0: missing run or uses key", platforms},
 	}
 	// These tests are sufficient to only check syntax.
 	ctx := common.WithDryrun(context.Background(), true)
--- a/pkg/runner/testdata/job-nil-step/push.yml
+++ b/pkg/runner/testdata/job-nil-step/push.yml
@@ -1,0 +1,7 @@
+on: push
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+    -
+    - run: exit 0


KnisterPeter (approved):

KnisterPeter (changes requested):
I think the main problem is found in this function:
https://github.com/nektos/act/blob/a76c34987233b065109457893a334879a4615baf/pkg/model/workflow.go#L416-L428

It assumes that any type which isn't matched otherwise is a remote action.

KnisterPeter:
I've created https://github.com/nektos/act/pull/1157 to fix this

cplee (approved):

pkg/runner/job_executor.go
```diff
@@ -39,6 +39,11 @@ func newJobExecutor(info jobInfo, sf stepFactory, rc *RunContext) common.Executo
 	preSteps = append(preSteps, info.startContainer())
 
 	for i, stepModel := range infoSteps {
+		if stepModel == nil {
+			return func(ctx context.Context) error {
+				return fmt.Errorf("invalid Step %v: missing run or uses key", i)
+			}
+		}
 		if stepModel.ID == "" {
```
ChristopherHX, line 47: This is the nil dereference, which this change aims to replace with a sane error. 
> I've created https://github.com/nektos/act/pull/1157 to fix this

The same happens with `if s.Run == "" && s.Uses == "" {`, since `s` can be nil based on the bugreport.
_This cannot happen in my usecase for act, since github doesn't sent such invalid steps_

KnisterPeter: Please add a test for it then. We should be able to make sure it didn't happen again.

KnisterPeter: Also what I've meant is that we should fix the issue as early as possible.
The best would be right after yaml parsing. Otherwise we will have to add the fix in different places across the codebase.

ChristopherHX: The bigger fix would be removing the pointer from the Job.Steps `[]Step*` array to `[]Step`. Then it can never be nil, like in composite action steps. Seems like a large refactor, instead of a small patch without conflicts. You gave such conflicting PR's open.

KnisterPeter: Fine for me as well to merge this patch, but let's add a test case please.

ChristopherHX: Test case from description added, act is still pretty unstable.

cplee (approved):

KnisterPeter (approved):
Thanks 👍

Issue #1161, opened 2022-05-13
Enhancement: GitHub Enterprise server want to use some action from GitHub 


NeoHsu:
Hi, I use `act  --github-instance`  to validate GitHub actions flow on GitHub Enterprise server.

BTW, On GitHub Enterprise server, we can allow specified actions from GitHub.

When we set act `--github-instance`,  the act will clone all actions from GitHub Enterprise server.
If we allow specified actions on GitHub Enterprise server that will not found actions.
So I want to add `--through-action` to change some actions clone from GitHub Enterprise server to GitHub

<img width="752" alt="Screen Shot 2022-05-14 at 1 50 28 AM" src="https://user-images.githubusercontent.com/1842336/168339867-ded0de7c-a38a-4ab4-b80f-e876dcf66a3b.png">

NeoHsu:
<img width="752" alt="Screen Shot 2022-05-14 at 1 50 28 AM" src="https://user-images.githubusercontent.com/1842336/168339867-ded0de7c-a38a-4ab4-b80f-e876dcf66a3b.png">

NeoHsu:
this issue implement on #1162, release on [act v0.2.28](https://github.com/nektos/act/releases/tag/v0.2.28)

NeoHsu:
1.

PR #1162, opened 2022-05-13
feat: add option to bypass GHE for actions checkout

NeoHsu:
**What's this PR do?**
add `--through-action` arg to assigned actions clone from GitHub

**How should this be manually tested?**
local

**What are the relevant tickets number link?**
#1161 

**Screenshots (if appropriate)**

**Anything you want to say:**
On GitHub Enterprise server, we can allow specified actions from GitHub, 
When we use `act --github-instance` on  GitHub Enterprise server, the act will clone all actions from GitHub Enterprise server.
If we allow specified actions on GitHub Enterprise server that will not found actions.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -61,8 +61,8 @@ func Execute(ctx context.Context, version string) {
 	rootCmd.Flags().StringArrayVarP(&input.containerCapAdd, "container-cap-add", "", []string{}, "kernel capabilities to add to the workflow containers (e.g. --container-cap-add SYS_PTRACE)")
 	rootCmd.Flags().StringArrayVarP(&input.containerCapDrop, "container-cap-drop", "", []string{}, "kernel capabilities to remove from the workflow containers (e.g. --container-cap-drop SYS_PTRACE)")
 	rootCmd.Flags().BoolVar(&input.autoRemove, "rm", false, "automatically remove container(s)/volume(s) after a workflow(s) failure")
-	rootCmd.Flags().StringArrayVarP(&input.actionsFromGithub, "actions-from-github", "", []string{}, "If you are using GitHub Enterprise Server and allow specified actions from GitHub (github.com), you can set actions on this. (e.g. --actions-from-github=github/super-linter)")
-	rootCmd.Flags().StringVar(&input.actionsFromGithubToken, "actions-from-github-token", "", "If you are using actions-from-github and you want to use private actions on GitHub, you have to set personal access token")
+	rootCmd.Flags().StringArrayVarP(&input.replaceGheActionWithGithubCom, "replace-ghe-action-with-github-com", "", []string{}, "If you are using GitHub Enterprise Server and allow specified actions from GitHub (github.com), you can set actions on this. (e.g. --replace-ghe-action-with-github-com =github/super-linter)")
+	rootCmd.Flags().StringVar(&input.replaceGheActionTokenWithGithubCom, "replace-ghe-action-token-with-github-com", "", "If you are using replace-ghe-action-with-github-com  and you want to use private actions on GitHub, you have to set personal access token")
 	rootCmd.PersistentFlags().StringVarP(&input.actor, "actor", "a", "nektos/act", "user that triggered the event")
 	rootCmd.PersistentFlags().StringVarP(&input.workflowsPath, "workflows", "W", "./.github/workflows/", "path to workflow file(s)")
 	rootCmd.PersistentFlags().BoolVarP(&input.noWorkflowRecurse, "no-recurse", "", false, "Flag to disable running workflows from subdirectories of specified path in '--workflows'/'-W' flag")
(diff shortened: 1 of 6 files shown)


KnisterPeter:
@NeoHsu Thanks for your contribution. I've not yet reviewed your PR, but for your workflow GutHub recommends cloning the required actions into GHE.
They even provide a tool for that: https://docs.github.com/en/enterprise-server@3.4/admin/github-actions/managing-access-to-actions-from-githubcom/manually-syncing-actions-from-githubcom

But that doesn't mean I wouldn't support your PR

NeoHsu:
Yes, @KnisterPeter you are right. 
But GHE's owner didn't want to manager syncing actions list...so I need to implement this feature from act ... lol

NeoHsu:
Thanks @catthehacker,
I updated docs on [PR commit](https://github.com/nektos/act/pull/1162/commits/162725214c1ca34989959b27c74d126fbf0324ce)

KnisterPeter (dismissed):

KnisterPeter:
@NeoHsu Can you add a test case for your setup?

NeoHsu:
Hi @KnisterPeter thanks your suggest, 
I added test case for this PR, 
please review this case , thanks

NeoHsu:
Hey guys, I fixed conflict now, thanks

pkg/runner/step_action_remote.go
```diff
@@ -50,6 +50,14 @@ func (sar *stepActionRemote) pre() common.Executor {
...
 			}
 
+			sar.remoteAction.URL = sar.RunContext.Config.GitHubInstance
+			for _, throughAction := range sar.RunContext.Config.ThroughAction {
+				if (sar.remoteAction.Org + "/" + sar.remoteAction.Repo) == throughAction {
+					sar.remoteAction.URL = "github.com"
+					github.Token = sar.RunContext.Config.ThroughActionToken
+				}
+			}
+
```
donatobarone, line 60: first of all I want to say I have the exact same problem, I just started using act and I found this limitation in 5 mins using it :D so this is super useful. Even though I had never seen the code before, I would have approached the problem slightly different, I would have created a fallback mechanism, if you have the github-instance specified, then if the clone fails, then you can try as a backup github.com and if that fails as well then you fail the entire thing. Because assuming you have tens of actions in your workflows then you would end up having to specify a massive list, which doesn't seem super clean. Also this would open the possibility to have multiple "providers" essentially even though I don't see the use case for that...yet. Just my 2 cents

donatobarone: essentially here https://github.com/nektos/act/blob/d4272bd9fe13566ab76e5ad091aedbded74ff4e2/pkg/common/git.go#L190

donatobarone: Mmm... wouldn't checking for a 404 be enough?

NeoHsu:
Thanks @catthehacker, 
maybe I can change better naming for your suggested.
I think admins didn't want to use sync repo into GHE that will create not exist organization or duplicate organization on GHE.
We can think about the purpose of this act tool, 
Did we use this tool to test GitHub's actions workflow right?
So we should be don't care the provider have to sync repo

KnisterPeter:
I'm not your company's admin, but it should be his/her concern that you'd better clone reviewed code instead of allowing supply-chain attacks using custom code from the internet.
At least be super careful and think about this feature twice

ChristopherHX:
> As noted before, actions should be cloned to GHE and workarounding admins by bypassing restrictions is really bad

[Github Connect](https://docs.github.com/en/enterprise-server@3.1/admin/configuration/configuring-github-connect/about-github-connect#github-connect-features) is a feature of GHES and no bypass.

pkg/runner/step_action_remote.go
ChristopherHX: > I would have created a fallback mechanism

I created something like that in my action utility, well it is a pain to mirror all github actions to a fake GHES server.
- call github api `/repos/<owner>/<repo>/commits?sha=refs%2fheads%2fmaster&page=1&limit=1&per_page=1`, `refs%2fheads%2fmaster` is a url data encoded query component can be sha, ref, tag or full reference name
- the previous api call will respond with http 200 and the current sha to clone, seem to mirror the sha outputs seen on github
- other http status ask the api from github.com if `--github-connect` flag is specified
We can now log the exact sha which was used by an action, like actions/runner.
- If no github api responds with 200, just try to clone from the specified github-instance.

_Using act is not the most secure way of running actions_
I prefer act would download the archive via `<github-or-ghes-url>/<owner>/<repo>/archive/refs/heads/main.zip` / `/<owner>/<repo>/archive/refs/heads/main.tar.gz`, because a initial git clone may take a while and act stalls.

A `--though-action` flag seem to be harder to use in my point of view.

donatobarone: right, I meant regarding the strategy I was proposing

donatobarone: > Is that not what this PR does?

right, I meant regarding the strategy I was proposing

NeoHsu:
Yes, I will update PR this week

donatobarone:
just pinging again as I am really eager for this feature :)

pkg/runner/step_action_remote.go
```diff
@@ -50,6 +50,14 @@ func (sar *stepActionRemote) pre() common.Executor {
 				return nil
 			}
 
+			sar.remoteAction.URL = sar.RunContext.Config.GitHubInstance
+			for _, throughAction := range sar.RunContext.Config.ThroughAction {
+				if (sar.remoteAction.Org + "/" + sar.remoteAction.Repo) == throughAction {
```
ChristopherHX, line 55: Please use string.EqualFold instead of ==, repository names are case insensitive

cmd/root.go
```diff
@@ -61,6 +61,8 @@ func Execute(ctx context.Context, version string) {
 	rootCmd.Flags().StringArrayVarP(&input.containerCapAdd, "container-cap-add", "", []string{}, "kernel capabilities to add to the workflow containers (e.g. --container-cap-add SYS_PTRACE)")
 	rootCmd.Flags().StringArrayVarP(&input.containerCapDrop, "container-cap-drop", "", []string{}, "kernel capabilities to remove from the workflow containers (e.g. --container-cap-drop SYS_PTRACE)")
 	rootCmd.Flags().BoolVar(&input.autoRemove, "rm", false, "automatically remove container(s)/volume(s) after a workflow(s) failure")
+	rootCmd.Flags().StringArrayVarP(&input.throughAction, "through-action", "", []string{}, "If you are using GitHub Enterprise Server and allow specified actions from GitHub (github.com), you can set actions on this. (e.g. --through-action=github/super-linter)")
+	rootCmd.Flags().StringVar(&input.throughActionToken, "through-action-token", "", "If you are using through-action and you want to use private actions on GitHub, you have to set personal access token")
```
ChristopherHX, line 65: I prefer something like `--github-com-action`, **this is my opinion other maintainer may disagree.**
_I don't really like `bypass-ghe-action`._

NeoHsu: Can I use `--actions-from-github` ?
This PR is to make the actions mapping to `github.com`. 
The bypass flow is implement GHE infra on different companies, 
maybe we don't need to add this naming "bypass".

NeoHsu: ok, I got it

ChristopherHX (approved):
No further changes required in my point of view

KnisterPeter (dismissed):

donatobarone:
A couple of people already approved it which is great, I want to definitely get my hands on this feature. Shall the title and commit message be changed accordingly though? Considering the changes done to the name of the flag? 
Also when will this be available for me as a consumer on MacOs, do I just have to update via brew?

cplee (approved):

cplee:
@NeoHsu - recent merges have put this PR into conflict. I'll need your help resolving the conflicts. Appears to be whitespace?

NeoHsu:
Hi @cplee  thanks, I fixed this.
Looking forward to this feature being released soon.

cplee (approved):

cplee:
@NeoHsu - I just cut a new release yesterday (`v0.2.27`) but I know this PR was eagerly awaited so I just created a new release with this PR (`v0.2.28`).

donatobarone:
@cplee  I guess it takes some time to make that available for people via brew. As I just tried and the only one available was v.0.2.27. The related github actions seems to have succeeded :)

PR #1178, opened 2022-05-22
README.md: Mention github.token

jayvdb:


KnisterPeter:
I do agree with @catthehacker here

jayvdb:
ok

PR #1188, opened 2022-05-28
Decompress gzipped artifacts before saving to file

bogsen:
Artifacts being gzipped is just an implementation detail of the artifact upload protocol. Decompressing gzipped uploads immediately during upload allows for easier access to artifact files from outside of `act`, and simplifies artifact serving logic a bit since it can assume that artifacts aren't compressed.

### DIFF
--- a/pkg/artifacts/server.go
+++ b/pkg/artifacts/server.go
@@ -1,6 +1,7 @@
 package artifacts
 
 import (
+	"compress/gzip"
 	"context"
 	"encoding/json"
 	"errors"
@@ -11,7 +12,6 @@ import (
 	"os"
 	"path"
 	"path/filepath"
-	"strings"
 
 	"github.com/julienschmidt/httprouter"
 	"github.com/nektos/act/pkg/common"
@@ -65,8 +65,6 @@ func (fsys MkdirFsImpl) Open(name string) (fs.File, error) {
 	return os.OpenFile(fsys.dir+"/"+name, os.O_CREATE|os.O_RDWR, 0644)
 }
 
-var gzipExtension = ".gz__"
-
 func uploads(router *httprouter.Router, fsys MkdirFS) {
 	router.POST("/_apis/pipelines/workflows/:runId/artifacts", func(w http.ResponseWriter, req *http.Request, params httprouter.Params) {
 		runID := params.ByName("runId")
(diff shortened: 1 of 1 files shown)


ChristopherHX:
Decompressing the artifacts on the server really makes sense if you want direct access to the artifacts.

> if a file was uploaded as zip already this breaks the client side, right?

I agree, if you want to access artifacts from previous act builds. Otherwise I'm fine with this change.

I'm not shure about the go lint security error, all workarounds I found doesn't look to be great / fix the actual security issue.

bogsen:
> You are breaking the artifacts protocol of github here. By removing the logic if a file was uploaded as zip already this breaks the client side, right?

I'm not sure which logic you are specifically referring to. If we decompress gzipped artifacts immediately at upload time we don't need any special logic anywhere else (except maybe for artifacts that existed before this change as @ChristopherHX mentioned). And since we don't return Content-Encoding when downloading existing artifacts anymore, the client will know that the artifact isn't gzipped and handle it accordingly.

> I agree, if you want to access artifacts from previous act builds. Otherwise I'm fine with this change.

Does that currently work? I can leave those in but I'm not sure act implements the necessary APIs for something like https://github.com/marketplace/actions/download-workflow-artifact (for example) to work.

ChristopherHX:
> Does that currently work? I can leave those in but I'm not sure act implements the necessary APIs for something like https://github.com/marketplace/actions/download-workflow-artifact (for example) to work.

The action you mentioned won't work, but I think you can just use `actions/download-artifact` in a later workflow run with a different act version and the same artifacts folder. _I don't fully understand the actual reason for the requested changes._

KnisterPeter:
> > You are breaking the artifacts protocol of github here. By removing the logic if a file was uploaded as zip already this breaks the client side, right?
> 
> I'm not sure which logic you are specifically referring to. If we decompress gzipped artifacts immediately at upload time we don't need any special logic anywhere else (except maybe for artifacts that existed before this change as @ChristopherHX mentioned). And since we don't return Content-Encoding when downloading existing artifacts anymore, the client will know that the artifact isn't gzipped and handle it accordingly.
> 
> > I agree, if you want to access artifacts from previous act builds. Otherwise I'm fine with this change.
> 
> Does that currently work? I can leave those in but I'm not sure act implements the necessary APIs for something like https://github.com/marketplace/actions/download-workflow-artifact (for example) to work.

This does work and should be :smiley: 
These are the test (which might not be sufficient as we are not uploading zipped files I think): https://github.com/nektos/act/blob/master/pkg/artifacts/testdata/upload-and-download/artifacts.yml

The download protocol for the artifacts is defined by github with the [upload](https://github.com/actions/upload-artifact) and [download](https://github.com/actions/download-artifact) actions.
The expectation of someone who uploads a gzipped file would be to download a gzipped file as well.
If you uncompress the artifact during upload and store it like that, you will have to recreate the archive when downloading (and recreate an identical artifact).

Issue #1112, opened 2022-04-06
Issue: act not checking out in composite actions


alexharv074:
## System information

- Operating System: Mac OS X
- Architecture: ARM
- Docker version: 20.10.12
- Docker image used in `act`: custom
- `act` version: latest

## Contents of project dir

```text
▶ pwd
▶ ls -1
CODEOWNERS
README.md
foo.yaml
pull-request.json
```

## Workflow file

```yaml
---
name: PR Check values files

on:
  pull_request:
    branches: [master]

jobs:
  pr-tests:
    runs-on: example-base
    steps:
      - name: Run tests against the PR
        uses: example-org/private-github-actions/actions/example-project/umbrella-pr-tests@master
```

## yamllint action inside umbrella-pr-tests

Note that I have edited this file in the cache for debugging purposes.

```yaml
# /Users/alex.harvey/.cache/act/example-org-private-github-actions-actions-yaml-lint@master/actions/yaml-lint/action.yml 
---
name: "yaml-lint"

author: author@example.com
description: "runs vanilla yamllint on added & changed files"

runs:
  using: "composite"
  steps:
    - name: checkout
      uses: actions/checkout@v2

    - name: debug
      shell: bash
      run: |
        pwd
        ls -al
        yamllint foo.yaml
```

## Expected behaviour

It should yamllint the file `foo.yaml`.

## Actual behaviour

```text
▶ act -e pull-request.json -s GITHUB_TOKEN=ghp_xxxx -P example-base=gha-default-runner:latest pull_request -j pr-tests
[PR Check values files/pr-tests] 🚀  Start image=gha-default-runner:latest
[PR Check values files/pr-tests]   🐳  docker pull image=gha-default-runner:latest platform= username= forcePull=false
[PR Check values files/pr-tests]   🐳  docker create image=gha-default-runner:latest platform= entrypoint=["/usr/bin/tail" "-f" "/dev/null"] cmd=[]
[PR Check values files/pr-tests]   🐳  docker run image=gha-default-runner:latest platform= entrypoint=["/usr/bin/tail" "-f" "/dev/null"] cmd=[]
[PR Check values files/pr-tests]   🐳  docker exec cmd=[mkdir -m 0777 -p /var/run/act] user=root workdir=
[PR Check values files/pr-tests] ⭐  Run Run tests against the PR
[PR Check values files/pr-tests]   ☁  git clone 'https://github.com/example-org/private-github-actions' # ref=master
[...]

PR #1192, opened 2022-05-31
Rough handling for local checkouts in composite actions

jsoref:
This seems to work for me for #1112

Tested with:

```sh
git clone git@github.com:check-spelling/lunr.js.git
git checkout spell-check
 ~/code/nektos/act/dist/local/act -W .github/workflows/spelling.yml --container-architecture linux/arm64 -s GITHUB_TOKEN=$GITHUB_TOKEN
```

### DIFF
--- a/pkg/model/github_context.go
+++ b/pkg/model/github_context.go
@@ -34,6 +34,7 @@ type GithubContext struct {
 	RetentionDays    string                 `json:"retention_days"`
 	RunnerPerflog    string                 `json:"runner_perflog"`
 	RunnerTrackingID string                 `json:"runner_tracking_id"`
+	CopiedWorkspace  string                 `json:"copied_workspace"`
 }
 
 func asString(v interface{}) string {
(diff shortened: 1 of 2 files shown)


jsoref:
Note: I don't have the bandwidth to figure out how to make this do the right thing. So, it's more of a pointer which should hopefully help someone else fix this properly.

ChristopherHX:
...Isn't this logically equivalent to `act --no-skip-checkout -W .github/workflows/spelling.yml --container-architecture linux/arm64 -s GITHUB_TOKEN=$GITHUB_TOKEN`?
The only issue with that command is, that act always pulls from github.com or the specified github instance.

_I checked the readme and saw that this new cli option is missing, but is released in act `v0.2.26`_

jsoref:
It'd be better if the code properly stubbed checkout in composite actions...

ChristopherHX:
> I would love to get rid of that section from readme

I would approve removing this section.

> It'd be better if the code properly stubbed checkout in composite actions...

Yes, we need a mock which calls the docker cp once it sees actions/checkout to the specified path and remove the docker cp from startcontainer. I don't depend on this local checkout functionality, so only low priority for me.

Issue #1195, opened 2022-06-02
Unable to use a github environment variables on "env"


aanm:
### Bug report info

```plain text
$ act --bug-report
act version:            0.2.26
GOOS:                   linux
GOARCH:                 amd64
NumCPU:                 20
Docker host:            DOCKER_HOST environment variable is unset/empty.
Sockets found:
        /var/run/docker.sock
Config files:           
        /home/aanm/.actrc:
                -P ubuntu-latest=ghcr.io/catthehacker/ubuntu:act-latest
                -P ubuntu-20.04=ghcr.io/catthehacker/ubuntu:act-20.04
                -P ubuntu-18.04=ghcr.io/catthehacker/ubuntu:act-18.04
Docker Engine:
        Engine version:        20.10.16
        Engine runtime:        runc
        Cgroup version:        2
        Cgroup driver:         systemd
        Storage driver:        btrfs
        Registry URI:          https://index.docker.io/v1/
        OS:                    Fedora Linux 36 (KDE Plasma)
        OS type:               linux
        OS version:            36
        OS arch:               x86_64
        OS kernel:             5.17.11-300.fc36.x86_64
        OS CPU:                20
        OS memory:             63977 MB
        Security options:
                name=seccomp,profile=default
                name=cgroupns
```

### Command used with act

```sh
act -W test-act.yaml -v
```

### Describe issue

It looks it's not possible to use a `${{ github.repository_owner }}` in the environment variables as it results on a "bad substitution" error

### Workflow content

```yml
name: ACT Test

on:
  push:  {}

env:
  variable: "${{ github.repository_owner }}"

jobs:
  test:
    runs-on: ubuntu-latest
    steps:
      - name: print env.variable
        run: |
          echo ${{ env.variable }}
```

### Relevant log output

```sh
DEBU[0000] Loading environment from /github.com/nektos/act/.github/workflows/.env 
DEBU[0000] Loading secrets from /github.com/nektos/act/.github/workflows/.secrets 
DEBU[0000] Loading workflow '/github.com/nektos/act/.github/workflows/test-act.yaml' 
DEBU[0000] Reading workflow '/github.com/nektos/act/.github/workflows/test-act.yaml' 
DEBU[0000] Planning event: push                         
DEBU[0000] Loading slug from git directory '/github.com/nektos/act/.git' 
DEBU[0000] Loading revision from git directory '/github.com/nektos/act/.git' 
[...]

PR #1197, opened 2022-06-03
pkg/runner: add support to replace GitHub's env

aanm:
There might be use cases where users want to use GitHub's variables in
the environment variables, which is a valid use case.

This commits adds support for replacement of GitHub's env with GitHub's
values.

Signed-off-by: André Martins <aanm90@gmail.com>

Fixes: #1195

### DIFF
--- a/pkg/runner/testdata/issue-1195/push.yml
+++ b/pkg/runner/testdata/issue-1195/push.yml
@@ -1,0 +1,13 @@
+on: push
+
+env:
+  variable: "${{ github.repository_owner }}"
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - name: print env.variable
+        run: |
+          echo ${{ env.variable }}
+          exit ${{ (env.variable == 'nektos') && '0' || '1'}}
--- a/pkg/runner/runner.go
+++ b/pkg/runner/runner.go
@@ -14,6 +14,7 @@ import (
 
 	"github.com/nektos/act/pkg/common"
 	"github.com/nektos/act/pkg/container"
+	"github.com/nektos/act/pkg/exprparser"
 	"github.com/nektos/act/pkg/model"
 )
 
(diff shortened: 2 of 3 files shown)


pkg/runner/expression.go
```diff
@@ -55,13 +55,23 @@ func (rc *RunContext) NewExpressionEvaluator() ExpressionEvaluator {
...
 		}),
 	}
+	// evaluate environment variables since they can contain GitHub's special
+	// environment variables.
+	for k, v := range rc.GetEnv() {
+		valueEval, err := eV.evaluate(v, exprparser.DefaultStatusCheckNone)
+		if err == nil {
+			rc.Env[k] = fmt.Sprintf("%v", valueEval)
+		}
+	}
```
KnisterPeter, line 73: The expression evaluator by itself should probably not update the RunContext.
It would be done many many times, a better place would be the preparation of the pipeline right before act exectues all jobs.

KnisterPeter (changes requested):

pkg/runner/expression.go
aanm: > the preparation of the pipeline right before act exectues all jobs.
@KnisterPeter can you point me out where that is done?

KnisterPeter: The RunContext is created here:
https://github.com/nektos/act/blob/859445fb94e12cc600e5a59d9c9cd687214cb801/pkg/runner/runner.go#L157

From there you can go through the code and look for an appropriate place to add your code.

pkg/runner/testdata/issue-1195/push.yml
```diff
@@ -0,0 +1,13 @@
...
+  variable: "${{ github.repository_owner }}"
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - name: print env.variable
+        run: |
+          echo ${{ env.variable }}
+          exit ${{ (env.variable == 'nektos') && '0' || '1'}}
```
ChristopherHX, line 13: This test will fail in forks of act outside of the nektos org, maybe use a different context value like `github.event_name`. I still approve this change as is.

ChristopherHX (approved):
Thank you, I can confirm this works. Also with multiple expressions in a global env.

PR #1199, opened 2022-06-05
fix: dryrun skip local actions + enable Tests

ChristopherHX:
I broke DRYRUN for local actions last year, in a previous PR I replaced the panic with an error.

Now just skip local actions in DRYRUN, jobs using local actions now passing again.


KnisterPeter (approved):
Looks good

cplee (approved):

PR #1207, opened 2022-06-10
Added support for chunked uploads.

grigorye:


grigorye:
Closing this as apparently upload action does not do multi-part upload, rather does a [separate http request for each chunk](https://github.com/actions/upload-artifact/blob/3cea5372237819ed00197afe530f5a7ea3e805c8/dist/index.js#L7739).

PR #1206, opened 2022-06-10
Added tests for mid-size and big artifacts (to reproduce a problem with chunked uploads)

grigorye:
This should reproduce the problem (in [artifacts server](https://github.com/nektos/act/blob/8a473943c3dffbbf47c2a2ef8b261e845646fa6e/pkg/artifacts/server.go)?) with support for chunked uploads. Namely, artifacts cut at size of 8388608 (8M).

It adds two tests (artifacts 5 and 6):
- 5 is for mid-sized upload (1M), works fine
- 6 is for big upload (10M), gets cut on upload at 8388608.

```
$ act -W pkg/artifacts/testdata/upload-and-download/artifacts.yml --artifact-server-path artifacts
....
[Test that artifact uploads and downloads succeed/test-artifacts] ⭐ Run Main Download artifact #5
[Test that artifact uploads and downloads succeed/test-artifacts]   🐳  docker cp src=/Users/eg/.cache/act/actions-download-artifact@v2/ dst=/var/run/act/actions/actions-download-artifact@v2/
[Test that artifact uploads and downloads succeed/test-artifacts]   🐳  docker exec cmd=[mkdir -p /var/run/act/actions/actions-download-artifact@v2/] user= workdir=
[Test that artifact uploads and downloads succeed/test-artifacts] close /var/folders/mp/3jr0hvr922v_32bx7309bntr0000gn/T/act2616194157: file already closed
[Test that artifact uploads and downloads succeed/test-artifacts]   🐳  docker exec cmd=[node /var/run/act/actions/actions-download-artifact@v2/dist/index.js] user= workdir=
[Test that artifact uploads and downloads succeed/test-artifacts]   💬  ::debug::Resolved path is /Users/eg/Workbench/act/mid-size/artifact/path
| Starting download for Mid-Size-Artifact
[Test that artifact uploads and downloads succeed/test-artifacts]   💬  ::debug::Artifact Url: http://192.168.2.80:34567/_apis/pipelines/workflows/1/artifacts?api-version=6.0-preview
| Directory structure has been setup for the artifact
[Test that artifact uploads and downloads succeed/test-artifacts]   💬  ::debug::Download file concurrency is set to 2
| Total number of files that will be downloaded: 1
[Test that artifact uploads and downloads succeed/test-artifacts]   💬  ::debug::/Users/eg/Workbench/act/mid-size/artifact/path/file5.rnd size:(1048576) blksize:(4096) blocks:(2048)
| Skipping download validation.
[...]

PR #1208, opened 2022-06-10
Added support for chunked uploads.

grigorye:
This is follow-up for https://github.com/nektos/act/pull/1206, eventually addressing https://github.com/nektos/act/pull/677#discussion_r633667121.

It implements a naive handler that just appends every chunk, assuming that upload-artifact [does the chunk upload sequentially](https://github.com/actions/upload-artifact/blob/3cea5372237819ed00197afe530f5a7ea3e805c8/dist/index.js#L7683-L7739).

It passes the tests introduced for chunked uploads, that were failing in #1206, so I wonder if it might be enough for starters.

### DIFF
--- a/pkg/artifacts/server.go
+++ b/pkg/artifacts/server.go
@@ -50,6 +50,7 @@ type MkdirFS interface {
 	fs.FS
 	MkdirAll(path string, perm fs.FileMode) error
 	Open(name string) (fs.File, error)
+	OpenAtEnd(name string) (fs.File, error)
 }
 
 type MkdirFsImpl struct {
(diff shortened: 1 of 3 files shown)


ChristopherHX (approved):
Thank you for implementing this, _I actually don't using the artifact feature myself._
I trust your tests with uploading / downloading large random data, far better than before.

grigorye:
@ChristopherHX may you please take a look at this extra commit, that should fix the problem I introduced earlier (appending to instead of overwriting stale artifacts): https://github.com/nektos/act/pull/1208/commits/b6ae8b293f6ebec755a64a2a5dfa8c459cb51d03

ChristopherHX (approved):

grigorye:
> BTW I use this dirty code to parse the actual values of the range request: https://github.com/ChristopherHX/runner.server/blob/main/src/Runner.Server/Controllers/ArtifactController.cs#L150-L155

Yep, thanks! I also thought about further improving the parsing of Content-Length, however, given that my Go knowledge is basically non-existent, I decided to keep the parsing to the bare minimum that is probably enough "for now"...

KnisterPeter (approved):
Thank you for implementing this

PR #1204, opened 2022-06-07
fix: skip local actions pre step in any case

KnisterPeter:
We should skip local actions pre step, as it is not supported by github.
In turn we may need to late prepare remote actions which are run
as steps in a local composite action.

Fixes #1193

PR #1215, opened 2022-06-14
fix: keep action inputs

KnisterPeter:
Do not initialize inputs a second time (in pre and main step).
When the action setup already run during pre step, we must not
re-run it during the main step, otherwise the inputs will be
overwritten by possible other action inputs.

This is a regression introduced in https://github.com/nektos/act/pull/1204

### DIFF
--- a/pkg/runner/step_action_remote.go
+++ b/pkg/runner/step_action_remote.go
@@ -36,65 +36,64 @@ var (
 )
 
 func (sar *stepActionRemote) prepareActionExecutor() common.Executor {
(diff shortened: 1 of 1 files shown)


ChristopherHX (dismissed):
Please add a test, but this fix shouldn't wait for a test

cplee (approved):

ChristopherHX (approved):

cplee (approved):

cplee:
@Mergifyio refresh

cplee:
@Mergifyio refresh

PR #1202, opened 2022-06-06
feat: non root user container

ChristopherHX:
What is changed
- Call the `id` posix tool after starting any container to get the current container uid / gid, otherwise assume root without failing, **fails for docker actions**
- Try to chown the workingdir to fix the owner of the folders, **fails for docker actions**
- Add the uid and gid to the tar archive sent to docker, docker will write the files with the correct uid and gid
- Finally try to chown the dstdir to fix the owner of the folders

Closes #398
Closes #935

PR #1226, opened 2022-06-21
fix: checkout container path should always use `/`

ChristopherHX:
Container.CopyDir is no longer working with `\` as destpath. Should it support backslash as path separator e.g. replace with forwardslash.

It seems like my https://github.com/nektos/act/pull/1202 change causes issues with checkout on windows, files are no longer where they are expected to be.

_I wondered in the past why I saw `\mnt\c\...` in the act log, this now cause issues_

### DIFF
--- a/pkg/runner/step_action_remote.go
+++ b/pkg/runner/step_action_remote.go
@@ -123,7 +123,7 @@ func (sar *stepActionRemote) main() common.Executor {
 					return nil
 				}
 				eval := sar.RunContext.NewExpressionEvaluator(ctx)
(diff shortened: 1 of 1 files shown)


cplee (approved):

domagojmedo:
Any ETA on when this fix will be released?

ChristopherHX:
I guess july 1, as long as the new monthly release workflow works.

CI Tests don't cover windows + docker, windows support can regress easily. It isn't possible to run docker linux container on github hosted windows runners without self hosting them.

PR #1230, opened 2022-06-24
Fixed step environment not accounted in the composite action input.

grigorye:
This potentially fixes https://github.com/nektos/act/issues/1231.


KnisterPeter:
Hi @grigorye, can you please add a test case for this change?

grigorye:
@KnisterPeter Yep, I will try, when I have a bit of time.

PR #1247, opened 2022-07-07
fix: allow TimeoutMinutes to be expression in Jobs

hishamhm:
This change stops act from rejecting valid entries such as

```
    timeout-minutes: ${{ matrix.runtime == 'v8' && 30 || 15 }}
```

at the `jobs:` level.

This change complements the fix that was already in place for the Step struct, done by @catthehacker in #1217. See: https://github.com/nektos/act/commit/52f5c4592cbbf2e4dde3b35ad42f4e7475ad52df


catthehacker (approved):

ChristopherHX (approved):
I prefer yaml.Node, but since timeout is ignored nothing to change here

cplee (approved):

Issue #1248, opened 2022-07-07
 Error response from daemon: Head "https://ghcr.io/v2/catthehacker/ubuntu/manifests/act-latest": denied: denied


k8s-gpubuilder-markus:
### Bug report info

```plain text
act version:            0.2.29
GOOS:                   linux
GOARCH:                 amd64
NumCPU:                 16
Docker host:            DOCKER_HOST environment variable is unset/empty.
Sockets found:
        /var/run/docker.sock
Config files:           
        /root/.actrc:
                -P ubuntu-latest=ghcr.io/catthehacker/ubuntu:act-latest
                -P ubuntu-22.04=ghcr.io/catthehacker/ubuntu:act-22.04
                -P ubuntu-20.04=ghcr.io/catthehacker/ubuntu:act-20.04
                -P ubuntu-18.04=ghcr.io/catthehacker/ubuntu:act-18.04
Build info:
        Go version:            go1.18.3
        Module path:           github.com/nektos/act
        Main version:          (devel)
        Main path:             github.com/nektos/act
        Main checksum:         
        Build settings:
                -compiler:            gc
                -ldflags:             -s -w -X main.version=0.2.29 -X main.commit=9499612b5818774c5c6d73fbf2596ed464b7b01d -X main.date=2022-07-01T02:54:35Z -X main.builtBy=goreleaser
                CGO_ENABLED:          0
                GOARCH:               amd64
                GOOS:                 linux
                GOAMD64:              v1
                vcs:                  git
                vcs.revision:         9499612b5818774c5c6d73fbf2596ed464b7b01d
                vcs.time:             2022-06-30T02:19:35Z
                vcs.modified:         false
Docker Engine:
        Engine version:        20.10.12
        Engine runtime:        runc
        Cgroup version:        1
        Cgroup driver:         cgroupfs
        Storage driver:        overlay2
        Registry URI:          https://index.docker.io/v1/
        OS:                    Ubuntu 20.04.4 LTS
        OS type:               linux
        OS version:            20.04
        OS arch:               x86_64
        OS kernel:             5.13.0-52-generic
        OS CPU:                16
        OS memory:             31826 MB
        Security options:
                name=apparmor
                name=seccomp,profile=default
```

### Command used with act

```sh
act 
(the initial run to pull the image down)
```

### Describe issue

I expect for act to download the runner image medium.

### Workflow content

```yml
This is for the initial run of act, I'm not at this step yet.
```

### Relevant log output

```sh
./bin/act 
[...]

PR #1270, opened 2022-07-22
Update README with note about docker login ghcr.io

highb:
As discussed in https://github.com/nektos/act/issues/1248 , the images for this tool are now hosted in `ghcr.io` which (apparently) no longer allows unauthenticated pulls. This PR adds some documentation around how to login to `ghcr.io` as well as mentions the error that occurs if you are not in order to help new users get over this stumbling block.

New contributor here, just wanted to say thanks to all the contributors for making this project. ❤️ I'm looking forward to a large reduction in the number of `fix: fmt` and `fix: linting` commits in my PRs now that I can run my workflows locally. 😂

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,6 +122,13 @@ curl https://raw.githubusercontent.com/nektos/act/master/install.sh | sudo bash
 
 Download the [latest release](https://github.com/nektos/act/releases/latest) and add the path to your binary into your PATH.
 
+# Logging into GitHub Packages
+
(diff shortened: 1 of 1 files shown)


highb:
> GitHub fixed their stuff and authentication is not required anymore

When I tried using `act` on Friday, login was still required, but yeah if that was just a temporary issue then this documentation wouldn't be necessary.

highb:
Confirmed, it appears that GHCR no longer requires auth for public images now. I'll go ahead and close this.

PR #1285, opened 2022-07-29
refactor: ignore already closed error

KnisterPeter:
This hides the `file already cloesd` error as it is
distracting in the output and does not provide any value.


KnisterPeter:
I've tested it and it's fine

cplee (approved):

Issue #97, opened 2020-02-24
Add support for `--no-container` to run windows/macos


chenrui333:
Thanks for creating this wonderful tool. It would be nice to have MacOS also supported as well.

cplee:
All steps for act run in a docker container...not sure how to accomplish that with MacOS runner.  Would you want it to NOT use a container and just run the commands directly on the mac that act is running on?

fwilhe:
Kinda related: Sometime I think you really don't want to run stuff in containers, for example when act is invoked in a container itself which contains the required software already. Not sure if this is in scope, but it would be kind of nice if it had a `--no-container` flag which would do what it does now, except for using the locally installed tools..

How feasible would this be?

cplee:
I like the idea @fwilhe - would like to see a flag `--no-container` that runs all the steps directly on your local workstation, and assumes the right tools are installed.  This would enable both windows and macos environments.

albertofem:
@cplee I have a rough but working implementation I'm using locally to test some macOS-based actions. It should also be easy enough to have it working in Windows. Please take a look here:

https://github.com/nektos/act/compare/master...albertofem:feature/running_on_host?expand=1

If you think this approach is good, I'll go ahead and make the PR once I have a better implementation and have tested extensively on both Mac and Windows.

cplee:
@albertofem - this looks really promising!  What do you think about also supporting `--no-container` to force the platform to be `host`?

jcansdale:
@cplee,

I've been experimenting with running `act` inside GitHub Codespaces.

![image](https://user-images.githubusercontent.com/11719160/81909332-5c1fbf80-95c2-11ea-9a61-58e57409fada.png)

If there was a `--no-container` option to run actions on the host VM, that would be great!

Unfortunately the `actions/setup-dotnet` action is failing at the moment. I'm guessing this will be the same for many setup actions.

fwilhe:
In my experience it is a viable workaround for failing setup actions to build a docker image that has everything your build needs. Provides you the same experience locally and in actions if you run your workflow in the image.

jcansdale:
@fwilhe I did wonder about doing that. It's good to hear that it works well in practice!

Since you can optionally specify an image to use with Codespaces, this could be a good opportunity to sync-up build and development environments. 🤔

albertofem:
[...]

PR #1293, opened 2022-08-05
feat: Host environment

ChristopherHX:
This is my refactored implementation of a non docker based executor

To test running actions without docker specify `-P ubuntu-latest=-self-hosted`

I refactored a larger amount of code
- Allow to alter `/var/run/act`
  - otherwise concurrent executors on the same host conflict each other
  - this folder cannot be created on windows
- Tweak `PATH` manipulation to use custom list seperator, while not running in docker
- Different runner context if running on host vs. in a container

Fixes https://github.com/nektos/act/issues/97

_I need to look into possible linting violations_

**Known Issues**
- windows cmd not working, omitted here to shrink size
- Possible escaping problems for default shell, omitted here to shrink size
- On windows cmdline needs to be added to the container interface, omitted here to shrink size
- Default shell bash not suitable on windows
- docker actions are intentionally failing, since window and macos runner doesn't support them officially and some platforms like freebsd, etc. doesn't support docker

_I want to know if this design change has a chance to get merged_

_Some implementation Details_
- lookpath is mostly copied from the go stdlib, because the original didn't allow to provide a custom PATH variable for lookup and I want to make shure multiple parallel workers don't update `PATH` env var to use the stdlib directly
- Most changes are done by replace operations, API Change
- New if conditions to avoid a crash in Tests, these issues don't appear with a well constructed run_context

## Commit Message

{{title}} (#1293)


KnisterPeter:
I will look into this starting next week.
It's a lot to review. But I like the idea

pkg/container/linux_container_environment_extensions.go
```diff
@@ -0,0 +1,71 @@
...
+	"path/filepath"
+	"regexp"
+	"runtime"
+	"strings"
+
+	log "github.com/sirupsen/logrus"
+)
+
+type LinuxContainerEnvironmentExtensions struct {
+}
```
KnisterPeter, line 13: Shouldn't this implement the `ExecutionsEnvironment`?

Also, why is it called `LinuxContainerEnvironmentExtensions` instead of `LinuxContainerEnvironment`?

KnisterPeter (changes requested):
Not tested yet but I've added a few questions during my first scan through all the changes.
Nice work so far.

pkg/runner/run_context.go
```diff
@@ -136,6 +145,64 @@ func (rc *RunContext) startJobContainer() common.Executor {
 			}
 			return true
 		})
+		if image == "-self-hosted" {
+			cacheDir := rc.ActionCacheDir()
+			randBytes := make([]byte, 8)
+			_, _ = rand.Read(randBytes)
+			miscpath := filepath.Join(cacheDir, hex.EncodeToString(randBytes))
```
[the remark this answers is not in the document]
KnisterPeter: I see. I have less experience with windows. Maybe we can start with your approach and see if there any problems with it and change it later on if required.

pkg/container/linux_container_environment_extensions.go
KnisterPeter: Okay got it. I looked for a way to have the go compiler inform that the interface was not implemented.
If it is defined as an 'implicit' interface, it does work, but whenever the interface changes it does break and it's hard to understand why. Have seen that several times already.
But moving everything around and making test harder is also not a good option.

Do you see any other idea to make the compiler more aware of the requirements?

KnisterPeter: > Other than assigning a struct to an interface, I don't see how the compiler will alert you about an error by not meeting the interface requirements.

You are right, interfaces are implicit (I'm still learning go...)

This might be nitpicking and I will resolve this, but we could do this either near to the struct or in a test:
```go
type Interf interface {
	Run()
}

type Impl struct{}

var _ Interf = Impl{}
var _ Interf = (*Impl)(nil)
```

Anyway, I'll leave this decision up to you if you see value in it. :smiley:

KnisterPeter (approved):

KnisterPeter:
@ChristopherHX is this in a ready-for-review state? If yes I hope I can find some time to review and test this.

KnisterPeter (dismissed):

KnisterPeter:
Just as a question, I've tried to run your code with the workflow below and two different CLI calls:

```shell
x@y:~/source/act (host_environment)$ cat test.yaml 
name: test
on: push
jobs:
  test:
    runs-on: -self-hosted
    steps:
      - run: echo abc
x@y:~/source/act (host_environment)$ ./dist/local/act -W test.yaml 
[test/test] 🚧  Skipping unsupported platform -- Try running with `-P -self-hosted=...`
x@y:~/source/act (host_environment)$ ./dist/local/act -P -self-hosted=-self-hosted -W test.yaml 
[test/test] ⭐ Run Main echo abc
| abc
[test/test]   ✅  Success - Main echo abc
[test/test] 🏁  Job succeeded
```

Can we make it work with just `./dist/local/act -W test.yaml`? Since no mapping between a platform and self-hosted is required that would simplify it.
On the other hand I don't see this as a show-stopper. Having it mapped to a nice name is also always fine for me :smiley:

cplee:
@ChristopherHX I know this one has been out here for a long time but please give me some more time to review. This is a highly requested feature so I'm very eager to get this merged but I need more time to get through it.

KnisterPeter (dismissed):

ZauberNerd (approved):

cplee (approved):
@ChristopherHX - sorry for the delays with this one. Finally got a chance to try out locally. Looks great, thanks for your patience!!

cplee:
@ChristopherHX - I'll be watching 👀

cplee:
🎉🎉🎉

KnisterPeter:
Thanks for your hard work @ChristopherHX

PR #1326, opened 2022-08-30
help output: use consistent terminology between `-j` & `-l`

ljharb:
I was *very* confused for awhile trying to figure out how to go from `act -l`, to "run a specific job".

Since the column in `act -l` is titled "Job ID", i figured that adding the literal text "job ID" to the help output for `-j` would have clued me in immediately as to which command I needed to run.

Happy to make whatever changes you like, but I'm hoping this is a trivial enough change :-)


cplee (approved):
Great catch - thanks!

PR #1356, opened 2022-09-24
fix yay command

sloganking:
`-Syu` is for updating all packages. `-S` is for installing a package.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,7 +76,7 @@ scoop install act
 [![aur-shield](https://img.shields.io/aur/version/act)](https://aur.archlinux.org/packages/act/)
 
 ```shell
-yay -Syu act
+yay -S act
 ```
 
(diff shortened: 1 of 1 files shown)


cplee:
This was updated in #1355 

@hXtreme - can you comment on the change?

hXtreme:
On arch it's bad practice to do partial upgrades, which is what happens if you only have `-S`, pr #1355 has a link to arch wiki explaining things in more detail.

cplee:
@sloganking ☝️ ?

sloganking:
I can't get any `yay -Syu` to work right now. As some other package fail to build. But `yay -Syu act` seems like it will install `act` correctly. And I think `yay -Syu act` is more correct in theory, even if it may make users run into more build issues with other packages. I'll close this PR for now.

PR #1351, opened 2022-09-16
fix: support docker create arguments from container.options (#1022)

alex-savchuk:
Important note: parsing of CLI options of docker is fetched from docker/cli sources explicitly - see docker_cli.go. 
Please confirm it's acceptable.

Fixes https://github.com/nektos/act/issues/711
Fixes https://github.com/nektos/act/issues/1022

Issue #1022, opened 2022-03-01
Issue: parse all container/services options


catthehacker:
```
on: push

jobs:
  docker-container:
    runs-on: ubuntu-latest
    container:
      image: alpine
      options: --it --rm
    steps:
      - uses: actions/checkout@v2
```

bhaveshprajapat:
bump

KnisterPeter:
@cplee @alex-savchuk This change contains a major issue. The workflow below does not return the hostname of the host but some other invalid fqdn.
I'm not sure why this happens, but it is a side-effect of this PR. Reverting it does restore the previous behavior.

```yaml
name: test
on: push

jobs:
  test:
    runs-on: ubuntu-latest
    steps:
      - run: hostname -f
```

KnisterPeter:
According to the [docs](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idcontainer), if no container is set, the job should run on the host.
This was imitated with act until this fix was applied.
We should restore the default behavior unless there is a `container` definition on a job.

alex-savchuk:
Checking it.

alex-savchuk:
@KnisterPeter 
Could you please provide output for previous output of ```hostname -f``` and new one?
I checked and see "docker-desktop" for both cases :(

KnisterPeter:
The 'previous' output was the output of the host machine as act was used to start the docker container in network mode `host`. Now with your fix it is started with the default network mode.
As you are parsing an empty string into docker options (using the cli tool) the output are all the docker defaults. Then you merge that into (and over) the act defaults.
I've already build a fix for this: https://github.com/nektos/act/pull/1375

Please test if your change did not break with that. :smiley:

PR #1375, opened 2022-10-07
test: add test for networking setup in act

KnisterPeter:
This test makes sure that the hostname inside of act is resolvable.
This will prevent regressions in the network setup (see #1351, #1022)

Act need to be started with docker network mode `host` by default. This was changed recently, leading to networking changes. That do break the previous (and expected) behavior.


KnisterPeter:
@alex-savchuk Please have a look as well next to the maintainers :pray:

alex-savchuk:
@KnisterPeter Hello.
Looking at this issue right now.

d66515:
Looks good from my point of view.

Main change, as i understand, is that we do not merge nothing if options are empty. With my solution we did.

But i see that test does not catch the problem but only prints data for manual checking.
Maybe we somehow pass hostname of host system as input and verify that it equals to current container hostname?
But maybe current test is enough.

Thank you for catching this problem and providing solution!
I'm traveling right now so wouldn't be able to fix that at least until Wednesday :(

alex-savchuk:
Looks good from my point of view.

Main change, as i understand, is that we do not merge nothing if options are empty. With my solution we did.

But i see that test does not catch the problem but only prints data for manual checking.
Maybe we somehow pass hostname of host system as input and verify that it equals to current container hostname?
But maybe current test is enough.

Thank you for catching this problem and providing solution!
I'm traveling right now so wouldn't be able to fix that at least until Wednesday :(

KnisterPeter:
> Main change, as i understand, is that we do not merge nothing if options are empty. With my solution we did.
> 
Exactly. I'm still unsure if that is good enough, but it should restore basic functionality.

> But i see that test does not catch the problem but only prints data for manual checking. Maybe we somehow pass hostname of host system as input and verify that it equals to current container hostname? But maybe current test is enough.
> 
The test `host $(hostname -f)` does fail with your change, but succeeds with the expected state.
At least on linux and GitHub it does, so it should be sufficient I think.

> Thank you for catching this problem and providing solution! I'm traveling right now so wouldn't be able to fix that at least until Wednesday :(

No problem, I did not see this issue while I've reviewed your code. This happens all the time as our tests are not yet good enough. A fix is already included in this PR.

cplee (approved):

PR #1376, opened 2022-10-07
fix: interpolate action input defaults

KnisterPeter:
This fixes the regression to interpolate input defaults which contain
expressions.


cplee (approved):
Thanks @KnisterPeter

PR #1385, opened 2022-10-11
fix: re-evaluate env for remote composite actions

KnisterPeter:
The env for remote composite actions need to be re-evaluated
at every stage (pre, main, post) as it is created during the
pre stage but there might be changes used as input to this
actions main stage (e.g. outputs for another action).

This is not required for local actions as their env is created
for the main stage (there is no pre stage).
Post stages do not need an updated env since they cannot recieve
inputs from other actions.


ZauberNerd (approved):

KnisterPeter:
I've updated the local actions and the remote actions test. Only the remote one fails.
But that might not be sufficient as test case.

But at least we have one more (or better) test case now. :smiley:

KnisterPeter:
And with 'post stages cannot receive inputs' I've meant that in a post step you can not declare a specific `with` input that is not there for the main input.
Not sure exactly how GitHub behaves here. The main issue I think is, that the expressions are evaluated too eager instead of as-lazy-as-possible.

cplee (approved):

PR #1391, opened 2022-10-14
feat: GITHUB_STATE and GITHUB_OUTPUT file commands

### LINKED ISSUE #1386: Add support for `>> $GITHUB_OUTPUT` because `set-output` will be deprectated.
### Act version

0.2.32.r13.gff5e289

### Feature description

I think this functionality isn't yet part of act, which isn't strange since i think it is just announced.
But it looks like they are going to deprecate the `set-output` and `set-state` options and replace them with `>> $GITHUB_OUTPUT` and `>> $GITHUB_STATE`

See: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
And also: https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-output-parameter

ChristopherHX:
Retargeted to master.

_One problem with the GITHUB_ENV logic, seem to be that it is never cleaned or recreated if deleted._

~~**DO NOT MERGE, before "feat: Host Environment" has been merged**~~

~~_I don't like merge conflicts due to actpath changes_~~

Resolves #1386, #1442

### DIFF
--- a/pkg/runner/step.go
+++ b/pkg/runner/step.go
@@ -97,7 +97,7 @@ func runStepExecutor(step step, stage stepStage, executor common.Executor) commo
 		logger.Infof("\u2B50 Run %s %s", stage, stepString)
 
 		// Prepare and clean Runner File Commands
-		actPath := ActPath
+		actPath := rc.JobContainer.GetActPath()
 		outputFileCommand := path.Join("workflow", "outputcmd.txt")
 		stateFileCommand := path.Join("workflow", "statecmd.txt")
 		(*step.getEnv())["GITHUB_OUTPUT"] = path.Join(actPath, outputFileCommand)
--- a/pkg/runner/step.go
+++ b/pkg/runner/step.go
@@ -3,9 +3,11 @@ package runner
 import (
 	"context"
 	"fmt"
+	"path"
 	"strings"
 
 	"github.com/nektos/act/pkg/common"
+	"github.com/nektos/act/pkg/container"
 	"github.com/nektos/act/pkg/exprparser"
 	"github.com/nektos/act/pkg/model"
 )
@@ -94,6 +96,20 @@ func runStepExecutor(step step, stage stepStage, executor common.Executor) commo
 		}
 		logger.Infof("\u2B50 Run %s %s", stage, stepString)
 
+		// Prepare and clean Runner File Commands
+		actPath := ActPath
+		outputFileCommand := path.Join("workflow", "outputcmd.txt")
+		stateFileCommand := path.Join("workflow", "statecmd.txt")
+		(*step.getEnv())["GITHUB_OUTPUT"] = path.Join(actPath, outputFileCommand)
+		(*step.getEnv())["GITHUB_STATE"] = path.Join(actPath, stateFileCommand)
+		_ = rc.JobContainer.Copy(actPath, &container.FileEntry{
+			Name: outputFileCommand,
+			Mode: 0666,
+		}, &container.FileEntry{
+			Name: stateFileCommand,
+			Mode: 0666,
+		})(ctx)
+
 		err = executor(ctx)
 
 		if err == nil {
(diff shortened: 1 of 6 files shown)


BlackDex:
It works for my workflow file btw :).
Thanks

KnisterPeter (approved):

pkg/runner/step.go
```diff
@@ -94,6 +96,20 @@ func runStepExecutor(step step, stage stepStage, executor common.Executor) commo
 		}
 		logger.Infof("\u2B50 Run %s %s", stage, stepString)
 
+		// Prepare and clean Runner File Commands
+		actPath := ActPath
```
veleek, line 100: Unnecessary assignment, I would just use `ActPath` in the `path.Join` calls.

pkg/runner/step.go
```diff
@@ -94,6 +96,20 @@ func runStepExecutor(step step, stage stepStage, executor common.Executor) commo
...
+		outputFileCommand := path.Join("workflow", "outputcmd.txt")
+		stateFileCommand := path.Join("workflow", "statecmd.txt")
+		(*step.getEnv())["GITHUB_OUTPUT"] = path.Join(actPath, outputFileCommand)
+		(*step.getEnv())["GITHUB_STATE"] = path.Join(actPath, stateFileCommand)
```
veleek, line 104: Could we set these once inside `run_context.go` instead of setting them to the same value whenever each step runs.  Then update this code to just clear out the files using the paths grabbed from the step env?

veleek (approved):

pkg/runner/step.go
veleek: Okay, so I fiddled around with this on a real action, and it looks like GITHUB_OUTPUT is set to a different path for each step (like `/home/github/github-actions/_work/_temp/_runner_file_commands/set_output_18bd9c4c-7e30-40eb-b659-4a0a3be207b4`)

image

I also tried overriding the value using an `env:` block but it appears to have been ignored.  So maybe leaving this logic here and having it reset/clear out on every step is the right way to go. Or if you want to be even MORE like the real thing, then set it to a random file name on each run.

veleek: And it looks like they actually use the same guid (maybe a generated step ID from each run) for all of the file commands and then merge stuff in later.

```
  echo "Writing to $GITHUB_OUTPUT"
  echo "Writing to $GITHUB_STATE"
  echo "Writing to $GITHUB_ENV"

Writing to /home/github/github-actions/_work/_temp/_runner_file_commands/set_output_67a16ed1-fc66-[4](https://github.com/Unity-Technologies/mp-groups/actions/runs/3472080444/jobs/5802425679#step:4:4)fc9-9a4a-a601a4abdf[5](https://github.com/Unity-Technologies/mp-groups/actions/runs/3472080444/jobs/5802425679#step:4:5)0
Writing to /home/github/github-actions/_work/_temp/_runner_file_commands/save_state_[6](https://github.com/Unity-Technologies/mp-groups/actions/runs/3472080444/jobs/5802425679#step:4:6)[7](https://github.com/Unity-Technologies/mp-groups/actions/runs/3472080444/jobs/5802425679#step:4:7)a16ed1-fc66-4fc[9](https://github.com/Unity-Technologies/mp-groups/actions/runs/3472080444/jobs/5802425679#step:4:9)-9a4a-a601a4abdf50
Writing to /home/github/github-actions/_work/_temp/_runner_file_commands/set_env_67a[16](https://github.com/Unity-Technologies/mp-groups/actions/runs/3472080444/jobs/5802425679#step:4:17)ed1-fc66-4fc9-9a4a-a601a4abdf50
```

cplee (approved):

cplee:
@Mergifyio unqueue

cplee (approved):

till:
This works beautifully. :)

```
brew unlink act
brew install act --HEAD
```

Thank you! :)

PR #1415, opened 2022-10-28
refactor: allow error state on conditional executor

KnisterPeter:
The conditional executor should be able to handle
an error during conditional evaluation.
Not being able to abort execution in case of an error is a severe limitation which complicates using code by opt-out of executor chaining.

### DIFF
--- a/pkg/common/dryrun.go
+++ b/pkg/common/dryrun.go
@@ -9,14 +9,14 @@ type dryrunContextKey string
 const dryrunContextKeyVal = dryrunContextKey("dryrun")
 
 // Dryrun returns true if the current context is dryrun
-func Dryrun(ctx context.Context) bool {
+func Dryrun(ctx context.Context) (bool, error) {
 	val := ctx.Value(dryrunContextKeyVal)
 	if val != nil {
 		if dryrun, ok := val.(bool); ok {
-			return dryrun
+			return dryrun, nil
 		}
 	}
-	return false
+	return false, nil
 }
 
 // WithDryrun adds a value to the context for dryrun
--- a/pkg/common/executor.go
+++ b/pkg/common/executor.go
@@ -27,7 +27,7 @@ func Warningf(format string, args ...interface{}) Warning {
 type Executor func(ctx context.Context) error
 
 // Conditional define contract for the conditional predicate
-type Conditional func(ctx context.Context) bool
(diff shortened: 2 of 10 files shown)


KnisterPeter:
> In my opinion, I prefer to add a new IfWithError chaining function instead of make the existing common.Dryrun return an error which is currently impossible.

I'm about to use this `If` executor in the future for reusable workflows. And there are error cases. Therefore I've updated the code.
Having an `If` and `IfWithError` seems like adding mental overhead during reading the code. But I can change based on other opinions if you are fine with that.

cplee:
This change adds a lot of unnecessary `if err != nil` for dryrun. I'm struggling to see the value this is adding. I can't find anywhere in the change where a conditional is returning something other than `nil` for the error. Are we sure we need this?

KnisterPeter:
It is a preparation for this:
https://github.com/nektos/act/pull/1423

And potential other cases.
It would allow to keep the functional way, but it you prefer I can change it to be more imperative.

cplee:
I don't see where #1423 uses error response in a conditional. Sorry, I'm missing it. Can you point to a line of code so I can better understand?

Tell me more about what you mean by functional vs. imperative. The signature for `Conditional` is intended to be functional by following the definition of a predicate function - returns a single value, `true` or `false`.  If we add an additional return type, then it would no longer be a predicate function. In what ways would that change make it more functional and less imperative?

Again, apologies for missing it. I have no doubt that you are trying to solve a real problem, I'm just not seeing the problem clearly yet. 🙏

KnisterPeter:
No apologies 😊
It's line 286 (I'm on mobile so it's hard to add a link).
The isEnabled function need to return an error besides the boolean value. So any predicate with side effects (e g. IO) would need this.

cplee:
I see it now. I don't think complexity that this change introduces is worth the benefit I see on line 286. Additionally, this moves us away from a functional predicate function. My suggestion is to simplify that code in run_context by wrapping `rc.isEnabled` with something like:

```
isEnabled := func(ctx context.Context) bool {
  res, _ := rc.isEnabled(ctx)
  return res
}
return executor.If(isEnabled)
```

KnisterPeter:
That would ignore the error which I definitely don't want.
But I'm okay with your decision. I'll keep the related PR in imperative code then and will remove the comment.

PR #1429, opened 2022-11-07
feat: add support for writing to GITHUB_OUTPUT

veleek:
As of the beginning of October, [the `::set-output` command is being deprecated](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/) in favor of using [Environment Files](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#environment-files) to avoid exposing data by untrusted loggers.

Use of these commands will begin failing as of 31st May 2023.  

This PR attempts to enable `act` to support the `$GITHUB_OUTPUT` environment file.  I'm a new contributor, so I'm not 100% sure if this covers all of the expected edge cases, so please let me know if there's a better way to accomplish this (e.g. maybe by adding an additional exectutor wrapper or something).

### DIFF
--- a/pkg/runner/run_context.go
+++ b/pkg/runner/run_context.go
@@ -187,6 +187,10 @@ func (rc *RunContext) startJobContainer() common.Executor {
 				Name: "workflow/paths.txt",
 				Mode: 0666,
 				Body: "",
+			}, &container.FileEntry{
+				Name: "workflow/outputs.txt",
+				Mode: 0666,
+				Body: "",
 			}),
 		)(ctx)
 	}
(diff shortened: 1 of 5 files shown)


veleek:
> Already exists as #1391. Including the GITHUB_STATE file command

Ahhh shooot!!! I swear I did search for PRs with GITHUB_OUTPUT in them before I did this.....  yup, this is probably duplicated.  

@ChristopherHX - I added a few additional test cases (in https://github.com/nektos/act/pull/1429/files#diff-e3d14d4179a7e428ad54950c2153af387053f4d0e283f95433a4d1cb52eca973) that could be useful to add to your PR if you want to pull them over?

veleek:
Closing as duplicate.

Issue #1012, opened 2022-02-24
Issue: hashFiles function attempts to hash directories and errors with io.Copy error


leighmcculloch:
## System information

- Operating System: `macOS`
- Architecture: `arm64 (64-bit)`
- Apple M1: `yes`
- Docker version: `20.10.12`
- Docker image used in `act`: `ghcr.io/catthehacker/ubuntu@act-latest`
- `act` version: `v0.2.26-0.20220215200300-ff13844b8643`

## Expected behaviour

Running a workflow that includes this string interpolated works on GitHub actions:

```
${{ github.workflow }}-${{ github.job }}-${{ matrix.go }}-${{ runner.os }}-go-all-${{ github.ref }}-${{ hashFiles('**', '!.git') }}
```

It results in a string like so:

```
Go-check-1.17-Linux-go-all-refs/pull/3727/merge-5ee8a89e3d82367e7e6cc247d85e042a1952b32fd4331db7d74cdd08b5bb80ad
```

Example: https://github.com/stellar/go/runs/5309071935?check_suite_focus=true

## Actual behaviour

Running the action with the string interperolated with `act` displays this error:

```
ERRO[0037] Unable to interpolate expression 'format('{0}-{1}-{2}-{3}-go-all-{4}-{5}', github.workflow, github.job, matrix.go, runner.os, github.ref, hashFiles('**', '!.git'))': Unable to io.Copy: read /Users/leighmcculloch/Code/stellar--go/.circleci: is a directory 
```

It appears that `act` is interpreting the `hashFiles('**', '!.git')` incorrectly and is attempting to hash directories, not just files.

## Workflow and/or repository

https://github.com/stellar/go/blob/189950c43f89ea908699953d298b120f1a81b371/.github/workflows/go.yml

## Steps to reproduce

```
git clone https://github.com/stellar/go
cd go
git checkout 189950c43f89ea908699953d298b120f1a81b371
act -j check
```

## `act` output

<details>
  <summary>Log</summary>

```none
[Go/check] 🧪  Matrix: map[go:1.17 os:ubuntu-latest]
[Go/check] 🚀  Start image=ghcr.io/catthehacker/ubuntu:act-latest
[Go/check]   🐳  docker pull image=ghcr.io/catthehacker/ubuntu:act-latest platform= username= forcePull=false
[...]

PR #1446, opened 2022-11-19
fix: handle subdirectories and exclusions (#1012)

avx-rchung:


KnisterPeter (approved):
Look fine for me. Thanks @avx-rchung :+1:

cplee (approved):

PR #1451, opened 2022-11-22
ENH: enable capability from command line to bind host volumes to containers

rraymondgh:
See also question I raised on SO https://stackoverflow.com/questions/74516196/nektos-act-docker-containers-how-to-mount-a-volume

There are quite significant overheads with **conda** and act.  Sharing the downloaded packages between host and containers by binding host directory of downloaded packages with **act** containers significantly improves situation.

This is the first time I've coded in **go**.  I'm not really aware of things like PEP8/**black**/**flake8** and **pytest** that are standard in **python**

### DIFF
--- a/README.md
+++ b/README.md
@@ -199,6 +199,7 @@ It will save that information to `~/.actrc`, please refer to [Configuration](#co
       --use-gitignore                               Controls whether paths specified in .gitignore should be copied into container (default true)
       --userns string                               user namespace to use
   -v, --verbose                                     verbose output
+      --volume stringArray                          volume to bind to container
   -w, --watch                                       watch the contents of the local repo and run when files change
   -W, --workflows string                            path to workflow file(s) (default "./.github/workflows/")
 ```
--- a/cmd/input.go
+++ b/cmd/input.go
@@ -17,6 +17,7 @@ type Input struct {
 	bindWorkdir                        bool
 	secrets                            []string
 	envs                               []string
+	volumes                            []string
 	platforms                          []string
 	dryrun                             bool
 	forcePull                          bool
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -47,6 +47,7 @@ func Execute(ctx context.Context, version string) {
 	rootCmd.Flags().StringVar(&input.remoteName, "remote-name", "origin", "git remote name that will be used to retrieve url of git repo")
 	rootCmd.Flags().StringArrayVarP(&input.secrets, "secret", "s", []string{}, "secret to make available to actions with optional value (e.g. -s mysecret=foo or -s mysecret)")
 	rootCmd.Flags().StringArrayVarP(&input.envs, "env", "", []string{}, "env to make available to actions with optional value (e.g. --env myenv=foo or --env myenv)")
+	rootCmd.Flags().StringArrayVarP(&input.volumes, "volume", "", []string{}, "volume to bind to container")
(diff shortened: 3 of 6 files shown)


KnisterPeter:
Hmm, I think I don't really understand the use case.
How would that work on GitHub?
And aren't GitHub Caches good for that?

rraymondgh:
@KnisterPeter 

> Hmm, I think I don't really understand the use case.
> How would that work on GitHub?
> And aren't GitHub Caches good for that?

**conda** can be a very big overhead on resolving dependencies and downloading packages.   Try cloning **geopandas** and run **act** on it.  Many hours just to build envs when using **act** then a few minutes to run tests.  I've looked through all the options of https://github.com/marketplace/actions/setup-miniconda which is why I have moved to lock files for running locally.  I'm trying to avoid completely restructuring actions that work find on GitHub servers to run locally.   Possibly the **conda** use case with complex package dependencies  is too specific a use case for **act**.  I've extended catthehacker/ubuntu:act-latest to optimise **conda** and then using a local docker repo to serve it up to **act**

rraymondgh:
> I don't know conda and I still struggle to understand why it works fine (fast) on GitHub while it's slow with act.
> Maybe one if the other maintainers knows conda or understands your case.

It is quite difficult to get working...  I haven't got it working in a satisfactory way yet.  One thing I needed was a bit more flexibility to be able to mount volumes in **act** containers hence decided to contribute this PR back.  When I have it working I can add more commentary to this PR.  For now the change sitting in my fork is fine.

KnisterPeter:
> Wouldn't it make sense to add one --container-options flag and get rid of the extra flags, to clean up the act cli? We already have a docker option parser merged to the codebase

That sounds like a good idea.

rraymondgh:
@ChristopherHX 

> 
> * If yes
>   
>   * `act --reuse` makes it possible to persist the package installation, conda would probably detect that your packages are already installed.
>   * start the container yourself via docker cli, then execute act(master/v0.2.34) via `-P ...=-selfhosted` inside the container.
> * If no
>   
>   * execute act(master/v0.2.34) via `-P ...=-selfhosted` on your linux system, no bind mounts needed, no container

I have tried `--reuse` unfortunately https://github.com/marketplace/actions/setup-miniconda isn't smart enough to realise **conda** env and packages are already in place and then just gets errors from it...  I have now pretty much got it working with the fork associated with this PR, a **vagrant** box and a bit of dark magic with dangerous **conda** options.  Are you referring to this https://github.com/nektos/act/pull/1015 / your other project wrt `-P ...=selfhosted`?

> 
> _Wouldn't it make sense to add one `--container-options` flag and get rid of the extra flags, to clean up the act cli? We already have a docker option parser merged to the codebase_

I spotted that but honestly my **go** knowledge is really just that I'm a computer scientist / coder, so can read code on any language.  I don't have time to invest in really getting to a level in **go** to make this level of change.

I guess the conclusion is close this PR,  too specific use case to be valid....

rraymondgh:
@ChristopherHX 

> ```
> act --container-options "-v $pwd/host_environment:/host_environment" -W test.yml -P ubuntu-latest=ubuntu:latest
> ```
> 
> This avoids adding every docker cli option one by one, it's a mess.

This is excellent.  There is a piece of code still looking for **docker**.  Following doesn't stop running, but more of an FYI.
```
ERRO[0000] failed to obtain container engine info: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
```

rraymondgh:
self hosted works much better.  no need for this

Issue #1421, opened 2022-10-31
Environment variables not available in the context of the the workflow step definition


cyrus-mc:
### Bug report info

```plain text
› act --bug-report
act version:            0.2.32-20-gbc17371
GOOS:                   darwin
GOARCH:                 arm64
NumCPU:                 10
Docker host:            DOCKER_HOST environment variable is unset/empty.
Sockets found:
        /var/run/docker.sock
Config files:
        /Users/matthewceroni/.actrc:
                -P ubuntu-latest=catthehacker/ubuntu:act-latest
                -P ubuntu-22.04=catthehacker/ubuntu:act-22.04
                -P ubuntu-20.04=catthehacker/ubuntu:act-20.04
                -P ubuntu-18.04=catthehacker/ubuntu:act-18.04
        .actrc:
                -P self-hosted=catthehacker/ubuntu:act-latest
                --bind
                -v
Build info:
        Go version:            go1.19
        Module path:           command-line-arguments
        Main version:
        Main path:
        Main checksum:
        Build settings:
                -compiler:            gc
                -ldflags:             -X main.version=0.2.32-20-gbc17371
                CGO_ENABLED:          1
                CGO_CFLAGS:
                CGO_CPPFLAGS:
                CGO_CXXFLAGS:
                CGO_LDFLAGS:
                GOARCH:               arm64
                GOOS:                 darwin
Docker Engine:
        Engine version:        20.10.12
        Engine runtime:        runc
        Cgroup version:        2
        Cgroup driver:         cgroupfs
        Storage driver:        overlay2
        Registry URI:          https://index.docker.io/v1/
        OS:                    Docker Desktop
        OS type:               linux
        OS version:
        OS arch:               aarch64
        OS kernel:             5.10.76-linuxkit
        OS CPU:                5
        OS memory:             7936 MB
        Security options:
                name=seccomp,profile=default
                name=cgroupns
```

### Command used with act

```sh
act -P self-hosted=catthehacker/ubuntu:act-latest --bind
```

### Describe issue

I have the following sample workflow

```
name: example

on: [ pull_request ]

jobs:
  diff:
    runs-on: [ self-hosted ]

    name: ArgoCD diff
    steps:
      - name: checkout
        uses: actions/checkout@v3
        with:
          fetch-depth: 0

      - name: test
        id: test
        run: |
          echo "fooBar=fooValue" >> $GITHUB_ENV

      - name: output
        run: |
          env

      - name: diff
        uses: ./
        with:
[...]

PR #1458, opened 2022-11-23
fix: step env is unavailable in with property expr

ChristopherHX:
Weird inputs and env are both evaluated in the env context. This change delays the evaluation of INPUT_ env vars and makes the step env available. This feels like a hack to make act do what we expect.

Related to #1421, but more bugs are mentioned there


pkg/runner/step.go
```diff
@@ -179,7 +179,17 @@ func setupEnv(ctx context.Context, step step) error {
...
 	exprEval := rc.NewExpressionEvaluator(ctx)
 	for k, v := range *step.getEnv() {
-		(*step.getEnv())[k] = exprEval.Interpolate(ctx, v)
+		if !strings.HasPrefix(k, "INPUT_") {
+			(*step.getEnv())[k] = exprEval.Interpolate(ctx, v)
+		}
+	}
+	// after we have an evaluated step context, update the expresson evaluator with a new env context
+	// you can use step level env in the with property of a uses construct
+	exprEval = rc.NewExpressionEvaluatorWithEnv(ctx, *step.getEnv())
```
KnisterPeter, line 188: Shouldn't this then the `NewStepExpressionEvaluator`?
It seems wrong to create an `ExpressionEvaluator` with a different env.

KnisterPeter (approved):

casaroli:
Try with the following job:
```yaml
  Env-Test:
    runs-on: self-hosted
    steps:
      - run: echo LOGS_DIR=/tmp >> $GITHUB_ENV
      - name: do something
        env:
          OUTDIR: ${{ env.LOGS_DIR }}
        run: |
          echo LOGS_DIR=$LOGS_DIR
          echo OUTDIR=$OUTDIR
```

PR #1462, opened 2022-11-24
feat: `--container-options`

ChristopherHX:
This deprecates the following options
- `--privileged`
- `--container-cap-add`
- `--container-cap-drop`
- `--container-architecture`
- `--userns`

If we decide to get rid of the options, we might want to add a warning before removal.


KnisterPeter (commented):
Before we merge this there should be warnings about changed parameters, you are right.
I will approve when you add warnings.

KnisterPeter (approved):

PR #1463, opened 2022-11-24
refactor: move autoremove into the jobexecutor

ChristopherHX:
breaking: docker container are removed after job exit

In github-act-runner I directly instantiate the RunContext, but I also want to make use of autoremove due to randomized job container name.

### DIFF
--- a/pkg/runner/job_executor.go
+++ b/pkg/runner/job_executor.go
@@ -128,6 +128,24 @@ func newJobExecutor(info jobInfo, sf stepFactory, rc *RunContext) common.Executo
 			return postExecutor(ctx)
 		}).
 		Finally(info.interpolateOutputs()).
+		Finally(func(ctx context.Context) error {
+			logger := common.Logger(ctx)
+			if rc.Config.AutoRemove {
+				var cancel context.CancelFunc
+				if ctx.Err() == context.Canceled {
+					ctx, cancel = context.WithTimeout(context.Background(), 5*time.Minute)
+					defer cancel()
+				}
+
+				logger.Infof("Cleaning up container for job %s", rc.JobName)
+
+				if err := rc.stopJobContainer()(ctx); err != nil {
+					logger.Errorf("Error while cleaning container: %v", err)
+				}
+			}
+
+			return nil
+		}).
 		Finally(info.closeContainer()))
 }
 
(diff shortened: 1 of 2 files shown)


pkg/runner/job_executor.go
```diff
@@ -95,21 +95,13 @@ func newJobExecutor(info jobInfo, sf stepFactory, rc *RunContext) common.Executo
...
-		} else {
-			err := info.stopContainer()(ctx)
-			if err != nil {
-				return err
-			}
-			info.result("success")
-			logger.WithField("jobResult", "success").Infof("\U0001F3C1  Job succeeded")
+		var err error
+		if rc.Config.AutoRemove || jobError == nil {
+			err = info.stopContainer()(ctx)
```
KnisterPeter, line 101: This will fail if the job was canceled before (e.g. by os signal).
Please make sure that a fallback context is used in that case.

KnisterPeter (changes requested):

pkg/runner/job_executor.go
```diff
@@ -95,21 +95,16 @@ func newJobExecutor(info jobInfo, sf stepFactory, rc *RunContext) common.Executo
...
-			logger.WithField("jobResult", "failure").Infof("\U0001F3C1  Job failed")
-		} else {
-			err := info.stopContainer()(ctx)
-			if err != nil {
-				return err
-			}
-			info.result("success")
-			logger.WithField("jobResult", "success").Infof("\U0001F3C1  Job succeeded")
+		var err error
+		if rc.Config.AutoRemove || jobError == nil {
```
KnisterPeter, line 100: Why shouldn't this run if we have a `jobError`?
If a user wanted to `auto-remove` that should be respected I think.

(In our case this is running on a CI server and it is crucial that cleanup after a job run)

KnisterPeter (changes requested):

pkg/runner/job_executor.go
KnisterPeter: > Do we want to add a forceCancel context as field to the context, so we can force quit act by pressing ctrl+c twice? This could be the parent context of the context with a timeout.

That was the case before. The `BackgroundContext` used as a fallback was cancelable as well. That would then cancel/stop the cleanup.

pkg/runner/job_executor.go
KnisterPeter: You are right. Nevermind :smiley:

KnisterPeter (commented):
I will do some tests before approving, but the code looks fine now for me

KnisterPeter (approved):
Works as expected

pkg/runner/job_executor.go
KnisterPeter: I understood and I'm fine with that.
I just thought that it did work before, as a second cancelation was respected by act (if I remember correctly).

Issue #1459, opened 2022-11-23
GITHUB_STATE and save-state command ignored in pre step


ChristopherHX:
### Bug report info

```plain text
N/A
```

### Command used with act

```sh
./act -P self-hosted=-self-hosted -W /home/christopher/github-actions/host_environment/pkg/runner/testdata/failing-pre-step-logging/push.yml | grep STATE_
```

### Describe issue

The state is of a pre steps is lost in the void. Using it in the main stage works as expected.
Only mystate2 and mystate3 are set as env varables, mystate and mystate0 are lost.

### Workflow content

```yml
on: push
jobs:
  _:
    runs-on: self-hosted
    steps:
    - uses: christopherhx/script@main
      with:
        pre: |
          env
          echo mystate=mystateval > $GITHUB_STATE
          echo "::save-state name=mystate0::mystateval"
        main: |
          env
          echo mystate2=mystateval > $GITHUB_STATE
          echo "::save-state name=mystate3::mystateval"
        post: |
          env
```

### Relevant log output

```sh
[push.yml/_]   | STATE_mystate2=mystateval
[push.yml/_]   | STATE_mystate3=mystateval
```

PR #1466, opened 2022-11-26
refactor: fix savestate in pre steps

ChristopherHX:
This now refactores the stepresult code
- pre- and post- are no longer in stepresults
- stepresults no longer has an invalid state property
- IntraActionState now stores the GITHUB_STATE

~This uncovers a design problem:~
~you can create steps with an id prefixed with pre-~

~Reduced the change to only fix a problem, tests in https://github.com/nektos/act/pull/1457 needs to be changed to enable checking this bug~

Fixes: #1459

### DIFF
--- a/pkg/model/workflow.go
+++ b/pkg/model/workflow.go
@@ -100,6 +100,44 @@ func (w *Workflow) WorkflowDispatchConfig() *WorkflowDispatch {
 	return &config
 }
 
+type WorkflowCallInput struct {
+	Description string `yaml:"description"`
+	Required    bool   `yaml:"required"`
+	Default     string `yaml:"default"`
+	Type        string `yaml:"type"`
+}
+
+type WorkflowCallOutput struct {
+	Description string `yaml:"description"`
+	Value       string `yaml:"value"`
+}
+
+type WorkflowCall struct {
+	Inputs  map[string]WorkflowCallInput  `yaml:"inputs"`
+	Outputs map[string]WorkflowCallOutput `yaml:"outputs"`
+}
+
+func (w *Workflow) WorkflowCallConfig() *WorkflowCall {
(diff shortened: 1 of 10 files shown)


pkg/runner/action.go
```diff
@@ -371,11 +371,14 @@ func newStepContainer(ctx context.Context, step step, image string, cmd []string
 }
 
 func populateEnvsFromSavedState(env *map[string]string, step actionStep, rc *RunContext) {
-	stepResult := rc.StepResults[step.getStepModel().ID]
-	if stepResult != nil {
-		for name, value := range stepResult.State {
-			envName := fmt.Sprintf("STATE_%s", name)
-			(*env)[envName] = value
+	// design problem: you can create steps with an id prefixed with pre-
```
KnisterPeter, line 374: I think I don't get this. Do we have a test case for that?
And why is it a problem if there is a step with a `pre-` id?

KnisterPeter: Ouch, if that's the case it would be a critical error. Right.

KnisterPeter: But the `step.ID` is not changed for 'pre-steps'. It's just the `step.Name` which is prefixed.
The ID should be stable and in combination with the step stage it indicates the stage of that step.

KnisterPeter: I see, we use the `CurrentStep` in the command handler.
Wouldn't it be possible to add the `CurrentStepID` to the `RunContext` as well. That should be way more stable than a name.

KnisterPeter: > Shouldn't we remove presteps from the steps context, called stepresults? To align to actions/runner.

Would be a good idea I think.

cplee (approved):

Issue #1308, opened 2022-08-18
Consider passing `LANG=C.UTF-8`


bric3:
### Bug report info

```plain text
act version:            0.2.30
GOOS:                   darwin
GOARCH:                 amd64
NumCPU:                 8
Docker host:            DOCKER_HOST environment variable is unset/empty.
Sockets found:
        /var/run/docker.sock
Config files:           
        /Users/brice.dutheil/.actrc:
                -P ubuntu-latest=catthehacker/ubuntu:act-latest
                -P ubuntu-22.04=catthehacker/ubuntu:act-22.04
                -P ubuntu-20.04=catthehacker/ubuntu:act-20.04
                -P ubuntu-18.04=catthehacker/ubuntu:act-18.04
Build info:
        Go version:            go1.18.4
        Module path:           command-line-arguments
        Main version:          
        Main path:             
        Main checksum:         
        Build settings:
                -compiler:            gc
                -ldflags:             -X main.version=0.2.30
                CGO_ENABLED:          1
                CGO_CFLAGS:           
                CGO_CPPFLAGS:         
                CGO_CXXFLAGS:         
                CGO_LDFLAGS:          
                GOARCH:               amd64
                GOOS:                 darwin
                GOAMD64:              v1
Docker Engine:
        Engine version:        20.10.17
        Engine runtime:        runc
        Cgroup version:        1
        Cgroup driver:         cgroupfs
        Storage driver:        overlay2
        Registry URI:          https://index.docker.io/v1/
        OS:                    Docker Desktop
        OS type:               linux
        OS version:            
        OS arch:               x86_64
        OS kernel:             5.10.104-linuxkit
        OS CPU:                6
        OS memory:             3933 MB
        Security options:
                name=seccomp,profile=default
```

### Command used with act

```sh
act release --reuse --secret GITHUB_TOKEN
```

### Describe issue

When using the command above my gradle build fails due to encoding problems. While the title is using the word _consider_ I believe this is a bug as the goal of this tool is to somewhat mimic Github Action, and this behavior is hard to troubleshoot if you didn't encounter it before.

```
error: unmappable character (0xA6) for encoding US-ASCII
```

On Github Action this doesn't fail. Since it appears a locale / lang problem I ran a GA job with a step simply printing the environment `- run: env`, and indeed it appears `LANG` is set to `C.UTF-8` in GA, but it's missing when running `act`.

Indeed, passing this to the `act` command fixes the issue:

```
[...]

Issue #639, opened 2021-04-26
Issue: FileSystem Encoding


dadoonet:
## Act version

```none
act version 0.2.21
```

## Expected behaviour

I'm trying to copy some files using the maven resources plugin. I have 2 files:

* `foo.txt`
* `中文名称.txt`

When I run `mvn resources:resources`, both files should be copied to `target/classes`.

## Actual behaviour

When the filename is using non ascii characters, the file is totally ignored. So at the end I have only `foo.txt` in the `target/classes`.
It happens when running with `act`. But everything is normally copied when I run on Github Actions. See https://github.com/dadoonet/github-actions-maven/runs/2440336772?check_suite_focus=true

## Workflow and/or repository

<details>
  <summary>workflow</summary>

```none
name: Java CI with Maven

on: [push]

jobs:
  build:

    runs-on: ubuntu-latest

    steps:
      - uses: actions/checkout@v2
      - name: List the files in the target dir. We should have foo.txt and 中文名称.txt
        run: |
          ls -l src/main/resources
          files=$(ls src/main/resources | wc -l)
          if [ "$files" -lt 2 ]; then exit $files; fi
      - name: Set up JDK 15
        uses: actions/setup-java@v2
        with:
          java-version: '15'
          distribution: 'adopt'
      - name: Copy the resources using maven
        run: mvn --batch-mode resources:resources > /tmp/maven.log
      - name: List the files in the target dir. We should have 2 files. foo.txt and 中文名称.txt
        run: |
          ls -l target/classes
          files=$(ls target/classes | wc -l)
          if [ "$files" -lt 2 ]; then exit $files; fi
```

</details>

## Steps to reproduce

```sh
git clone git@github.com:dadoonet/github-actions-maven.git
cd github-actions-maven
act
```

## `act` output

I'm first highlighting the main differences. When I run `ls -l` on the source dir, I see:

With `act`: 

```none
| -rw-r--r-- 1 501 dialout 8 Apr 26 15:01  foo.txt
| -rw-r--r-- 1 501 dialout 8 Apr 26 15:01 ''$'\344\270\255\346\226\207\345\220\215\347\247\260''.txt'
```

With Github actions:

```none
2021-04-26T16:31:45.8926980Z -rw-r--r-- 1 runner docker 8 Apr 26 16:31 foo.txt
2021-04-26T16:31:45.8928565Z -rw-r--r-- 1 runner docker 8 Apr 26 16:31 中文名称.txt
```

And the target dir...

[...]

PR #1476, opened 2022-11-28
Pass `LANG=C.UTF-8` to environment

bric3:
In order to use `act` with workflows where any tools, like build tools, depends on the locale being `UTF-8`, one must pass via command line via the `--env` option.

However the Github Actions runner already passes `LANG=C.UTF-8` by default, which makes the need to explicitly pass this variable via command line un-intuitive.

The PR here aims to make `act` pass that environment variable by default.

Fixes: #1308
Possibly #639

### DIFF
--- a/pkg/runner/run_context.go
+++ b/pkg/runner/run_context.go
@@ -225,6 +225,7 @@ func (rc *RunContext) startJobContainer() common.Executor {
 		envList = append(envList, fmt.Sprintf("%s=%s", "RUNNER_OS", "Linux"))
 		envList = append(envList, fmt.Sprintf("%s=%s", "RUNNER_ARCH", container.RunnerArch(ctx)))
 		envList = append(envList, fmt.Sprintf("%s=%s", "RUNNER_TEMP", "/tmp"))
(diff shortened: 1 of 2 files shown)


bric3:
Hi thanks for the feedback !

> I want to let you know:
This PR partially enforces LANG=C.UTF-8 and overrides an env defined in the workflow file. 

Ha ok, I thought the env map was later merged, and workflow envs were overidden.
Where do you think declaring the `LANG` should happen ? 
I was also looking at this location

https://github.com/nektos/act/blob/7754ba7fcf3f80b76bb4d7cf1edbe5935c8a6bdc/pkg/runner/run_context.go#L72

Could you point me to a better place for doing that.

> This now also defines LANG on windows, I'm not shure that github actions does this on windows.

Ah good point ! I have no idea about that. I don't know how windows runners operate. What would be the best course of action if Windows runner don't.


----
I'm a novice in Go, so I'm happy for any advice in the current matter.

bric3:
Thanks @ChristopherHX

KnisterPeter (approved):

PR #1494, opened 2022-12-07
fix: use correct path to toolcache

KnisterPeter:
The toolcache on GitHub Actions need to be in
/opt/hostedtoolcache. This is the case for all
environment variables set by act, but it's not the case for the volume mounted into the container.


KnisterPeter:
We have `/opt/hostedtoolcache` hardcoded in a few places in the act codebase. Not being able to mount a volume to that means that the toolcache is unusable and broken.
Tools will be reinstalled all the time.

@nektos/act-maintainers Would it be okay for your to remove the hardcoded paths and replace with with a configurable option (while still defaulting to the hardcoded value).

KnisterPeter:
Cool. I think this PR is obsolete then and I can close it.

KnisterPeter:
I thought your changes will resolve the more general toolcache issue.
I'll keep this PR open then until #1503 is merged and see if I can continue on this.

KnisterPeter:
@ZauberNerd I think we can close that, can't we?

lawndoc:
> While the tests are now passing. We would still have to move the tools of the default docker images out of the mount point.
> 
> Otherwise these docker images end up with an empty tool_cache and the preinstalled `node` etc. are away.

Why do you need the mount point to be empty? Doesn't docker initialize new named volumes using the contents of the image?

lawndoc:
Thank you for your work on this! It's greatly appreciated!

PR #1472, opened 2022-11-27
refactor: fix add-path / GITHUB_PATH commands

ChristopherHX:
The existing code applies prepend path for all actions, while it shouldn't do it for docker actions.
Additionally GITHUB_PATH shouldn't update the env directly, instead add to extrapath

Seems like the old ::add-path:: never worked correctly, because the first add-path always wins e.g.
- add node12 as node
- use node ver 12
- add node16
- still in version 12, node16 added to a later path component after node12

Duplicated path are now eliminated.

Resolves #1384

Still to complex for get it ready.

PR #1496, opened 2022-12-08
feat: JobLoggerFactory

ChristopherHX:
- Remove overriding io.Stdout in TestMaskValues to prevent deadlock in GitHub Actions
- Print log of TestMaskValues in Test

_captureOutput is unsafe and causes a deadlock in https://github.com/nektos/act/pull/1472_

_I tested merging this change into #1472 and it unblocks the failing test._


KnisterPeter (approved):

KnisterPeter:
Looks fine. Strange that it didn't deadlock earlier, as that code is in master for a long time already.

PR #1497, opened 2022-12-09
feat: Support "result" on "needs" context.

### LINKED ISSUE #1367: Support needs.*.result
### Act version

0.2.31

### Feature description

Act is missing support for the `result` property in the needs context object. This property tells you if a job succeeded or failed.

https://docs.github.com/en/actions/learn-github-actions/contexts#needs-context

I expect this workflow to work, but `needs.build.result` is not defined when running with act.
```
name: Build and deploy
on: push

jobs:
  build:
    runs-on: ubuntu-latest
    outputs:
      build_id: ${{ steps.build_step.outputs.build_id }}
    steps:
      - uses: actions/checkout@v3
      - name: Build
        id: build_step
        run: |
          ./build
          echo "::set-output name=build_id::$BUILD_ID"
  deploy:
    needs: build
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v3
      - run: ./deploy --build ${{ needs.build.outputs.build_id }}
  debug:
    needs: [build, deploy]
    runs-on: ubuntu-latest
    if: ${{ needs.build.result == 'failure' }}
    steps:
      - uses: actions/checkout@v3
      - run: ./debug
 ```

### PR DESCRIPTION
**Pull request by aholmes**
This change adds "result" to a job's "needs" context, as documented [here](https://docs.github.com/en/actions/learn-github-actions/contexts#needs-context). `act` currently tracks the success/failure/cancelled status of a job, but does not include this value the `needs` context.

Fixes #1367

### DIFF
--- a/pkg/exprparser/interpreter.go
+++ b/pkg/exprparser/interpreter.go
@@ -20,8 +20,13 @@ type EvaluationEnvironment struct {
 	Secrets  map[string]string
 	Strategy map[string]interface{}
 	Matrix   map[string]interface{}
-	Needs    map[string]map[string]interface{}
+	Needs    map[string]Needs
 	Inputs   map[string]interface{}
+}
+
+type Needs struct {
+	Outputs map[string]string `json:"outputs"`
+	Result  string            `json:"result"`
 }
 
 type Config struct {
--- a/pkg/exprparser/interpreter.go
+++ b/pkg/exprparser/interpreter.go
@@ -20,7 +20,7 @@ type EvaluationEnvironment struct {
 	Secrets  map[string]string
 	Strategy map[string]interface{}
 	Matrix   map[string]interface{}
-	Needs    map[string]map[string]map[string]string
+	Needs    map[string]map[string]interface{}
 	Inputs   map[string]interface{}
 }
 
--- a/pkg/exprparser/interpreter_test.go
+++ b/pkg/exprparser/interpreter_test.go
@@ -594,12 +594,12 @@ func TestContexts(t *testing.T) {
 		Matrix: map[string]interface{}{
 			"os": "Linux",
 		},
-		Needs: map[string]map[string]interface{}{
+		Needs: map[string]Needs{
 			"job-id": {
-				"outputs": map[string]string{
+				Outputs: map[string]string{
 					"output-name": "value",
 				},
-				"result": "success",
+				Result: "success",
 			},
 		},
 		Inputs: map[string]interface{}{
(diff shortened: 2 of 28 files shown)

### CODE REVIEW

File: pkg/exprparser/interpreter.go [hunk 1]
```diff
@@ -20,7 +20,7 @@ type EvaluationEnvironment struct {
 	Secrets  map[string]string
 	Strategy map[string]interface{}
 	Matrix   map[string]interface{}
-	Needs    map[string]map[string]map[string]string
+	Needs    map[string]map[string]interface{}
```
KnisterPeter, line 23: Code looks fine, but I would prefer if this uses a more concrete type.
The actual type would be `map[string]Needs`
with `Needs` like this:

```go
type Needs struct {
  outputs map[string]string
  result  string
}
```

Isn't it?

KnisterPeter (changes requested):
Maybe I'm wrong with that not sure.

pkg/exprparser/interpreter.go
aholmes: > I would prefer if this uses a more concrete type.

Me too!

> The actual type would be `map[string]Needs` with `Needs` like this:
> 
> ```go
> type Needs struct {
>   outputs map[string]string
>   result  string
> }
> ```

Thanks for this. I'm not familiar with Go, so wasn't sure what to do about this. I've pushed the change.

aholmes:
> Code looks fine, would it be possible to add a testcase for this?

Can you point me in the right direction to do so? I can reference @ChristopherHX's suggestion too.

aholmes:
> I got no `yes`/`no` about adding an integration test I'm not touching this PR without your explicit consent.

By all means! I am not sure where to add an integration test for this change, so please add one or point me in the right direction and I can take care of it.

aholmes:
> > I am not sure where to add an integration test for this change
> 
> You can add a folder with a push.yml file in https://github.com/nektos/act/tree/master/pkg/runner/testdata
> 
> Then add a line here to reference it https://github.com/nektos/act/blob/master/pkg/runner/runner_test.go#L188.

Done. How does it look?

dan-knight (dismissed):

aholmes:
@KnisterPeter can you re-review?

KnisterPeter:
I'm on sick leave and will when I'm back at work.
Other maintainers may approve as well

cplee (dismissed):

cplee (dismissed):

aholmes:
Sorry for those request spams, not sure what happened there.

aholmes:
@cplee I think I've got those conflicts and tests fixed up now. 🤞

cplee (approved):

KnisterPeter (approved):
Awesome. Thank you @aholmes

Issue #1390, opened 2022-10-14
Docker Desktop 22.06.0-beta: invalid filter 'reference'


drew-viles:
### Bug report info

```plain text
act version:            0.2.32                                          
GOOS:                   linux                                           
GOARCH:                 amd64                                           
NumCPU:                 2                                               
Docker host:            DOCKER_HOST environment variable is unset/empty.
Sockets found:                                                          
        /var/run/docker.sock                                            
Config files:                                                           
        /home/drew/.actrc:                                              
                -P ubuntu-latest=catthehacker/ubuntu:act-latest         
                -P ubuntu-22.04=catthehacker/ubuntu:act-22.04           
                -P ubuntu-20.04=catthehacker/ubuntu:act-20.04
                -P ubuntu-18.04=catthehacker/ubuntu:act-18.04
Build info:
        Go version:            go1.18.6
        Module path:           github.com/nektos/act
        Main version:          (devel)
        Main path:             github.com/nektos/act
        Main checksum:
        Build settings:
                -compiler:            gc
                -ldflags:             -s -w -X main.version=0.2.32 -X main.commit=21484b5c1ecbd9e1ddb6a57debd7731a3e1fc25f -X main.date=2022-10-01T03:01:40Z -X main.builtBy=goreleaser
                CGO_ENABLED:          0
                GOARCH:               amd64
                GOOS:                 linux
                GOAMD64:              v1
                vcs:                  git
                vcs.revision:         21484b5c1ecbd9e1ddb6a57debd7731a3e1fc25f
                vcs.time:             2022-09-29T05:59:52Z
                vcs.modified:         false
Docker Engine:
        Engine version:        22.06.0-beta.0-372-gd3bb8227ce.m
        Engine runtime:        runc
        Cgroup version:        1
        Cgroup driver:         cgroupfs
        Storage driver:        stargz
        Registry URI:          https://index.docker.io/v1/
        OS:                    Docker Desktop
        OS type:               linux
        OS version:
        OS arch:               x86_64
        OS kernel:             5.10.102.1-microsoft-standard-WSL2
        OS CPU:                2
        OS memory:             3924 MB
        Security options:
                name=seccomp,profile=builtin
```

### Command used with act

```sh
act
or
act -j kube_image_builder_job

Both create the same output
```

### Describe issue

I just expected it to be able to run.

### Workflow content

```yml
name: Lint and Test Charts

on: [push]


jobs:
  lint-chart:
    runs-on: ubuntu-latest
[...]

PR #1501, opened 2022-12-10
refactor: remove docker image list reference filter

ChristopherHX:
~_Should I preserve the public ImageExistsLocally signature and make a private helper function?_~, reverted the new return value.

This uses docker inspect and remove docker list.
You cannot have multiple images of the same name+tag, regardless of platform metadata

Fixes https://github.com/nektos/act/issues/1390


KnisterPeter (approved):
:+1:

cplee (approved):

PR #1502, opened 2022-12-11
Remove red color from logging messages

patinthehat:
This PR removes the color `red` from logging messages as red is typically associated with error messages, and can result in confusion with lots of output from many containers. 

Avoiding red text removes the ambiguity of red text that is not an error message.

### DIFF
--- a/pkg/runner/logger.go
+++ b/pkg/runner/logger.go
@@ -17,7 +17,6 @@ import (
 
 const (
 	// nocolor = 0
-	red     = 31
 	green   = 32
 	yellow  = 33
 	blue    = 34
@@ -33,7 +32,7 @@ var mux sync.Mutex
 func init() {
 	nextColor = 0
 	colors = []int{
-		blue, yellow, green, magenta, red, gray, cyan,
+		blue, yellow, green, magenta, gray, cyan,
 	}
 }
 


patinthehat:
> It might be better to add an alternative color instead of only removing one.

@ChristopherHX I'm not sure there is another color that can be added while staying within the default ANSI color code range. I'd be willing to implement using something like the 256-color palette in another PR, though.

patinthehat:
> We're using it for jobs names only, also I don't think it's good change to make given it's affecting only one person

@catthehacker  I strongly disagree.  It is common practice to display fatal error messages in red text, and if I've had confusion here I am sure others have as well.  What if we moved to a 256-color palette and I added additional colors to offset the removal of `red`?

patinthehat:
> Source?
I haven't polled anyone, but it's a common assumption.  It's also a standard in style guidelines, such as Github's Primer guidelines: [primer.style/primitives/colors](https://primer.style/primitives/colors#danger)

My point is just that, imo, using red text is not the best design decision because red commonly means "error".  I'd be happy to open another PR that adds more ANSI colors, but I believe that removing red from the palette is the right decision.  Just my thoughts on it - thanks for your quick response! :+1:

PR #1514, opened 2022-12-18
revert: deprecation of containerArchitecture

ChristopherHX:
This option isn't part of parsed docker cli flags of act.

_Nobody protected me doing nonsense, I really expected that the `--platform` has been parsed._

_Even if this flag has been parsed by docker options logic, this flag wouldn't be applied to docker_pull and other parts of act_


KnisterPeter (approved):
> Nobody protected me doing nonsense, I really expected that the --platform has been parsed

Happens to all of us :wink:. Don't want to name my previous failures here.

PR #1524, opened 2022-12-19
Input

### LINKED ISSUE #1021: Enhancement: A flag for providing inputs
Would be great if I could say `--input name=value` instead of having to create a file as per https://github.com/nektos/act/issues/332.

I am using GitHub Actions for tasks like "deploy a certain branch to staging", etc, and nearly all my actions currently require inputs. (While, on the other hand, I don't have any `on: push` actions.)

shubhbapna:
Fixes #1021 

Added 2 flags `--input` and `--input-file` to manually pass inputs

You will notice that in the `New` function for `runner`, I only set the inputs if event payload was not defined. IMO, if the user is passing a payload, then they can set their input in that json file itself rather than passing it separately using the flags.


mikerjacobi (approved):
This is exactly what I am looking for as well. This PR looks great

cplee (approved):

cplee (approved):

PR #1525, opened 2022-12-20
feat: add remote reusable workflows

### LINKED ISSUE #826: Issue: support for reusable workflows
ref: https://github.com/github/roadmap/issues/98

GitHub is going to support reusable workflows (similar to Azure DevOps Pipelines)

> How will it work?
> To reuse a workflow, ensure it listens to a new `workflow_call` event trigger, and then reference it with the familiar `uses` syntax. Optionally, use `inputs` and `outputs` to pass data between them, just like with actions.
> 
> For example:
> 
> ```yaml
> name: Build & Release App
> 
> on:
>   push:
> 
> jobs:
>   build:
>     uses: example-org/automation/workflows/build.yml@1
> 
>   release:
>     needs: build
>     uses: example-org/automation/workflows/release.yml@1
>     with:
>       artifact-url: jobs.build.outputs.artifact-url
> ```

KnisterPeter:
This changes adds cloning of a remote repository to
run a workflow included in it.

Closes #826


gblikas:
@KnisterPeter is there a corresponding PR or release that supports the local Github workflow references, as in: https://github.blog/changelog/2022-01-25-github-actions-reusable-workflows-can-be-referenced-locally/

Issue #1528, opened 2022-12-20
PATH is overwritten


KnisterPeter:
### Bug report info

```plain text
act version:            0.2.34-18-g4989f44
GOOS:                   linux
GOARCH:                 amd64
NumCPU:                 12
Docker host:            DOCKER_HOST environment variable is unset/empty.
Sockets found:
	/var/run/docker.sock
Config files:           
	/home/markusw/.actrc:
		-P ubuntu-latest=node:12-buster-slim
		-P ubuntu-20.04=node:12-buster-slim
		-P ubuntu-18.04=node:12-buster-slim
		-P ubuntu-16.04=node:12-stretch-slim
Build info:
	Go version:            go1.19.4
	Module path:           command-line-arguments
	Main version:          
	Main path:             
	Main checksum:         
	Build settings:
		-compiler:            gc
		-ldflags:             -X main.version=0.2.34-18-g4989f44
		CGO_ENABLED:          1
		CGO_CFLAGS:           
		CGO_CPPFLAGS:         
		CGO_CXXFLAGS:         
		CGO_LDFLAGS:          
		GOARCH:               amd64
		GOOS:                 linux
		GOAMD64:              v1
Docker Engine:
	Engine version:        20.10.21+dfsg1
	Engine runtime:        runc
	Cgroup version:        2
	Cgroup driver:         systemd
	Storage driver:        overlay2
	Registry URI:          https://index.docker.io/v1/
	OS:                    Debian GNU/Linux bookworm/sid
	OS type:               linux
	OS version:            
	OS arch:               x86_64
	OS kernel:             6.0.0-5-amd64
	OS CPU:                12
	OS memory:             31734 MB
	Security options:
		name=apparmor
		name=seccomp,profile=default
		name=cgroupns
```

### Command used with act

```sh
act
```

### Describe issue

The PATH variable is overwritten (or reset) when an action is executed wich creates some output.
The workflow below prints the PATH in two places:

1. after `actions/setup-node` and the node bins are in the PATH
2. after the `fea/changelog-action`. In that place the PATH was reset.

The changelog actions is private, but in essence it just defines an output string. It does not install tools or modifies the PATH.

### Workflow content

```yml
name: test
on: push
jobs:
  test:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/setup-node@v3
        with:
          node-version: 18
      - name: debug 0
        run: |
          echo PATH=$PATH
      - id: changelog
        uses: fea/changelog-action@v1
[...]

PR #1529, opened 2022-12-20
test: define test case of path issues

KnisterPeter:
Test case for #1528


KnisterPeter:
@ChristopherHX since you did work a lot on the PATH handling recently, do you have an idea where this breaks? This is kind of critical for us and I would like to build a fix tomorrow. Any starting point would help a lot.

PR #829, opened 2021-09-27
enable using worktrees

lvlrt:
- Enables using act in a worktree of Git (needs to follow the .git file directions)

PR #1530, opened 2022-12-20
Feature/allow worktrees

### LINKED ISSUE #657: Enhancement: working with git worktree
## Describe feature

For a working act workflow, when you are using `git worktree` ACT does not work. It does not fill repository data, certainly because it does not follow links in `.git` file in a git worktree.

If there is a workaround, i'd take it! :)

Example:
```
WARN[0000] unable to get git repo: unable to find git repo
WARN[0000] unable to get git revision: unable to find git repo
WARN[0000] unable to get git ref: unable to find git repo
WARN[0000] unable to get git repo: unable to find git repo
WARN[0000] unable to get git revision: unable to find git repo
WARN[0000] unable to get git ref: unable to find git repo
```

```
$ cat .git
gitdir: /home/xxx/yyy/zzz.git/worktrees/dev
```

lavigne958:
Handle cases where the user uses worktrees and not a regular git clone.

This PR uses some work previously done by @lvlrt and my contribution to make it work in any cases regardless of the type of worktrees.

this PR handle 2 cases:
1. when the git dir is a regular git clone: so the worktree contains a file `.git` that contains the path to the git data inside the regular git clone. usually under: `/path/to/repo/.git/worktrees/worktre_name`
2. When the git dir is a barre clone (so no actual files, only git data) and so the worktree contains a file `.git` that contains the path to the git data directly. usually under: `/path/to/repo/commondir`

This work fine, I tried it using both worktree types (actually both git clone types) and this code was pushed using a worktree and I managed to run `act` on it.

closes #657 
Uses commits from #829

### DIFF
--- a/pkg/common/git/git.go
+++ b/pkg/common/git/git.go
@@ -54,30 +54,33 @@ func (e *Error) Commit() string {
 // FindGitRevision get the current git revision
 func FindGitRevision(ctx context.Context, file string) (shortSha string, sha string, err error) {
 	logger := common.Logger(ctx)
-	gitDir, err := findGitDirectory(file)
+
+	gitDir, err := git.PlainOpenWithOptions(
+		file,
+		&git.PlainOpenOptions{
+			DetectDotGit:          true,
+			EnableDotGitCommonDir: true,
+		},
+	)
+
+	if err != nil {
+		logger.WithError(err).Error("path", file, "not located inside a git repository")
+		return "", "", err
+	}
+
+	head, err := gitDir.Reference(plumbing.HEAD, true)
 	if err != nil {
 		return "", "", err
 	}
 
-	bts, err := os.ReadFile(filepath.Join(gitDir, "HEAD"))
-	if err != nil {
-		return "", "", err
+	if head.Hash().IsZero() {
+		return "", "", fmt.Errorf("HEAD sha1 could not be resolved")
 	}
 
-	var ref = strings.TrimSpace(strings.TrimPrefix(string(bts), "ref:"))
-	var refBuf []byte
-	if strings.HasPrefix(ref, "refs/") {
-		// load commitid ref
-		refBuf, err = os.ReadFile(filepath.Join(gitDir, ref))
-		if err != nil {
-			return "", "", err
-		}
-	} else {
-		refBuf = []byte(ref)
-	}
+	hash := head.Hash().String()
 
-	logger.Debugf("Found revision: %s", refBuf)
-	return string(refBuf[:7]), strings.TrimSpace(string(refBuf)), nil
+	logger.Debugf("Found revision: %s", hash)
+	return hash[:7], strings.TrimSpace(hash), nil
 }
 
 // FindGitRef get the current git ref
--- a/pkg/common/git/git.go
+++ b/pkg/common/git/git.go
@@ -17,6 +17,7 @@ import (
 	"github.com/go-git/go-git/v5"
 	"github.com/go-git/go-git/v5/config"
 	"github.com/go-git/go-git/v5/plumbing"
+	"github.com/go-git/go-git/v5/plumbing/storer"
 	"github.com/go-git/go-git/v5/plumbing/transport/http"
 	"github.com/mattn/go-isatty"
 	log "github.com/sirupsen/logrus"
@@ -86,12 +87,8 @@ func FindGitRevision(ctx context.Context, file string) (shortSha string, sha str
 // FindGitRef get the current git ref
 func FindGitRef(ctx context.Context, file string) (string, error) {
 	logger := common.Logger(ctx)
-	gitDir, err := findGitDirectory(file)
-	if err != nil {
-		return "", err
-	}
-	logger.Debugf("Loading revision from git directory '%s'", gitDir)
 
+	logger.Debugf("Loading revision from git directory")
 	_, ref, err := FindGitRevision(ctx, file)
 	if err != nil {
 		return "", err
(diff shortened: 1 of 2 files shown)


KnisterPeter (commented):
Would it somehow possible to create a test for this?

lavigne958:
> Would it somehow possible to create a test for this?

Hi sure it is, I am looking at it right now.

for sure I can add some unit test on the function I patch. 
About functional tests, I can't tell yet.

KnisterPeter (commented):
The code looks fine.

lavigne958:
> The code looks fine.

thanks, I am working on a test suite that covers all scenarios but it's not that easy as the function actually reads the file system, I must init git repositories in the exact state needed by the test. 

It takes time but I'm getting there.

lavigne958:
Hi quick update: i made progress writing the test scenario but I'm still facing issues with the test case of a bare git clone and I don't use it so I need to better understand how it works and how i can create a worktree of a bare clone during the test setup .

Once I have it sorted the PR I will update

lavigne958:
@KnisterPeter  I am facing an issue with this PR.

I have a test suite that runs and passes, that part is solved.

My problem is:

- we use the function `findGitDirectory` for multiple purposes
- In some cases we read the `HEAD` file to identify the currently checkout ref
- in other cases we read the `config` file to read the remote URLs
- in other cases we use the `go-git` package to open the git repository

When using worktrees the config, and the _open repository_ parts should be done using the root level git directory (so in a regular clone, the `.git` folder as we know it, in the bare clone, the actual bare git clone folder) **but** when reading the file `HEAD` to identify the ref checked out we should pass the worktree subfolder (located under `(.git|bare_clone)/worktress/<worktree name>/HEAD`).

so if we want to be compatible with worktrees (which we should in my opinion) we must adapt the code to handle all above cases correctly.

How would you like to proceed about this ? 

I thought about a couple of solutions but first I'd like to have you opinion about it.

KnisterPeter:
@lavigne958 I would suggest to copy the `findGitDirectory` function and rename it for the specific use-cases. That would allow to have slightly different implementations while not making the change too big I guess.

What other options do you see?

lavigne958:
> 

That's one of my proposal, it make sense and it makes it obvious when you look at the function call what you expect (read the HEAD for the ref, or actually read the repository configuration)

lavigne958:
> > What other options do you see?
> 
> I would replace the following cases with go-git or git cli
> 
> > * In some cases we read the HEAD file to identify the currently checkout ref
> > * in other cases we read the config file to read the remote URLs
> 
> I don't like that act reads internal git configuration files without using an external library.
> 
> I have a git cli reference implementation to get point 1&2, only needs to be rewitten in go. Not a big deal for me.

That would solve quiet a few problems too, I will have a look at the `go-git` package and check if it can actually open a worktree, and handle the file `.git` and if so, that makes it surely much easier to resolve the issue.

What we should check here too is:

- what file is given when searching for the git directory

one of the purposes of the function `findGitDir` is to go up each layers from a given path until it finds the root git directory where is located the `.git/` directory and return that path.

depending on the use case the `go-git` packages might actually need the same process to be done first, which brings us back to point 1 :laughing:

lavigne958:
> > What other options do you see?
> 
> I would replace the following cases with go-git or git cli
> 
> > * In some cases we read the HEAD file to identify the currently checkout ref
> > * in other cases we read the config file to read the remote URLs
> 
> I don't like that act reads internal git configuration files without using an external library.
> 
> I have a git cli reference implementation to get point 1&2, only needs to be rewritten in go. Not a big deal for me.

That's right, for the first 2 points, we could use the `go-git` packages that should be able to resolve that for us.

the only issue is: we must find the root repository folder, where the `.git` folder is located. This is the purpose of the function `findGitDir`, so yes a external package can do it instead of `act` doing it manully but is this package able to open a worktree ? :thinking:  I will check that and see if it works, if so that's good if not, that's going to be trickier or we will need some manual steps.

lavigne958:
> 

I ran a quick test, and yes this option allows us to walk parent folders until we find the repository.

but when this commands are run from the worktree it returns nothing.
It does not fail to open a worktree but the remotes/configs/ref etc are empty :disappointed: 

So we still need to identify if we are in a worktree and if so, return the path to the original git dir (regular clone or bare clone)

then workout what information we want, checking out the git plugin I noticed that we can interract with worktrees but we don't have much options. so that does not solve our issue yet.

lavigne958:
> > but when this commands are run from the worktree it returns nothing.
> 
> What about opening the real repository? git cli doesn't have such a bug. The asumption of filesystem.Storage can be enshured by a test.
> 
> ```go
> 	storage, _ := repo.Storer.(*filesystem.Storage)
> 	fs := storage.Filesystem()
> 
> 	repo, _ = git.PlainOpenWithOptions(filepath.Join(fs.Root(), ".."), &git.PlainOpenOptions{
> 		DetectDotGit: true,
> 	})
> ```
> 
> Not clean, but we assume less about git's internal structure than before.

That works for some parts of our problem.

I think we can do a combination of both solutions.

Use the above solutions to read:

- the configuration (the remote URLs for example)
- open the actual repository for some operations

Then use the `go-git` package to open the repo, then if the HEAD ref is empty, check if we are in a worktree and then just read the file under `fs.Root() / HEAD` as the git package points to the right place: `/path/to/repo/.git/worktree/<worktree_name>/`

lavigne958:
after the above discussion this PR has been reworked:

- remove the function `findGitDir`
- use the package `go-git/go-git` for any git operation
- add extra tests if necessary
- Tested with worktrees :heavy_check_mark: 

I rebased the branch onto the master branch and dropped any previous commit, started from scratch. 
Each commit introduces the replacement of a function that used to use `findGitDir` and now used the `go-git` package instead.

lavigne958:
Should I add more tests to increase coverage on the patch i introduce?

lavigne958:
> _I will test this carefully, I don't trust our CI to much_
> 
> I think you fixed the endless recursion on windows if your cwd is on `d:\` or `e:\`.
> 
> > Should I add more tests to increase coverage on the patch i introduce?
> 
> In my opinion, only if I find something odd while testing.

That's a very good point I don't have access to windows computer so you'd better carefully test it on windows if you can just in case.

lavigne958:
> Great work, I like your change. No additional tests needed from my side, I couldn't find any issues.
> I see you found a way to make go-git open worktrees correctly, all these hidden flags.

Thank you very much 😁

You pointed them out I just played with them 😉

If it's good for you, then it's good for me.

KnisterPeter (approved):
Awesome. Thank you @lavigne958 :heart:

PR #1557, opened 2023-01-12
feat: add support for building docker actions with private registries

ZauberNerd:
This commit adds a new `LoadDockerAuthConfigs` function, which loads all registry auths that are configured on the host and sends them with the build command to the docker daemon.

This is needed in case act builds a docker action and the images referenced in that docker action are located on private registries or otherwise require authentication (e.g. to get a higher rate limit).

The code is adapted from how the docker cli works: https://github.com/docker/cli/blob/257ff41304bf121bdf1acdf00a1c7a896ed038d1/cli/command/image/build.go#L323-L332

### DIFF
--- a/pkg/container/docker_auth.go
+++ b/pkg/container/docker_auth.go
@@ -38,3 +38,24 @@ func LoadDockerAuthConfig(ctx context.Context, image string) (types.AuthConfig, 
 
 	return types.AuthConfig(authConfig), nil
 }
+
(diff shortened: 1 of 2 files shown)


KnisterPeter (approved):
Looks good, but there seems to be some formatting issues

KnisterPeter (approved):

cplee (approved):

Issue #1069, opened 2022-03-23
Enhancement: --version and --bugreport should report tag or commit hash for compiled Act instances


maxsu:
## Enhancement

When act is built from scratch or installed with `go install`, it would be helpful for the `--version` and `--bug-report` flags to report the tag and/or commit hash (and indicate act was compiled).

## Existing Behavior

 `--bug-report` leaves the version field empty, and the `--version` flag is not compiled:

```
$ act --version
Error: unknown flag: --version
```

```
$ act --bug-report
act version:
GOOS:                   linux
GOARCH:                 amd64
NumCPU:                 8
.
.
.
```

This leaves some room for surprise, and requires an workaround (see below).

## System information

Act version: 0.2.26
Install route: `go install github.com/nektos/act@latest`
OS: Pop_OS 21.10
Arch: x86-64

## Workaround
I can find the tag like so:

```bash
$ ls $GO_PATH/pkg/mod/github.com/nektos
act@v0.2.26
```

maxsu:
Is the existing behavior designed to encourage people to file bug reports against official versions?

In that case, consider the requested feature makes it easier to comply with the issue template requirement:

    if you've built `act` yourself, please provide commit hash
    
There are good and bad sides to this:

```diff

Pros:
+ The feature makes it easier to report bugs
+ This may increase the rate of bug reports
+ Of those reports, more will be in compliance /w the issue template

Cons: 
- The feature lets inexperienced users comply /w the issue template
- ... So the additional issues may otherwise be of lower quality
```

catthehacker:
It already provides version, but you have to supply it via ldflags

maxsu:
@catthehacker thanks for the lead!

 I can now set Act to a very special version:
 
```bash
$ go install --ldflags "-X main.version=6.6.6.😈" github.com/nektos/act@latest
$ act --version
act version 6.6.6.😈
```

The project has two ways of auto-setting the version:

1. Manual builds via Makefile: `Makefile` lines [2](https://github.com/nektos/act/blob/master/Makefile#L2) and [26](https://github.com/nektos/act/blob/2bb3e74616f86ece6a11cd854ad54701542410fd/Makefile#L26) set the version as described in [this article](https://vincent.bernat.ch/en/blog/2019-makefile-build-golang). If the user is not at a tag, the logic in line 2 gives us a hash.

[...]

PR #1562, opened 2023-01-14
feat: add check for newer versions

cplee:
Notify user when a newer version of `act` is available. Support disabling of version check by setting environment variable `ACT_DISABLE_VERSION_CHECK=1`


KnisterPeter (commented):
Looks good.
I have two questions:

1. The logger does output only text. If act is started with '--json' it should respect that. Otherwise the log would not be parsable
2. Why is the go-git patch by ZauberNerd removed?

cplee:
Good catch @KnisterPeter . I updated to honor the `--json` for log format. Also reverted the change to go.mod (I was having issues with that patch and wasn't sure it was still needed?)

KnisterPeter (approved):
👍

KnisterPeter:
We should check with @ZauberNerd.
What where the issues?

cplee:
It was looking for tag `v5.4.3` in his fork, which didn't exist

cplee:
@ChristopherHX - updated to keep the source version aligned with release version. This will avoid alerting on new versions if  built from source...unless there is in fact a new version available. Fixed #1069 too...

KnisterPeter (approved):

ZauberNerd:
Oh, I rebased my PR https://github.com/go-git/go-git/pull/485 last month, that's probably why it couldn't find the tag anymore.

Issue #1567, opened 2023-01-18
Reduce logs in act CI


lavigne958:
### Act version

master

## Overview

Reduce the log footprint in the CI (currently using github action).
The logs from the CI workflow `test-linux` reaches over 70K log lines. 
It could help putting some of them in a log file

## Details

The current CI prints a lot of informations: the go test logs (test name, test result,...) and the logs from `act` itself.

We reach over 70K log lines in each run, just for the step `go test`. We could reduce that by putting the logs from `act` into a dedicated file. 

## Limitations

The issue/limitation with this is that if we need the logs from `act` in order to understand a failing test we need to be able to retrieve that log file for that test (yes, having a log file for each test is better :wink: ).

That means we must some how upload that files to some packages storage and have it cleaned after some time (15 days is enough ?? ) that require extra work on the CI (cron jobs running that list packages and clean log files that are too old).

We can find other solutions for that, as long as we split the logs from the `go test` command from the `act` logs we'll make the CI output much easier to read.

lavigne958:
Yes, when trying to contribute to the project, if a step fails it makes it much easier to read only golang output to identify the faulty test and then read the test output elsewhere. Or specifically run that test locally, or something else depending on the solution of choice.

In a word: it makes it easier for people to contribute.

PR #1580, opened 2023-01-23
feat: step summary of test results

ChristopherHX:
Examples
- https://github.com/nektos/act/actions/runs/3988464133/attempts/1#summary-10827280870
- https://github.com/nektos/act/actions/runs/3989524486/attempts/1#summary-10830484825

Logs of each tests are uploaded additionally as an artifact

Closes https://github.com/nektos/act/issues/1567

### DIFF
--- a/.github/actions/run-tests/action.yml
+++ b/.github/actions/run-tests/action.yml
@@ -1,77 +1,77 @@
-name: 'run-tests'
-description: 'Runs go test and upload a step summary'
+name: "run-tests"
+description: "Runs go test and upload a step summary"
 inputs:
   filter:
-    description: 'The go test pattern for the tests to run'
+    description: "The go test pattern for the tests to run"
     required: false
-    default: ''
+    default: ""
   upload-logs-name:
-    description: 'Choose the name of the log artifact'
+    description: "Choose the name of the log artifact"
     required: false
     default: logs-${{ github.job }}-${{ strategy.job-index }}
   upload-logs:
-    description: 'If true uploads logs of each tests as an artifact'
+    description: "If true uploads logs of each tests as an artifact"
     required: false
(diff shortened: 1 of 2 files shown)


lavigne958:
> @lavigne958 What is your opinion about this change? I left the workflow log as it was.

Hi thank you for this feature, it helps a lot.
the markdown at the bottom of the checks summary is very nice.

Though I did not expect something that complexe using JS, there are many way to achieve the same goal in software development, I thought about something like changing the stdout stream for all tests using the `TestMain` feature that setup for all tests, then redirect the stdout to some files `os.Stdout = ....`.

If it works it's fine though.

.github/actions/run-tests/action.yml
```diff
@@ -0,0 +1,77 @@
...
+description: 'Runs go test and upload a step summary'
+inputs:
+  filter:
+    description: 'The go test pattern for the tests to run'
+    required: false
+    default: ''
+  upload-logs-name:
+    description: 'Choose the name of the log artifact'
+    required: false
+    default: logs-${{ github.job }}-${{ strategy.job-index }}
```
lavigne958, line 11: I am not too sure the `strategy` object exists in a composite action. quiet a few objects don't exists in a composite action which why we need to pass things using inputs.

I'll suggest to use the run number and the run attemps to make it unique `github.run_id` and `github.run_attemps`: https://docs.github.com/en/actions/learn-github-actions/contexts#github-context

lavigne958: my bad, should be fine then.

thank you for the improvement, it looks awesome on the summary page.

lavigne958 (approved):

KnisterPeter (approved):

PR #1582, opened 2023-01-24
feat: allow overriding of `GITHUB_` env variables

### LINKED ISSUE #1424: Allow GITHUB_REPOSITORY to be overridden
### Act version

0.2.33

### Feature description

When developing a github action, it would be cool to be able to write end to end tests with act. We would need the ability to override GITHUB_REPOSITORY (and SHA) to emulate a user using our action in their repo.

Currently GITHUB_REPOSITORY is automatically set to the action we are developing the action in.

shubhbapna:
Fixes #1424 

By loading the `GITHUB_` env variables before we load the user provided env variables, we can let them override these variables.

I ended up doing a little bit of refactoring as well to keep the linter happy about the cyclomatic complexity of  `getGithubContext`. I shifted the logic of setting some of the variables to `github_context.go`


shubhbapna:
@ChristopherHX any idea how `rc.Config.Env` is different from `rc.Env`? Is `rc.Env` the env variables of the current job only and `rc.Config.Env` are the global env variables passed by the user via cli options?

KnisterPeter (approved):

adnanjpg:
hi, thanks for the work. trying to use this now, would this allow `act --env GITHUB_SHA="mycustomsha" ...` to work and expect to see the value of `GITHUB_SHA` as `mycustomsha` in my action file? doing it like this does not work and keeps providing me the real SHA every time. am I doing something wrong?

PR #1597, opened 2023-02-02
Update max container name length

artificial-aidan:
I _think_ I understood this function correctly. Please chime in if I was wrong. I had an issue when running multiple jobs in parallel with long names, the container names were conflicting.

I think this function is supposed to make sure container names don't get longer than what is allowed. But container names can be up to 128 characters I believe.

Signed-off-by: Aidan Jensen <aidan@artificial.com>


KnisterPeter:
As far as I can tell (without having a deeper look) the method trims every `part` to the len of 30 chars. With your change every part could be 128 chars which could easily be longer than the maximum 128 chars for a container name.

If you need to change the method I would suggest to create a full name (without being shortened) and then append a hash for the full name to it. This has could probably be shortened like git commit hashes are.

artificial-aidan:
Maybe I misunderstood it. I read it as truncating each part to `30/numParts`. This way no matter how many parts you had, the max length would be 30.

I agree with the hash for the full name, I'll give that a shot.

KnisterPeter:
I could be wrong as well 😝

artificial-aidan:
I think a hash is a better answer anyway. A SHA1 hash should never be longer than 40 characters. So I'll trim the name to 40 88 characters, and append a hash

KnisterPeter (approved):
:+1:

PR #1607, opened 2023-02-04
fix: add GITHUB_STEP_SUMMARY

### LINKED ISSUE #1533: Add support for `>> $GITHUB_STEP_SUMMARY`
### Act version

0.2.34

### Feature description

With the usage of `$GITHUB_STEP_SUMMARY` in workflows it's possible to generate markdown files as a summary of a job like this:
```bash
echo "{markdown content}" >> $GITHUB_STEP_SUMMARY
``` 
The usage is pretty similar to `$GITHUB_ENV` or `$GITHUB_OUTPUT`.
Ref.: https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#adding-a-job-summary

However, act fails on this point, because the `$GITHUB_STEP_SUMMARY` variable is not set:
```txt
/var/run/act/workflow/3: line 14: $GITHUB_STEP_SUMMARY: ambiguous redirect
```
I don't think that `act` should actually generate a markdown file as well, IMO it's enough to just ensure that the workflow doesn't fail if the syntax is used.

Thanks for the cool project :heart:

catthehacker:
Fixes #1533 

## Commit Message

{{title}} (#1607)


KnisterPeter (dismissed):

joanrodriguez:
Let's goooo let's merge this!

cplee (approved):

PR #1656, opened 2023-03-01
Improve XDG Spec supporting

Tony-Sol:
Switch from direct using of `XDG_CONFIG_HOME` env variable to `adrg/xdg` lib.

Now, it will be support xdg across windows/darwin/unix platforms and looks for config file like:
```
~/.actrc
"${XDG_CONFIG_HOME}/act/actrc"
"${XDG_CONFIG_HOME}/.actrc"
./.actrc
```

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -12,6 +12,7 @@ import (
 	"strings"
 
 	"github.com/AlecAivazis/survey/v2"
+	"github.com/adrg/xdg"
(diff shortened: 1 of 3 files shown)


KnisterPeter (approved):

cplee (approved):

mrnossiom:
It seems that even though act looks for config files in these dirs, it still writes the default config to `~/.actrc`

PR #1674, opened 2023-03-10
Support for docker steps in host environment

shubham149:
Network mode in host environment is set to container network earlier. But the container network is not present. This PR uses default as network mode in case host environment is enabled.

Also, setting working directory for container step to same as one specified in the input params.


shubham149:
@ChristopherHX I have reverted workdir changes from this pull request. 

Is there a way we can set the working directory for container steps same as input working directory for host env?

PR #1682, opened 2023-03-16
Draft: LXC support for self-hosted runners

earl-warren:
## Description

The  [LXC](https://linuxcontainers.org/) support for self-hosted runners is used to **run tests that do not fit the constraints imposed by the docker backend** such as having a systemd capable environment.

It creates a container from scratch on every run. If the tests accidentally damage essential services such as the ssh server, it will not have any impact on the host running the LXC container. If the same accident happens without the isolation provided by the LXC container, the host itself will be damaged. The **LXC support provides a robust isolation for each job in the workflows, which the self-hosted platform does not.**

## Implementation details

It is roughly the equivalent of doing the following:

* Creating a LXC container (with lxc-create)
* Running the commands in the container (with lxc-attach)
* Destroying the LXC container (with lxc-destroy)

This is inherently insecure, in the same way the self-hosted platform is. Hardening LXC containers is possible but it makes them no more useful than docker containers.

## FAQ

* **Why not run LXC inside a Docker container?** Because [it does not work](https://github.com/woodpecker-ci/woodpecker/pull/1565#issuecomment-1413922298). Although it could probably be hacked, that would not be more than that.
* **Why not run the LXC logic from the sef-hosted platform?** Because it would require significant work and LXC knowledge from each ACT user. Instead they only have to care for their own tests, not to setup and teardown well isolated LXC containers.
* **Why not run the LXC containers as unprivileged to improve security?** Because it does not work with [most templates](https://linuxcontainers.org/lxc/getting-started/#creating-unprivileged-containers-as-a-user). Given the restrictions it imposes, the user is better of using Docker. 
* **Why is apparmor disabled?** Because it cannot be configured [to allow essential operations such as mounting file systems](https://discuss.linuxcontainers.org/t/what-does-security-nesting-true/7156/2)

### DIFF
--- a/pkg/container/host_environment.go
+++ b/pkg/container/host_environment.go
@@ -25,16 +25,18 @@ import (
 )
 
 type HostEnvironment struct {
+	Name      string
 	Path      string
 	TmpDir    string
 	ToolCache string
 	Workdir   string
 	ActPath   string
+	Root      string
 	CleanUp   func()
 	StdOut    io.Writer
 }
 
-func (e *HostEnvironment) Create(capAdd []string, capDrop []string) common.Executor {
+func (e *HostEnvironment) Create(capAdd, capDrop []string) common.Executor {
 	return func(ctx context.Context) error {
 		return nil
 	}
@@ -60,7 +62,7 @@ func (e *HostEnvironment) Copy(destPath string, files ...*FileEntry) common.Exec
 	}
 }
 
-func (e *HostEnvironment) CopyDir(destPath string, srcPath string, useGitIgnore bool) common.Executor {
+func (e *HostEnvironment) CopyDir(destPath, srcPath string, useGitIgnore bool) common.Executor {
 	return func(ctx context.Context) error {
 		logger := common.Logger(ctx)
 		srcPrefix := filepath.Dir(srcPath)
@@ -254,7 +256,7 @@ func getEnvListFromMap(env map[string]string) []string {
 	return envList
 }
 
-func (e *HostEnvironment) exec(ctx context.Context, command []string, cmdline string, env map[string]string, user, workdir string) error {
+func (e *HostEnvironment) exec(ctx context.Context, commandparam []string, cmdline string, env map[string]string, user, workdir string) error {
 	envList := getEnvListFromMap(env)
 	var wd string
 	if workdir != "" {
@@ -266,6 +268,15 @@ func (e *HostEnvironment) exec(ctx context.Context, command []string, cmdline st
 	} else {
 		wd = e.Path
 	}
+	command := make([]string, len(commandparam))
+	copy(command, commandparam)
+	if user == "root" {
+		command = append([]string{"/usr/bin/sudo"}, command...)
+	} else {
+		common.Logger(ctx).Debugf("lxc-attach --name %v %v", e.Name, command)
+		command = append([]string{"/usr/bin/sudo", "--preserve-env", "--preserve-env=PATH", "/usr/bin/lxc-attach", "--keep-env", "--name", e.Name, "--"}, command...)
+	}
+
 	f, err := lookupPathHost(command[0], env, e.StdOut)
 	if err != nil {
 		return err
@@ -308,7 +319,7 @@ func (e *HostEnvironment) exec(ctx context.Context, command []string, cmdline st
 	}
 	err = cmd.Run()
 	if err != nil {
-		return err
+		return fmt.Errorf("RUN %w", err)
 	}
 	if tty != nil {
 		writer.AutoStop = true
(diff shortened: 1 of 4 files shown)


earl-warren:
This is an early draft go gauge interest and I'm ready to do the needful if there is. It is developed and used in the context of the https://forgejo.org project.

pkg/runner/run_context.go
```diff
@@ -139,6 +141,113 @@ func (rc *RunContext) GetBindsAndMounts() ([]string, map[string]string) {
...
+# https://github.com/nodesource/distributions#debinstall
+apt-get install -y curl git
+curl -fsSL https://deb.nodesource.com/setup_16.x | bash -
+apt-get install -y nodejs
+EOF
+chmod +x /var/lib/lxc/{{.Name}}/rootfs/tmpdir/node.sh
+
+lxc-attach --name {{.Name}} -- /tmpdir/node.sh
+
+`))
```
panekj, line 213: This should be separate file an embedded into go

pkg/runner/run_context.go
```diff
@@ -139,6 +141,113 @@ func (rc *RunContext) GetBindsAndMounts() ([]string, map[string]string) {
...
+
+var stopTemplate = template.Must(template.New("stop").Parse(`#!/bin/sh -x
+lxc-ls -1 --filter="^{{.Name}}" | while read container ; do
+   lxc-stop --kill --name="$container"
+   umount "/var/lib/lxc/$container/rootfs/{{ .Root }}"
+   umount "/var/lib/lxc/$container/rootfs/tmpdir"
+   lxc-destroy --force --name="$container"
+done
+`))
```
panekj, line 222: ditto

pkg/container/host_environment.go
```diff
@@ -266,6 +268,15 @@ func (e *HostEnvironment) exec(ctx context.Context, command []string, cmdline st
 	} else {
 		wd = e.Path
 	}
+	command := make([]string, len(commandparam))
+	copy(command, commandparam)
+	if user == "root" {
+		command = append([]string{"/usr/bin/sudo"}, command...)
```
panekj, line 274: hardcoded `sudo`

pkg/container/host_environment.go
```diff
@@ -266,6 +268,15 @@ func (e *HostEnvironment) exec(ctx context.Context, command []string, cmdline st
...
+	} else {
+		common.Logger(ctx).Debugf("lxc-attach --name %v %v", e.Name, command)
+		command = append([]string{"/usr/bin/sudo", "--preserve-env", "--preserve-env=PATH", "/usr/bin/lxc-attach", "--keep-env", "--name", e.Name, "--"}, command...)
```
panekj, line 277: ditto

panekj (changes requested):

panekj:
This should be separate backend from host

KnisterPeter:
@earl-warren This sounds very interesting. I haven't read through the code but I would like to see it integrated into act. I guess we need to discuss about the structure and interfaces a bit more. It's not always clearly separated currently.

@ChristopherHX might have some ideas and expectations here as well.

earl-warren:
Great to hear :tada: 

It could also be an optional feature of the self-hosted platform. If LXC is available and the option is set, wrap it into a LXC container for better isolation. Otherwise not. It would be entirely transparent to the workflows being run.

earl-warren:
Thanks for the feedback @ChristopherHX 

> I prefer that you create a new struct, which overrides (in golang it is probably a different wording, using interfaces it should work) some HostEnvironment functions. Then instantiate it via a custom identifier from the runcontext, maybe -lxc:<baseimage> instead of -self-hosted ( needs design consent, it took long for my PR to land ).

I'm happy to go in this direction. The current implementation is more of a proof of concept / hack than anything else and reworking it entirely is what I was expecting anyways.

If anyone think it should go differently please speak up. Otherwise I'll start working as soon as time permits, which could be as soon as in two weeks from now.

earl-warren:
I'm still motivated, it just takes a little longer than expected.

YOU54F:
I was watching this, as it looks really interesting.

I was using lxd for ephemeral gh runners

- https://github.com/stgraber/lxd-github-actions
- https://github.com/rgl/lxd-github-actions-runner

Leveraging `macpine` which is a macos tool in go that creates alpine vms with qemu (similar to lima)

https://beringresearch.github.io/macpine/lxd_macpine/

>Tbh, I need more disk space / do that on a raspberry pi to actually make that run like the macOS tart.

☝🏾 this! I ended up buying a 4tb nvme and external tb4 dock to satisfy chonki vm's and my desire to create many of them 😅

PR #1705, opened 2023-03-28
Avoid using `log.Fatal` in `pkg/*`

wolfogre:
Although act is a command line tool, it is also a bad idea to use `log.Fatal` in packages.

To be clear, only packages which are used as libraries. It's fine to use fatal in `cmd`.


pkg/common/outbound_ip.go
```diff
@@ -2,16 +2,15 @@ package common
...
 )
 
 // https://stackoverflow.com/a/37382208
 // Get preferred outbound ip of this machine
 func GetOutboundIP() net.IP {
 	conn, err := net.Dial("udp", "8.8.8.8:80")
 	if err != nil {
-		log.Fatal(err)
+		// return nil instead of fatal
+		return nil
```
wolfogre, line 13: Just like what `net.ParseIP` does.

pkg/model/workflow.go
```diff
@@ -670,3 +664,17 @@ func (w *Workflow) GetJobIDs() []string {
 	}
 	return ids
 }
+
+var OnDecodeNodeError = func(node yaml.Node, out interface{}, err error) {
```
wolfogre, line 668: So it's possible to ignore errors when decoding nodes by setting `OnDecodeNodeError` to nil.

KnisterPeter:
Hi @wolfogre, can you explain a bit on the reasoning of your changes. Are you using some packages outside of act? Did you stumble upon an error?

wolfogre:
>  Are you using some packages outside of act?

TBH, yes. I am a maintainer of [Gitea](https://github.com/go-gitea/gitea). And Gitea has a CICD runner named [act_runner](https://gitea.com/gitea/act_runner), it based on a soft [fork](https://gitea.com/gitea/act) of act.

More context:
- https://blog.gitea.io/2023/03/gitea-1.19.0-is-released/#-gitea-actions-21937httpsgithubcomgo-giteagiteaissues21937
- https://github.com/go-gitea/gitea/issues/13539

KnisterPeter:
@wolfogre Ah I see. The missing piece for me was that you are a maintainer of gitea. That makes sense.

pkg/common/outbound_ip.go
wolfogre: In fact, I noticed that it cannot work without internet access, so #1707

KnisterPeter (approved):

cplee (approved):

PR #1706, opened 2023-03-28
Use os.UserHomeDir

wolfogre:
`github.com/mitchellh/go-homedir` is made for the old version. After go 1.12, there's `os.UserHomeDir`.

### DIFF
--- a/cmd/notices.go
+++ b/cmd/notices.go
@@ -11,7 +11,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/mitchellh/go-homedir"
 	log "github.com/sirupsen/logrus"
 )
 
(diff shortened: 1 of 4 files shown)


wolfogre:
BTW, I don't know why act uses `github.com/ZauberNerd/go-git`, it still uses go-homedir while the upstream already dropped it:

https://github.com/go-git/go-git/pull/535

ZauberNerd:
> BTW, I don't know why act uses `github.com/ZauberNerd/go-git`, it still uses go-homedir while the upstream already dropped it:
> 
> [go-git/go-git#535](https://github.com/go-git/go-git/pull/535)

We're using the fork because of this issue: https://github.com/go-git/go-git/pull/485

PR #1732, opened 2023-04-16
fix: environment handling windows (host mode)

ChristopherHX:
Why?

Currently we merge env variables with case sensitive key, this is problematic on windows where PATH and Path are the same key.

Other Changes considered:
Create a custom env type
```go
type Map[K comparable, V any] interface {
	Set(key K, val V)
	Get(key K) V
	TryGet(key K) (V, bool)
	AsNative() map[K]V
}

type CaseInsensitiveMap[V any] struct {
	lookUp map[string]string
	Native map[string]V
}

func (m *CaseInsensitiveMap[V]) getKey(key string) string {
	return strings.Map(func(r rune) rune {
		for {
			next := unicode.SimpleFold(r)
			if next <= r {
				return r
			}
			r = next
		}
	}, key)
}

func (m *CaseInsensitiveMap[V]) Set(key string, val V) {
	foldkey := m.getKey(key)
	if m.Native == nil {
		m.Native = map[string]V{}
	}
	if fk, ok := m.lookUp[foldkey]; ok {
		m.Native[fk] = val
	} else {
		if m.lookUp == nil {
			m.lookUp = map[string]string{}
		}
		m.lookUp[foldkey] = key
		m.Native[key] = val
	}
}

func (m *CaseInsensitiveMap[V]) Get(key string) V {
	v, _ := m.TryGet(key)
	return v
}

func (m *CaseInsensitiveMap[V]) TryGet(key string) (V, bool) {
	foldkey := m.getKey(key)
	if m.Native == nil {
		m.Native = map[string]V{}
	}
	if fk, ok := m.lookUp[foldkey]; ok {
		return m.Native[fk], true
	}
	return m.Native[key], false
}

func (m *CaseInsensitiveMap[V]) AsNative() map[string]V {
	return m.Native
}

type RegularMap[K comparable, V any] struct {
	Native map[K]V
}

func (m *RegularMap[K, V]) Set(key K, val V) {
	if m.Native == nil {
		m.Native = map[K]V{}
	}
	m.Native[key] = val
}

func (m *RegularMap[K, V]) Get(key K) V {
	return m.Native[key]
}

func (m *RegularMap[K, V]) TryGet(key K) (V, bool) {
	v, ok := m.Native[key]
	return v, ok
}

func (m *RegularMap[K, V]) AsNative() map[K]V {
	return m.Native
}
```

However this would require a lot of changes across the codebase.

Next steps:
- make the `env` expressions context case sensitive again for non windows, because the env keys can be duplucated in the map


pkg/runner/step.go
```diff
@@ -258,10 +259,47 @@ func isContinueOnError(ctx context.Context, expr string, step step, stage stepSt
...
+	}
+}
+
+func mergeIntoMapCaseSensitive(target *map[string]string, maps ...map[string]string) {
 	for _, m := range maps {
 		for k, v := range m {
 			(*target)[k] = v
 		}
 	}
 }
```
wolfogre, line 276: I think this should work too:
```suggestion
func mergeIntoMapCaseSensitive(target map[string]string, maps ...map[string]string) {
	for _, m := range maps {
		for k, v := range m {
			target[k] = v
		}
	}
}
```

pkg/runner/step.go
```diff
@@ -258,10 +259,47 @@ func isContinueOnError(ctx context.Context, expr string, step step, stage stepSt
...
+}
+
+func mergeIntoMapCaseInsensitive(target *map[string]string, maps ...map[string]string) {
+	lookUp := map[string]string{}
+	// Need to initialize the lookUp table
+	mergeSingleMapIntoMapCaseInsensitive(&lookUp, target, *target)
+	for _, m := range maps {
+		mergeSingleMapIntoMapCaseInsensitive(&lookUp, target, m)
+	}
+}
```
wolfogre, line 305: How about?
```suggestion
func mergeIntoMapCaseInsensitive(target map[string]string, maps ...map[string]string) {
	foldKeys := make(map[string]string, len(target))
	for k := range target {
		foldKeys[strings.ToLower(k)] = k
	}
	toKey := func(s string) string {
		foldKey := strings.ToLower(s)
		if k, ok := foldKeys[foldKey]; ok {
			return k
		}
		foldKeys[strings.ToLower(foldKey)] = s
		return s
	}
	for _, m := range maps {
		for k, v := range m {
			target[toKey(k)] = v
		}
	}
}
```

wolfogre (approved):

KnisterPeter (approved):

PR #1733, opened 2023-04-16
feat: support yaml env/secrets/inputs file

ChristopherHX:
Why?

I don't like the godotfile syntax.

Proposal:
Let us support yaml if the user specify a `.yml` or `.yaml` file.

This has been added to my other github actions local runner tool, since a while. Act should support it too.

Multiline secrets with ease
```yaml
MYSECRET: |
  l1
  l2
  l3

```
without a newline after the last line of the secret
```yaml
MYSECRET: |-
  l1
  l2
  l3

```

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -55,7 +55,7 @@ jobs:
         with:
           upload-logs-name: logs-linux
       - name: Upload Codecov report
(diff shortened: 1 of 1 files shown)


KnisterPeter (approved):
Fine since this is not a breaking change. We are currently using the secrets file format in our production setup.

PR #1739, opened 2023-04-18
fix: ensure networkmode "host" unless explicitly specified

ZauberNerd:
act defaults network mode to "host", but when `--container-options` are passed on the CLI, it uses the docker CLI options parser, which fills empty values with defaults, in which case network mode is set to "default".
Unless the user explicitly sets `--container-options="--network=xxx"`, we should always default to "host", to keep act's behaviour.


KnisterPeter (approved):

ZauberNerd:
> @ZauberNerd this pull request has failed checks hammer_and_wrench

Downloading node.js from nodejs.org failed with an HTTP error 500. Retriggering the test-linux checks fixed it.

PR #1740, opened 2023-04-18
allow command line container options to be appended to workflow specified options

### LINKED ISSUE #1696: Add customizable "network" option.
### Act version

v0.2.43

### Feature description

My workflow has multiple jobs which run in parallel. In GitHub, each runner/job is on its own network. In act, however, the jobs share the same "host" network as can be seen here: https://github.com/nektos/act/blob/master/pkg/runner/run_context.go#L263

The _problem_ is that there are many networking ports being used on the host machine, leading to intermittent port conflicts causing failures.

Passing in a new network option to a container within the workflow doesn't seem to work.

eljohnson92:
This PR allows for command line options passed through `--container-options` to override any specified in the workflow file. It fixes #1696 which is needed for situations where you need to override a container option just in act and not in the GHA themselves. 

In the GIthubs section on [Container Options](https://docs.github.com/en/actions/using-jobs/running-jobs-in-a-container#setting-container-resource-options) they call out they do not support the `--network` option for containers because they run containers explicitly in the bridge network and you cannot specify the `--network` flag twice in an invocation. With this change, ACT users can run both GHA and ACT runs from the same workflow file with the same network type.

### DIFF
--- a/pkg/runner/run_context.go
+++ b/pkg/runner/run_context.go
@@ -521,11 +521,13 @@ func (rc *RunContext) platformImage(ctx context.Context) string {
 func (rc *RunContext) options(_ context.Context) string {
 	job := rc.Run.Job()
 	c := job.Container()
-	if c == nil {
-		return rc.Config.ContainerOptions
+	containerOptions := ""
+	if c != nil {
+		containerOptions = c.Options
 	}
+	containerOptions += " " + rc.Config.ContainerOptions
 
-	return c.Options
+	return containerOptions
 }
 
 func (rc *RunContext) isEnabled(ctx context.Context) (bool, error) {


eljohnson92:
@KnisterPeter @ChristopherHX Thank you for your responses, I did not consider the conditional options and can try this out. the issue for us right now is ACT is acting differently than our GHA Runners. Our GHA Runners start containers on a specific bridge network and not on the host network so we are looking for a way to have act mimic our GHA environment.

eljohnson92:
@ChristopherHX I have updated my code to now append to the end of any specified workflow file options. let me know if this is a more appropriate solution to the problem.

pkg/runner/run_context.go
```diff
@@ -480,11 +480,13 @@ func (rc *RunContext) platformImage(ctx context.Context) string {
 func (rc *RunContext) options(ctx context.Context) string {
 	job := rc.Run.Job()
 	c := job.Container()
-	if c == nil {
-		return rc.Config.ContainerOptions
+	containerOptions := ""
+	if c != nil {
+		containerOptions = c.Options
 	}
+	containerOptions += " " + rc.Config.ContainerOptions
```
wolfogre, line 487: I don't really think it's a good idea. What do we expect Docker to do when the result is `"--network bridge" + " " + "--network host" `?

KnisterPeter:
> The official GitHub workflow vscode plugin, doesn't like `env.ACT`

According to the docs, `env` is allowed in container settings: 
https://docs.github.com/en/actions/learn-github-actions/contexts#context-availability

I think the vscode plugin is not a good reference for truth (or the docs).

KnisterPeter:
I'm with @wolfogre here. It would be unclear what to do in that case.
Also as pointed out above, act currently does not try to allow different behavior (related to what is stated in the workflow file).

*We, with our integration pattern, are relying on network mode `host` currently, but that by itself should not block this PR if it's expected behavior.*

If we really want this feature, I would suggest to make this even more explicit, by adding a `--network-mode-override` flat or maybe `--container-options-override`. But actually I don't like it that much.

eljohnson92:
@KnisterPeter the issue we are running into is containers in our gha runners are being started in the bridge network so the fact that act is starting it's container on the host network is causing unexpected behavior for us which we are trying to reconcile. It seems we don't have a solution here that everyone can agree on so let me first test your suggestion to see if that will work for us. The main issue is gha complains if we add a `--network` flag to the workflow file because it identifies it will conflict unexpectedly so we cannot put it in our workflow file but we require a way to set the network if we are going to make act work the same as GHA

eljohnson92:
@KnisterPeter I've tried this in GHA and it does not seem to be working. I used the following:
```
    container:
      image: some-image
      options: '${{ env.ACT == "true" && --network bridge }}'
```
this does work in act, but when I run it in GHA I'm getting
```
Unrecognized named-value: 'env'. Located at position 1 within expression: env.ACT == 'true' && --network bridge
```

additionally when trying to just set the `--network bridge` flag in the workflow file I get this error in GHA
```
docker: conflicting options: cannot attach both user-defined and non-user-defined network-modes.
```

KnisterPeter:
@eljohnson92 Tested as well and it will work like this:

```yaml
name: test

on: push

jobs:
  check:
    runs-on: ubuntu-latest
    steps:
      - id: check-env
        run: echo "custom-network=${{ env.ACT }}" >> $GITHUB_OUTPUT
    outputs:
      custom-network: ${{ steps.check-env.outputs.custom-network }}
  test:
    needs: check
    runs-on: ubuntu-latest
    container:
      image: ubuntu:23.04
      # use a custom network for act and an option without side-effect on GHA
      options: "${{ needs.check.outputs.custom-network == 'true' && '--network host' || '--label test' }}"
    steps:
      - run: echo "hello world"
```

I know it's not pretty, but it does work.

eljohnson92:
@KnisterPeter Testing this out now and running into quite a few issues between how ACT and GHA handle this but I think there is a path forward possibly. it would be really nice if we had a way to just force ACT to work identically to GHA rather than having to do this more complex checking to handle differences between the ACT environment and GHA environment

eljohnson92:
I've found a slightly simpler and more fragile solution
```
name: test

on: push

jobs:
  test:
    runs-on: ubuntu-latest
    container:
      image: ubuntu:23.04
      # use a bridge network for act and specify a label otherwise because something must be specified if options exists
      options: "${{ github.server_url == '' && ' --network bridge' || ' --label GHA' }}"
    steps:
      - run: echo "hello world"
```
One important note with both your solution above and this one is there has to be an arbitrary label added from my testing because if `options:` is used something must be specified

Issue #1756, opened 2023-04-24
Failed to bind docker socket


KnisterPeter:
### Bug report info

```plain text
act version:            0.2.44-27-g2ea7891
GOOS:                   linux
GOARCH:                 amd64
NumCPU:                 12
Docker host:            DOCKER_HOST environment variable is not set
Sockets found:
        /var/run/docker.sock
Config files:
        /home/markusw/.actrc:
                -P ubuntu-latest=node:16-buster-slim
                -P ubuntu-22.04=node:16-bullseye-slim
                -P ubuntu-20.04=node:16-buster-slim
                -P ubuntu-18.04=node:16-buster-slim
Build info:
        Go version:            go1.18.10
        Module path:           command-line-arguments
        Main version:
        Main path:
        Main checksum:
        Build settings:
                -compiler:            gc
                -ldflags:             -X main.version=0.2.44-27-g2ea7891
                CGO_ENABLED:          1
                CGO_CFLAGS:
                CGO_CPPFLAGS:
                CGO_CXXFLAGS:
                CGO_LDFLAGS:
                GOARCH:               amd64
                GOOS:                 linux
                GOAMD64:              v1
Docker Engine:
        Engine version:        23.0.3
        Engine runtime:        runc
        Cgroup version:        2
        Cgroup driver:         cgroupfs
        Storage driver:        overlay2
        Registry URI:          https://index.docker.io/v1/
        OS:                    Debian GNU/Linux 11 (bullseye)
        OS type:               linux
        OS version:            11
        OS arch:               x86_64
        OS kernel:             5.15.90.1-microsoft-standard-WSL2
        OS CPU:                12
        OS memory:             15891 MB
        Security options:
                name=seccomp,profile=builtin
                name=cgroupns


INFO    ️📣 A newer version of 'act' is available - consider ugrading to 0.2.44.
```

### Command used with act

```sh
./dist/local/act --verbose --rm --rebuild --pull=false -C ./pkg/runner/testdata/ -W ./basic/push.yml
```

### Describe issue

Since https://github.com/nektos/act/pull/1745 the binding of the docker socket fails (at least in some cases).
We did saw this on a linux host as well as on a host running in WSL2 on windows (with docker installed inside WSL).

### Workflow content

```yml
The basic push workflow from our testdata (see cli above).
```

### Relevant log output

```sh
[...]

Issue #1744, opened 2023-04-19
Error: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running? on Ubuntu while docker commands are working


kintel:
### Bug report info

```plain text
gh act --bug-report
act version:            0.2.44
GOOS:                   linux
GOARCH:                 amd64
NumCPU:                 20
Docker host:            DOCKER_HOST environment variable is unset/empty.
Sockets found:
        /var/run/docker.sock
Config files:           
        /home/kintel/snap/gh/502/.actrc:
                -P ubuntu-latest=catthehacker/ubuntu:act-latest
                -P ubuntu-22.04=catthehacker/ubuntu:act-22.04
                -P ubuntu-20.04=catthehacker/ubuntu:act-20.04
                -P ubuntu-18.04=catthehacker/ubuntu:act-18.04
Build info:
        Go version:            go1.18.10
        Module path:           github.com/nektos/act
        Main version:          (devel)
        Main path:             github.com/nektos/act
        Main checksum:         
        Build settings:
                -compiler:            gc
                CGO_ENABLED:          0
                GOARCH:               amd64
                GOOS:                 linux
                GOAMD64:              v1
                vcs:                  git
                vcs.revision:         65088b8f28d44da128b95ec194a04277b8856be9
                vcs.time:             2023-01-20T18:39:07Z
                vcs.modified:         true

Error: permission denied while trying to connect to the Docker daemon socket at unix:///var/run/docker.sock: Get "http://%2Fvar%2Frun%2Fdocker.sock/v1.24/info": dial unix /var/run/docker.sock: connect: permission denied
```

### Command used with act

```sh
gh act
```

### Describe issue

Error: permission denied while trying to connect to the Docker daemon socket at unix:///var/run/docker.sock: Get "http://%2Fvar%2Frun%2Fdocker.sock/v1.24/info": dial unix /var/run/docker.sock: connect: permission denied

### Workflow content

```yml
Any workflow
```

### Relevant log output

```sh
DEBU[0000] Loading environment from /home/kintel/code/OpenSCAD/openscad-testing/.env 
DEBU[0000] Loading action inputs from /home/kintel/code/OpenSCAD/openscad-testing/.input 
DEBU[0000] Loading secrets from /home/kintel/code/OpenSCAD/openscad-testing/.secrets 
DEBU[0000] Evaluated matrix inclusions: map[]           
[...]

PR #1760, opened 2023-04-24
Don't bind host docker daemon socket

vicamo:
Binding host docker daemon socket into executor containers can be a terrible security hole. Don't do this unless explicitly specified through customized mounts.

To exploit this issue, run following test job:
```yml
on:
  pull_request:
    branches:
      - 'main'

jobs:
  test:
    runs-on: ubuntu-latest
    container: docker:git
    steps:
      - name: Bind host rootfs
        run: |
          docker run --detach --name i_am_almighty -v /:/host-rootfs ubuntu:latest /bin/tail -f /dev/null
          docker exec i_am_almighty id
          docker exec i_am_almighty ls -la /host-rootfs
```
See also #1744 , #1756, #1757.

### DIFF
--- a/pkg/runner/run_context.go
+++ b/pkg/runner/run_context.go
@@ -95,9 +95,7 @@ func (rc *RunContext) GetBindsAndMounts() ([]string, map[string]string) {
 		rc.Config.ContainerDaemonSocket = "/var/run/docker.sock"
 	}
 
-	binds := []string{
-		fmt.Sprintf("%s:%s", rc.Config.ContainerDaemonSocket, "/var/run/docker.sock"),
-	}
+	binds := []string{}
 
 	ext := container.LinuxContainerEnvironmentExtensions{}
 


vicamo:
> This is by design and required. `act` is designed to run with docker and it is required to operate (e.g. to build containers inside workflows). `act` should usually be run by users on their local machine to test workflows.

Docker builds can still be done with service containers just like all other native GitHub workflows do.

> If it's unsecure to run act on your machine or run a workflow locally, the security issue is already there and act should better not be used. Therefore I would opt against this change.

This issue was originally found from a [fork](https://gitea.com/gitea/act.git) meant for Gitea's Actions support. Not actually a problem of a local running act as it was meant to be, but maybe it will be a very critical problem when it's going to be adopted as a viable alternative of other online CI solutions.

KnisterPeter:
> Docker builds can still be done with service containers just like all other native GitHub workflows do.

Since native GitHub Actions are running on a VM on Azure where docker is installed, this is equally insecure.
There is no need to use service containers, just use plain docker.

vicamo:
> We bind the docker socket, because GitHub Actions https://github.com/actions/runner does it.

I think you're right, especially with the self-hosted container feature. I'm closing this PR.

PR #1764, opened 2023-04-24
Add ubuntu 22.04-based images (x86 and arm)

fwilhe2:
Provides a 'base' image with about 1gig (compressed) and variants for jvm, rust and gcc. Use github's provisioning scripts.


wolfogre:
To be honest, I don't really think they are necessary, or why not `latest-golang`, `latest-php`, `latest-dotnet`, `latest-python`?

fwilhe2:
> To be honest, I don't really think they are necessary, or why not `latest-golang`, `latest-php`, `latest-dotnet`, `latest-python`?

Hi @wolfogre, sure they are not 'necessary' in a strict sense, but I think they might still be useful. The base variant is relatively small while still being quite similar to the hosted runner environment, allowing to run many workflows. The additional variants are just those that felt most common to me, sparing the extra download when using the jdk or rust.

I think the images might be useful that's why I thought it would be nice to add them here.

PR #1784, opened 2023-05-01
fix: remove hardcoded reference to github.com when using reusable remote workflows and remote actions

### LINKED ISSUE #1588: Support for ssh remote workflows (and actions) or using git config
### Act version

act version 0.2.40-14-g932863b

### Feature description

I'm using a build from `master` that contains the [new feature for remote workflows](https://github.com/nektos/act/pull/1525). The remote workflows are hard coded to use HTTPS, and I need SSH support.

https://github.com/nektos/act/blob/82a8c1e80dfcd73842bc0359ca8447303b6fcc08/pkg/runner/reusable_workflow.go#L109

Normally I would use the `insteadOf` config to fix this but it didn't seem to work.

```shell
git config --global url."ssh://git@github.com/".insteadOf "https://github.com/"
```

```shell
% act pull_request
[build-test-release.yaml/MyWorkflow]   ☁  git clone 'https://github.com/myorg/reusable-workflows' # ref=my_branch
[build-test-release.yaml/MyWorkflow] Unable to clone https://github.com/myorg/reusable-workflows refs/heads/my_branch: authentication required
Error: authentication required
```

However, if I manually clone it, the `insteadOf` config works fine. I noticed the clone is not run in a container, so I'm not sure if it is otherwise ignoring the global git config.

```shell
% git clone 'https://github.com/myorg/reusable-workflows'
Cloning into 'reusable-workflows'...
remote: Enumerating objects: 1183, done.
remote: Counting objects: 100% (360/360), done.
...
```

shubhbapna:
fixes #1588 

By removing reference to the hardcoded `github.com` and instead using `serverURL` from github context, we can start controlling where we clone these actions/workflows from. So for example if we want `act` to use ssh then we can simply pass the following option:
```
--env GITHUB_SERVER_URL=ssh://git@github.com
```

Moreover, I noticed that for reusable workflows we are cloning the repositories in a directory named - `{owner}-{repo}-.github-workflows-{filename}@{ref}`.  However if we change this directory name to be `{owner}-{repo}@{ref}`, we will be able to reduce the number of clones `act` will make when using multiple reusable workflows from the same repository.


KnisterPeter (approved):
Looks good. Thanks for your contribution.

cplee (approved):

PR #1834, opened 2023-06-01
 Fix bug in processing jobs on platforms without Docker

psa:
The container.GetHostInfo() struct is returned empty on hosts without docker and this results in passing zero parallelism to the parallel executor.

Use Golang's inbuilt runtime.NumCPU() instead. This should be safe on Docker too.

I've also added a fair bit of debug logging to assist with debugging the struct passed to the runner. This is leftover from my original line of inquiry, but it seemed like useful code so I kept it in the change.

### DIFF
--- a/pkg/common/executor.go
+++ b/pkg/common/executor.go
@@ -3,6 +3,8 @@ package common
 import (
 	"context"
 	"fmt"
+
+	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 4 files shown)


wolfogre (approved):
LGTM

cplee (approved):

PR #1913, opened 2023-07-16
feat: Add new Action Cache

ChristopherHX:
This is a sample implementation of ActionCache.

See also #1912 and it's description for further links.


pkg/runner/action_cache.go
```diff
@@ -0,0 +1,177 @@
...
+	}
+	commit, err := gogitrepo.CommitObject(plumbing.NewHash(sha))
+	if err != nil {
+		return nil, err
+	}
+	files, err := commit.Files()
+	if err != nil {
+		return nil, err
+	}
+	rpipe, wpipe := io.Pipe()
```
cplee, line 120: This seems inefficient to use in-memory pipe for a tar file. How will this function be used? How about either passing in the writer to this function or passing in a callback function that will be called for each matching file?

pkg/runner/action_cache.go
```diff
@@ -0,0 +1,177 @@
...
+	"github.com/go-git/go-git/v5/plumbing/transport"
+	"github.com/go-git/go-git/v5/plumbing/transport/http"
+)
+
+type ActionCache interface {
+	Fetch(ctx context.Context, cacheDir, url, ref, token string) (string, error)
+	GetTarArchive(ctx context.Context, cacheDir, sha, fpath string) (io.ReadCloser, error)
+}
+
+type GoGitActionCache struct {
```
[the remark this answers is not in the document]
cplee: I assumed you'd need a new instance per action since it is the git repo of the action. Is that incorrect?

pkg/runner/action_cache.go
cplee: Agreed. Resolving.

cplee (approved):

Issue #1821, opened 2023-05-25
Hide workflow and job names from output using flag


lumunro:
### Act version

0.2.45

### Feature description

When running act it displays the workflow name and job name down the left hand side in the output. While this is great default behavior sometimes it would be good to not display this information (especially if you have long workflow and job names).

Would it be possible to hide this information from the output if you provide act with a flag?

Example: Hide the area in the red box

![view](https://github.com/nektos/act/assets/88164695/7037225a-93bb-4335-9eaa-46f670aa3fbf)

Thanks for your work on this awesome project!

PR #1920, opened 2023-07-19
Add new CLI flag to log just the job id and not the entire job name

ImTomEddy:
Log messages can already be fairly long which is exacerbated by the prefix, however the prefix is useful when running multiple jobs within a single act pipeline. The compromise in this PR is to allow for a shorter prefix using just the job ID, which is user configurable.

The flag chosen is intentionally verbose.

```bash
--log-prefix-job-id
```

Sample output without the flag
```bash
[call-workflow/Run In Container/run-in-container] [DEBUG] evaluating expression 'success()'
[call-workflow/Run In Container/run-in-container] [DEBUG] expression 'success()' evaluated to 'true'
[call-workflow/Run In Container/run-in-container] [DEBUG] Loading revision from git directory
```

Sample output with the flag
```bash
[run-in-container] [DEBUG] evaluating expression 'success()'
[run-in-container] [DEBUG] expression 'success()' evaluated to 'true'
[run-in-container] [DEBUG] Loading revision from git directory
```

Closes https://github.com/nektos/act/issues/1821

### DIFF
--- a/.github/actions/run-tests/action.yml
+++ b/.github/actions/run-tests/action.yml
@@ -54,7 +54,7 @@ runs:
               }
             }
           };
-          var args = ['test', '-v', '-cover', '-coverprofile=coverage.txt', '-covermode=atomic', '-timeout', '15m'];
+          var args = ['test', '-v', '-cover', '-coverprofile=coverage.txt', '-covermode=atomic', '-timeout', '20m'];
           var filter = process.env.FILTER;
           if(filter) {
             args.push('-run');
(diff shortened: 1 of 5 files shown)


ImTomEddy:
@nektos/act-maintainers I've had a look through the logs for `checks / test-linux (pull_request)` and can't see any reason for the test failure. Could someone please clarify what is wrong?

Many thanks!

ImTomEddy:
@ChristopherHX nice catch, thanks! I was looking for a failure not a panic - clearly I was being a bit silly.

The change will probably make _some_ performance impact but I wouldn't expect it to be significantly less performant than previously.

For now I've upped the timeout to 20m - I think it maybe could just be luck of the draw on the runner CPU cycle availability when running all the tests.

cplee (approved):

Issue #173, opened 2020-03-28
Services not working


fgimenez:
Hello! I have this workflow:
```name: test-and-build
on: [push]

jobs:
  tests:
    runs-on: ubuntu-latest
    container: node:12.16.1-stretch
    services:
      mongodb:
        image: "mongo:4.0.17"
    steps:
      - uses: actions/checkout@v2
      - run: yarn
      - run: yarn test
```
The `yarn test` command expects to find a mongodb service listening on 27017, but running act the command fails with `MongooseServerSelectionError: getaddrinfo ENOTFOUND mongodb`. The same workflow succeeds on GitHub.

I've also tried `-P ubuntu-latest=nektos/act-environments-ubuntu:18.04`, any help is very appreciated :)

thatdoogieguy:
I was having a similar problem, especially when i was trying to combine rails caching with postgresql and redis. In the end i created a seperate workflow file called ci-local.yml based off of the production Github Actions file i had created and manually defined the services by calling apt-get then service %service_name% start. 

Massive PITA but for localised testing this was the only way i could thoroughly test the functional parts of GitHub Actions where multiple parts are as yet unsupported. 

Not a fix, just a hacky workaround to get myself out of trouble!

ibrt:
Same problem here with postgres... It looks like the service container is not being run so DNS lookup of its name fails. Is this supposed to work?

Zeka13:
shiiiiiiiiiiiiit

jadhavmanoj:
Same here. Services not working. Any workaround for the same?

Zeka13:
@jadhavmanoj workaround is to use docker-compose

chenxilinsidney:
I have the same problem while using services of mysql and redis in action

creepinson:
Any updates? It's been another month

Lzok:
Hello, any update on this?

My code:
```yaml
jobs:
  test:
    runs-on: ubuntu-latest

    services:
      mongodb:
        image: mongo:4.0
        ports:
          - 27017/tcp
# ...etc
```

In the env section of my job I have: `MONGO_URI_TESTS: localhost:${{ job.services.mongodb.ports[27017] }}/dbtest`

The error: `ERRO[0082] Unable to interpolate string 'localhost:${{ job.services.mongodb.ports[27017] }}/dbtest' - [TypeError: Cannot access member 'ports' of undefined]`

Thank you!

Systho:
[...]

PR #1949, opened 2023-08-07
Add support for service containers

GuessWhoSamFoo:
Closes https://github.com/nektos/act/issues/173

This PR cherry picks:

- https://gitea.com/gitea/act/pulls/42
- https://gitea.com/gitea/act/pulls/45
- https://gitea.com/gitea/act/pulls/47
- https://gitea.com/gitea/act/pulls/51
- https://gitea.com/gitea/act/pulls/56
- https://gitea.com/gitea/act/pulls/60

https://gitea.com/gitea/act/pulls/50 is not moved over as it does not exist as a Github Action feature.

Gitea specific features such as `createSimpleContainerName` and the `AutoRemove` flag was removed during this process. In addition, `ContainerMaxLifetime` is also removed as since running `/bin/sleep 0` by default can be confusing and doesn't seem to have a use case outside of usage as an sdk.


pkg/runner/runner.go
```diff
@@ -20,44 +20,46 @@ type Runner interface {
...
+	ArtifactServerPath                 string                       // the path where the artifact server stores uploads
+	ArtifactServerAddr                 string                       // the address the artifact server binds to
+	ArtifactServerPort                 string                       // the port the artifact server binds to
+	NoSkipCheckout                     bool                         // do not skip actions/checkout
+	RemoteName                         string                       // remote name in local git repo config
+	ReplaceGheActionWithGithubCom      []string                     // Use actions from GitHub Enterprise instance to GitHub
+	ReplaceGheActionTokenWithGithubCom string                       // Token of private action repo on GitHub.
+	Matrix                             map[string]map[string]bool   // Matrix config to run
+	ContainerNetworkMode               docker_container.NetworkMode // the network mode of job containers (the value of --network)
+	ValidVolumes                       []string                     // only volumes (and bind mounts) in this slice can be mounted on the job container or service containers
```
wolfogre, line 62: I'm not sure if act needs `ValidVolumes`, since it always allows mounting.

GuessWhoSamFoo: Agreed, this config isn't being used anywhere and should be removed

GuessWhoSamFoo:
Not sure what is up with the artifact upload/download test. Maybe there's a side effect somewhere, gonna have to investigate later
```
[Test that artifact uploads and downloads succeed/test-artifacts]   | Create Artifact Container - Attempt 5 of 5 failed with error: connect ECONNREFUSED 127.0.0.1:12345
[Test that artifact uploads and downloads succeed/test-artifacts]   ❗  ::error::Create Artifact Container failed: connect ECONNREFUSED 127.0.0.1:12345
```

wolfogre (approved):
LGTM, and I believe more practical tests are needed.

Some notes for other reviewers:

- Services always run in containers, so docker will be required to run services (even run jobs in host mode).
- It will create a new docker network for service containers, and let the job container join the network. And delete the network when it has done.
- A new docker network make it possible to let services bind any ports without conflict, however, it also make things more complex. (That's why the team of Gitea are cautious to port it upstream)
- IIRC, act will keep the job container to reuse (while gitea/act never reuse), I'm not sure if the network could be used and deleted correctly.
- Docker can only provide a limited number of networks, so the network must be deleted timely.

GuessWhoSamFoo:
@ChristopherHX Thanks for the review - I'll take a closer look if something got missed in the backport and likely want to explicitly test https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idservicesservice_idvolumes as well

pkg/runner/job_executor.go
```diff
@@ -101,7 +102,26 @@ func newJobExecutor(info jobInfo, sf stepFactory, rc *RunContext) common.Executo
...
+			}
+			if rc.Config.ContainerNetworkMode == "" {
+				// if the value of `ContainerNetworkMode` is empty string,
+				// it means that the network to which containers are connecting is created by `act_runner`,
+				// so, we should remove the network at last.
+				logger.Infof("Cleaning up network for job %s, and network name is: %s", rc.JobName, rc.networkName())
+				if err := rc.removeNetwork(rc.networkName())(ctx); err != nil {
+					logger.Errorf("Error while cleaning network: %v", err)
+				}
+			}
```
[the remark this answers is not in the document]
GuessWhoSamFoo: Ported over https://gitea.com/gitea/act/commit/545802b97bdab805c9f8442b32c5b843fc827676 to ensure it is not running on host environment

pkg/runner/run_context.go
```diff
@@ -87,6 +87,12 @@ func (rc *RunContext) jobContainerName() string {
 	return createContainerName("act", rc.String())
 }
 
+// networkName return the name of the network which will be created by `act` automatically for job,
+// only create network if `rc.Config.ContainerNetworkMode` is empty string.
+func (rc *RunContext) networkName() string {
+	return fmt.Sprintf("%s-network", rc.jobContainerName())
```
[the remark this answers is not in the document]
GuessWhoSamFoo: Added job ID to help uniqueness

pkg/runner/runner.go
```diff
@@ -20,44 +20,45 @@ type Runner interface {
...
+	AutoRemove                         bool                         // controls if the container is automatically removed upon workflow completion
+	ArtifactServerPath                 string                       // the path where the artifact server stores uploads
+	ArtifactServerAddr                 string                       // the address the artifact server binds to
+	ArtifactServerPort                 string                       // the port the artifact server binds to
+	NoSkipCheckout                     bool                         // do not skip actions/checkout
+	RemoteName                         string                       // remote name in local git repo config
+	ReplaceGheActionWithGithubCom      []string                     // Use actions from GitHub Enterprise instance to GitHub
+	ReplaceGheActionTokenWithGithubCom string                       // Token of private action repo on GitHub.
+	Matrix                             map[string]map[string]bool   // Matrix config to run
+	ContainerNetworkMode               docker_container.NetworkMode // the network mode of job containers (the value of --network)
```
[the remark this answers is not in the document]
GuessWhoSamFoo: Removed the assigned test field. Instead the behavior is defaulting to host unless specified with `--network` or using a service container

GuessWhoSamFoo:
@wolfogre @ChristopherHX Still missing some tests, but wanted to run by the feedback so far. The `host` network will be used by default unless a service container is defined or passed through the flag. This can mitigate some concerns around breaking existing users and orphaned networks eventually reaching limits.

I have also thought about creating a network name by convention and re-using instead. Higher chance of port conflicts but less complexity

pkg/runner/run_context.go
```diff
@@ -379,6 +446,40 @@ func (rc *RunContext) stopJobContainer() common.Executor {
...
+		}
+		return common.NewParallelExecutor(len(execs), execs...)(ctx)
+	}
+}
+
+func (rc *RunContext) stopServiceContainers() common.Executor {
+	return func(ctx context.Context) error {
+		execs := []common.Executor{}
+		for _, c := range rc.ServiceContainers {
+			execs = append(execs, c.Remove())
```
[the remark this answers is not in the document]
GuessWhoSamFoo: Thanks for the pointers!

pkg/runner/run_context.go
```diff
@@ -269,31 +340,34 @@ func (rc *RunContext) startJobContainer() common.Executor {
...
-			NetworkMode: "host",
-			Binds:       binds,
-			Stdout:      logWriter,
-			Stderr:      logWriter,
-			Privileged:  rc.Config.Privileged,
-			UsernsMode:  rc.Config.UsernsMode,
-			Platform:    rc.Config.ContainerArchitecture,
-			Options:     rc.options(ctx),
+			Cmd:            nil,
+			Entrypoint:     []string{"/usr/bin/tail", "-f", "/dev/null"},
```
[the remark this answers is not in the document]
GuessWhoSamFoo: Rebased and reverted this line back to `tail`

hariso:
Kudos for this change, I'm really looking forward to it! Thanks @GuessWhoSamFoo and all reviewers!

pkg/runner/run_context.go
```diff
@@ -259,7 +271,66 @@ func (rc *RunContext) startJobContainer() common.Executor {
...
+			}
+			envs := make([]string, 0, len(interpolatedEnvs))
+			for k, v := range interpolatedEnvs {
+				envs = append(envs, fmt.Sprintf("%s=%s", k, v))
+			}
+			username, password, err = rc.handleServiceCredentials(ctx, spec.Credentials)
+			if err != nil {
+				return fmt.Errorf("failed to handle service %s credentials: %w", serviceID, err)
+			}
+			serviceBinds, serviceMounts := rc.GetServiceBindsAndMounts(spec.Volumes)
```
ZauberNerd, line 294: Volumes, Image, port and options should all be evaluated: https://github.com/ZauberNerd/act-test/actions/runs/5939905007/workflow

ZauberNerd:
I'm having trouble getting networking to work with this example workflow: https://github.com/ZauberNerd/act-test/actions/runs/5940389109/workflow

pkg/runner/run_context.go
```diff
@@ -259,7 +271,66 @@ func (rc *RunContext) startJobContainer() common.Executor {
...
+					logger.Errorf("Error while cleaning services: %v", err)
+				}
+				if !rc.IsHostEnv(ctx) {
+					// clean network in docker mode only
+					// if using service containers
+					// it means that the network to which containers are connecting is created by `act_runner`,
+					// so, we should remove the network at last.
+					logger.Infof("Cleaning up network for job %s, and network name is: %s", rc.JobName, rc.networkName())
+					if err := container.NewDockerNetworkRemoveExecutor(rc.networkName())(ctx); err != nil {
+						logger.Errorf("Error while cleaning network: %v", err)
```
ZauberNerd, line 330: I'm running into:
`Error while cleaning network: Error response from daemon: error while removing network: network act-push-job-2-cf0efa14d6309df9584a42f3b9aaec4bca252b8008d44003059bdc53346e5553-job-2-network id c65ca709ee6811b8c8907bf864aff5b3bb8ad65b0a62b841413d67dce8529ed2 has active endpoints`
when I stop a workflow with ctrl+c
And after that starting the workflow again results in:
`Error: failed to start container: Error response from daemon: network act-push-job-1-b5db9dd46498cf67a2d34f06188bd6188e01956cd86b4dbf3fe53923246b1467-job-1-network is ambiguous (2 matches found on name)`
Here's a simple example workflow to reproduce this behavior: https://github.com/ZauberNerd/act-test/actions/runs/5940389109/workflow

GuessWhoSamFoo: I'll take a closer look at this later in the week

GuessWhoSamFoo: Feel free to make any changes needed

wolfogre (dismissed):

ZauberNerd (changes requested):
> Mode (HostEnvironment) of job1 has no support for containers. You cannot tell docker to reassign docker local ports, otherwise breaks type job2.

Based on https://docs.github.com/en/actions/using-containerized-services/about-service-containers#communicating-with-service-containers and some tests, I would assume the following:
1. services are always started in their own network and can communicate with each other (https://github.com/ZauberNerd/act-test/actions/runs/5954867123/job/16152325020#step:14:26)
2. when the job itself runs in a container, the job container joins the service network (https://docs.github.com/en/actions/using-containerized-services/about-service-containers#running-jobs-in-a-container)
3. when the job runs on the "host" it can only communicate with services that have published their ports onto the host (https://docs.github.com/en/actions/using-containerized-services/about-service-containers#running-jobs-on-the-runner-machine)

I think we can implement this as follows:
2. check if `job.services && job.container` are set and create a bridge network for both, services and the job container
3. if only `job.services` are set create a bridge network for the services and run the job container in host mode

Or am I missing something?

Would that work for gitea ~~@GuessWhoSamFoo~~ @wolfogre or do you have any other networking requirements?

ZauberNerd:
I think it should still work (at least in act with regular containers, maybe not necessarily with the Host implementation), because both containers would run on the "host" (in that case the VM that Docker for Desktop creates on Windows/Linux).

Example (I could only test this on Linux and MacOS, but should work on Windows the same):
```sh
docker network create test
docker run --rm -p 8080:80 --network test --name nginx nginx:latest
docker run --rm --network test curlimages/curl:latest curl -sI http://nginx:80
docker run --rm --network host curlimages/curl:latest curl -sI http://localhost:8080
```

I'll try out a few things this week on top of this PR to see if I can get it to work as described above.

If not, I would propose to just allow the scenario of `job-2` (from this example: https://github.com/ZauberNerd/act-test/actions/runs/5940389109/workflow#L18-L23) and error out in case someone tries to use services without job containers (scenario `job-1` in my linked workflow).

ZauberNerd:
@GuessWhoSamFoo @ChristopherHX I've added three commits to add tests and implement the behavior I described above. Could you please verify that this works for you?

cmd/root.go
```diff
@@ -96,6 +97,7 @@ func Execute(ctx context.Context, version string) {
 	rootCmd.PersistentFlags().StringVarP(&input.cacheServerAddr, "cache-server-addr", "", common.GetOutboundIP().String(), "Defines the address to which the cache server binds.")
 	rootCmd.PersistentFlags().Uint16VarP(&input.cacheServerPort, "cache-server-port", "", 0, "Defines the port where the artifact server listens. 0 means a randomly available port.")
 	rootCmd.PersistentFlags().StringVarP(&input.actionCachePath, "action-cache-path", "", filepath.Join(CacheHomeDir, "act"), "Defines the path where the actions get cached and host workspaces created.")
+	rootCmd.PersistentFlags().StringVarP(&input.networkName, "network", "", "host", "Sets a docker network name. Defaults to host.")
```
ZauberNerd, line 100: @GuessWhoSamFoo what's the reason behind this flag?

pkg/runner/run_context.go
```diff
@@ -87,6 +88,18 @@ func (rc *RunContext) jobContainerName() string {
 	return createContainerName("act", rc.String())
 }
 
+// networkName return the name of the network which will be created by `act` automatically for job,
+// only create network if using a service container
+func (rc *RunContext) networkName() (string, bool) {
```
ZauberNerd, line 93: What should this function return for the service container network if `--network` is specified? Should the service containers also use the user-supplied network?

ZauberNerd (commented):
Overall I'm pretty happy with these changes and I have just one question left about how the `--network` CLI flag came to be (https://github.com/nektos/act/pull/1949/commits/d0aad7fa18ce66c069486dfe75a8d8c9e33fbbbb)

ZauberNerd:
> To be honest using service container without container delete and no reuse flags has caused inconsistent container network configurations...

Can you explain in more detail or give an example to reproduce?

ZauberNerd:
> * Create a workflow with a failung job with service container (happened while running the test with broken act)
> * Run the failing job
> * You will find the service container via `docker ps -a` after exit (--rm will cause container removal)
> * Now add a port mapping or otherwise change networking
> * The service container is reused with old network config see `docker inspect`
> 
> Maybe because we don't recreate service container if they are not deleted (we deleted the old job container at least in the past, without `--reuse`)

@ChristopherHX I pushed two commits to deal with this issue. Firstly we now always remove service containers, regardless of the CLI flags and secondly I had to check if a network is still in use before deleting it (got an error otherwise).
Could you please take a look if that is the correct way to address these issues?

pkg/runner/run_context.go
```diff
@@ -369,16 +467,50 @@ func (rc *RunContext) UpdateExtraPath(ctx context.Context, githubEnvPath string)
...
+		return common.NewParallelExecutor(len(execs), execs...)(ctx)
+	}
+}
+
+func (rc *RunContext) startServiceContainers(_ string) common.Executor {
+	return func(ctx context.Context) error {
+		execs := []common.Executor{}
+		for _, c := range rc.ServiceContainers {
+			execs = append(execs, common.NewPipelineExecutor(
+				c.Pull(false),
```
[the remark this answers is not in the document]
ZauberNerd: > stopJobContainer isn't called if no `--rm` is used and a job fails, so probably no.

Isn't that the intended behavior? 
From reading the flag descriptions:
`--reuse`: don't remove container(s) on successfully completed workflow(s) to maintain state between runs
`--rm`: automatically remove container(s)/volume(s) after a workflow(s) failure

I think we can infer that:

* with no flags passed, only containers of successful jobs should be removed
* for failed jobs the containers should be kept to allow debugging the issue inside the container
* with `--rm` passed, containers should always be removed, regardless of the job outcome
* with `--reuse` passed, containers should always be kept running to speed up subsequent `act` invocations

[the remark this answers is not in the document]
ZauberNerd: Ah, you mean that `stopJobContainer()` was always called previously before starting the next container. I'll take a look if I can restore that.

ZauberNerd (approved):
From my side this is now ready for review.

ZauberNerd (approved):

devnoname120:
PR to update the documentation accordingly: https://github.com/nektos/act-docs/pull/30

PR #1954, opened 2023-08-08
feat: cli option to enable the new action cache

ChristopherHX:
This is my initial code to switch the action cache over to my new implementation

Until all tests are migrated this will be kept as draft to avoid mergify spam.

I would be careful to test this change, to detect known issues of my previous PR's and this change.

Checklist
- [ ] Test symlinked action.yml file in remote actions
- [ ] Test symlinked action.yml file in local actions
- [ ] Test remote docker actions
- [ ] Test using images with a non root container (I expect we still need to fix permissions)
- [ ] Test error messages like
   - [ ] action.yml not found
   - [ ] Clone failed
- [ ] Make Tests Green (If I add a Feature Toggle for this change the test don't need to change)

Do you prefer to add a Feature Toggle for this change? Then we can merge this while iron out remaining bugs later and make it possible to make use of this change for people affected by the current problems.


cmd/root.go
```diff
@@ -613,6 +614,11 @@ func newRunCommand(ctx context.Context, input *Input) func(*cobra.Command, []str
 			ReplaceGheActionTokenWithGithubCom: input.replaceGheActionTokenWithGithubCom,
 			Matrix:                             matrixes,
 		}
+		if input.useNewActionCache {
+			config.ActionCache = &runner.GoGitActionCache{
+				Path: CacheHomeDir,
```
ZauberNerd, line 619:
```suggestion
				Path: config.ActionCacheDir,
```

cplee (approved):
tested locally 👍

PR #1966, opened 2023-08-14
Update readme: var option

YubinMoon:
I added variables option content to the secrets area.


ZauberNerd (commented):
I'm not sure how much we still want to change in the readme, but we also have the documentation repository over here: https://github.com/nektos/act-docs where the official docs reside: https://nektosact.com/

PR #1969, opened 2023-08-15
Update install instructions on Fedora

goncalossilva:
Disclaimer: I'm the author of this copr repo.

goncalossilva/cli has been up and running for much longer than the official recommendation, and over the years, it's been more reliable with its builds. See:
- https://github.com/goncalossilva/rpm-act/actions
- https://copr.fedorainfracloud.org/coprs/goncalossilva/act/monitor/

It's also fully automated. This means stable versions of Fedora get new builds quickly (even if builds on rawhide/preview versions tend to fail while the dust settles).

I think this is a better default choice for `act` users, but other than this PR, it's up to the reviewer. Regardless of it being merged or not, goncalossilva/act will continue to be built. :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,10 +87,10 @@ winget install nektos.act
 yay -Syu act
 ```
 
-### [COPR](https://copr.fedorainfracloud.org/coprs/rubemlrm/act-cli/) (Linux)
+### [COPR](https://copr.fedorainfracloud.org/coprs/goncalossilva/act/) (Linux)
 
 ```shell
-dnf copr enable rubemlrm/act-cli
+dnf copr enable goncalossilva/act
 dnf install act-cli
 ```
 


iTrooz:
Its really weird, the `--version` flag doesn't exist in your package

goncalossilva:
Huh, same here. It does look bizarre:

```
~ $ dnf list --installed act-cli
Installed Packages
act-cli.x86_64                                                                                               0.2.49-1.fc38                                                                                                @copr:copr.fedorainfracloud.org:goncalossilva:act
~ $ act --version
Error: unknown flag: --version
```

But it boils down to the version being [embedded at compile time](https://github.com/nektos/act/blob/7ba9f30f37af627f43adc2776c7f39ca3b5889bc/main.go#L13), and [the spec passing it wrongly](https://github.com/goncalossilva/rpm-act/blob/623f2cbacc39b1555e0578e56d00bcda194a1b32/act-cli.spec#L40C18-L40C18). It's based on [this repo's build](https://github.com/nektos/act/blob/7ba9f30f37af627f43adc2776c7f39ca3b5889bc/Makefile#L26), but within an RPM spec, it should be `%{version}` instead of ` ${version}`.

All fixed in https://github.com/goncalossilva/rpm-act/commit/f084af37b79bb0afc4052402ec64c5793f61cfcd. `--version` will work as expected on the next version. Thanks for flagging!

goncalossilva:
Migrated to: https://github.com/nektos/act-docs/pull/13

PR #1970, opened 2023-08-15
fix: fail if no stages were found

ctrlaltf24:
Adds a warning message if act is cannot find any stages to run with the filters provided.

Reproduction:
- run `act -j gibberish`

Desired behavior: some indication I did something silly
Actual behavior: no output, just exit with success.

As a human who often makes spelling mistakes,
it would be nice if act warned me what I was doing that was silly rather than exiting apparently doing
nothing with no obvious indication
I did something wrong.

Alternative solutions:
- Do nothing, it's the user's fault: True it is, would be nice to give them a hint where to resolve their issue
- Exit with error rather than warn: Reasonable behavior as well, not knowledgeable enough with the code base to know for sure this is enough justification to exit

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -527,9 +527,15 @@ func newRunCommand(ctx context.Context, input *Input) func(*cobra.Command, []str
 		if jobID != "" {
 			log.Debugf("Planning job: %s", jobID)
 			plan, plannerErr = planner.PlanJob(jobID)
+			if len(plan.Stages) == 0 {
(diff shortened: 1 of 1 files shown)


ZauberNerd (changes requested):
I think it would make more sense to return an error from here: https://github.com/nektos/act/blob/7ba9f30f37af627f43adc2776c7f39ca3b5889bc/pkg/model/planner.go#L331-L358
As you can see the `err` variable is unused, so I'd probably remove it and use the `if len(stages) == 0` to return `nil, fmt.Errorf("...")`

ZauberNerd (approved):

Issue #1967, opened 2023-08-14
Allow node20 in action.yml as action runtime


ChristopherHX:
### Act version

v0.2.49

### Feature description

See https://github.com/actions/runner/releases/tag/v2.308.0

Since act doesn't care about the version behind `node` why not just exec all `node<number>` via `node`?

gilzow:
Just ran into this issue as well where one of the actions in my workflow has updated to node20, but act now reports:

```
The runs.using key in action.yml must be one of: [composite docker node12 node16], got node20
```
and fails the action

mahmudz:
Same issue here

`
Error: The runs.using key in action.yml must be one of: [composite docker node12 node16], got node20
`
OS: Mac 15 Sonoma Beta 7
Act Version: 0.2.50
Docker Version: 4.21.1

MichelFR:
Same issue here, any workaround?
Using in gitea actions

mahmudz:
@MichelFR Whats the procedure?

NexZhu:
@MichelFR Same issue here. Could you explain in detail?

teffcz:
I have same problem with 
"The runs.using key in action.yml must be one of: [composite docker node12 node16 go], got node20" 

running basic CI job for building docker image using:
docker/setup-qemu-action@master
docker/setup-buildx-action@master
docker/login-action@master
docker/build-push-action@master

with docker image ghcr.io/catthehacker/ubuntu:act-20.04, but i tried sereval others (even node:20-bullseye) and nothing helped

from logs it looks it is failing at check docker/setup-qemu-action@master

jcarpenter12:
> > Same issue here, any workaround? Using in gitea actions
> > Edit:
> > Okay found a workaround for now, needed to turn all versions back of the actions
> 
> Can you tell me how to apply this workaround?

I had this issue this morning but the new version of act now supports node20 so it now works after upgrading act, unless I have misunderstood your question

PR #1988, opened 2023-08-31
feat: support node20 runtime

freshollie:
Closes https://github.com/nektos/act/issues/1967


freshollie:
Not sure I've done everything needed. Wanna see how CI deals with it

harryzcy:
@freshollie Can you also include node18 in this PR?

freshollie:
@cplee @ChristopherHX any reason not to merge this?

harryzcy:
Is there a reason node_modules are included in Git, instead of installing them in the test?

freshollie:
> A node20 test could be also without any node dependency, by using console.log / fs.writeSync(process.env.GITHUB_OUTPUT, 'out=test', { encoding: 'utf-8' }).

Makes sense, I just copied convention.

> The current approuch of adding node versions blows up the code of act and soon we have node12, node16, node20, node24 and we executute all with node
> I have had patched my act fork to ignore the number behind node.
There is definitely a better solution. This change for now allows people who have workflows on node20 to continue to use act.

cplee (approved):

harryzcy:
Thank you for reviewing cplee.

If I understand correctly, the cron job for next release runs at the end of the month. But since node16 already reached EOL on Sep 11th, can you manually trigger the CI this time? Thanks.

PR #2023, opened 2023-09-24
Add support for NetBSD

iamleot:
NetBSD can run Docker CLI and then use Docker on some remote machine via `DOCKER_HOST`.

(This can be probably extended to all other Unix-es capable of running just Docker CLI.)

---

This was done mechanically via:

```console
sed -i -e 's;linux || darwin || windows;linux || darwin || windows || netbsd;' **/*.go
```

---

I have tested it only on NetBSD/amd64 via Lima VM and `DOCKER_HOST` set to `unix:///${HOME}/.lima/docker/sock/docker.sock`.

I think that the same method will probably works on several other Unix-es (`dragonfly`, `freebsd`, `illumos`, `openbsd`, `solaris` and maybe others that I could not think of!) as long as they support at least Docker CLI. I have being conservative here to only what I was able to test though, sorry!


iamleot:
(Rebasing on latest `master` and pushing with the `--force` to avoid drifts.)

iamleot:
Hello @ChristopherHX!
Thanks, I will try to look at HEAD `master` in the next weekend and also adjust paths related to service support.

Marking as draft in the meantime and please ignore it for the moment!

Thanks!

cplee (approved):

PR #2061, opened 2023-10-23
fix: compute ref and ref_name prioritising HEAD pointer if tag is not set

### LINKED ISSUE #2056: github.ref_name is not set to the branch/tag that HEAD points to.
### Bug report info

```plain text
act version:            0.2.52
GOARCH:                 arm64
NumCPU:                 8
Docker host:            unix:///Users/dpc-g/.colima/default/docker.sock
Sockets found:
	$HOME/.colima/docker.sock
Config files:
	/Users/dpc-g/.actrc:
		-P ubuntu-latest=catthehacker/ubuntu:full-latest
		-P ubuntu-latest=catthehacker/ubuntu:full-20.04
		-P ubuntu-18.04=catthehacker/ubuntu:full-18.04
Build info:
	Go version:            go1.21.1
	Main version:
	Main path:
	Main checksum:
	Build settings:
		-buildmode:           exe
		-compiler:            gc
		-ldflags:             -X main.version=0.2.52
		CGO_ENABLED:          1
		CGO_CFLAGS:
		CGO_CPPFLAGS:
		CGO_CXXFLAGS:
		CGO_LDFLAGS:
		GOARCH:               arm64
		GOOS:                 darwin
Docker Engine:
	Engine version:        23.0.6
	Engine runtime:        runc
	Cgroup version:        1
	OS:                    Alpine Linux v3.18
	OS type:               linux
	OS version:            3.18.4
	OS arch:               aarch64
	OS kernel:             6.1.55-0-virt
	OS CPU:                4
	OS memory:             7938 MB
	Security options:
		name=seccomp,profile=builtin
```

### Command used with act

```sh
act schedule -W .github/workflows/context-check.yml --defaultbranch "5.x" --container-architecture linux/amd64 --rm -P ubuntu-latest=catthehacker/ubuntu:act-latest -v
```

### Describe issue

If a Git reference points to multiple branches the last pointer in the series is selected as the value for github.ref_name

i.e. 

```
6cae0a05 (HEAD -> 5.x, origin/5.x, build/git_ref-test) Ref with multiple branch pointers
```

### Expected behaviour (personally)
The values for "ref" and "ref_name" in the GitHub context are set to the branch/tag pointer that `HEAD` is pointing to i.e.
```
|   "ref": "refs/heads/5.x",
|   "ref_name": "5.x",
```

### Actual behaviour
See the output posted in the "Relevant log output" section.

### Link to GitHub repository

_No response_

### Workflow content

```yml
name: Test context values
run-name: GitHub context
on:
  schedule:
    - cron: '23 20 * * 0'
jobs:
  github_context:
    runs-on: ubuntu-latest
    steps:
      - name: Dump GitHub context
        env:
          GITHUB_CONTEXT: ${{ toJson(github) }}
        run: echo "$GITHUB_CONTEXT"
```

### Relevant log output

```sh
[Test context values/github_context] [DEBUG] Working directory '/repo/dir'
| {
|   "event": {
|     "repository": {
|       "default_branch": "5.x"
|     }
|   },
|   "event_path": "/var/run/act/workflow/event.json",
|   "workflow": "Test context values",
|   "run_id": "1",
|   "run_number": "1",
|   "actor": "GROwen",
|   "repository": "dpc-sdp/bay",
|   "event_name": "schedule",
|   "sha": "6cae0a053dc2fe331ddac2b3b0f6b8c8fb6b0269",
|   "ref": "refs/heads/build/git_ref-test",
|   "ref_name": "build/git_ref-test",
|   "ref_type": "branch",
|   "head_ref": "",
|   "base_ref": "",
|   "token": "***",
|   "workspace": "/Users/dpc-g/Projects/bay",
|   "action": "0",
|   "action_path": "",
|   "action_ref": "",
|   "action_repository": "",
|   "job": "github_context",
|   "job_name": "",
|   "repository_owner": "dpc-sdp",
|   "retention_days": "0",
|   "runner_perflog": "/dev/null",
|   "runner_tracking_id": "",
|   "server_url": "https://github.com",
|   "api_url": "https://api.github.com",
|   "graphql_url": "https://api.github.com/graphql"
| }
```

### Additional information

I was trying to solve this and provide a PR but am uncertain of the broader implications of the change that would fix the issue for my scenario.

[pkg/common/git/git.go:FindGetRef](https://github.com/nektos/act/blob/master/pkg/common/git/git.go#L71) is responsible for setting the values. I'm assuming the requirement for this and not just the value set for head in [pkg/common/git/git.go:FindGitRevision](https://github.com/nektos/act/blob/master/pkg/common/git/git.go#L71) is to lookup any tags that would be related to the revision. Otherwise I'd expect the pointer for HEAD could be used?

It'd be helpful to have this assumption verified by someone more experienced with the project before I dive into a solution.

[Expanding the existing tests](https://github.com/nektos/act/blob/master/pkg/common/git/git_test.go#L164) to check for this scenario seems trivial although I don't think that explicitly checks for what would be output to github.ref_name. Could I have missed that somewhere?
```
		"current_head_is_master_and_pointer_exists_to_another_branch": {
			Prepare: func(t *testing.T, dir string) {
				require.NoError(t, gitCmd("-C", dir, "checkout", "-b", "mybranch"))
				require.NoError(t, gitCmd("-C", dir, "checkout", "-b", "master"))
			},
			Assert: func(t *testing.T, ref string, err error) {
				require.NoError(t, err)
				require.Equal(t, "refs/heads/master", ref)
			},
		},
```

GROwen:
Fixes nektos/act#2056


pkg/common/git/git.go
```diff
@@ -86,78 +86,49 @@ func FindGitRevision(ctx context.Context, file string) (shortSha string, sha str
...
-		 */
-		if r.Hash().String() == ref {
-			if r.Name().IsTag() {
-				refTag = r.Name().String()
-			}
-			if r.Name().IsBranch() {
-				refBranch = r.Name().String()
-			}
-		}
+	// The assumption is made that if the revision has a tag associated with it the ref and ref_name should reference it.
```
GROwen, line 106: I’m not confident of when the ref and ref_name are set to the `refs/tags` prefix and therefore whether or not this assumption is accurate or even desired.

I’m thinking this is relevant to working with the [release workflow trigger](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#release). 

Either way I’m not looking to change the current behaviour but just fix the value when multiple branches point to the same commit.

GROwen:
The change that this PR is conflicting with could resolve the issue. The change introduced in https://github.com/nektos/act/pull/1530/files needs to be tested against the scenario that was raised in https://github.com/nektos/act/issues/2056 before I can continue work on this PR.

PR #2062, opened 2023-10-24
feat: support runs-on labels and group

raffis:
## Current situation

act does not work with our worklfow declarations which are as follow:
```yaml
  build:
    runs-on:
      labels:
        - generic-v2
```

acts `runs-on` only supports a string or or a string slice. However the github workflow schema does support group and labels keys.
See https://github.com/SchemaStore/schemastore/blob/master/src/schemas/json/github-workflow.json#L703-L721 for reference.

## Proposal
This pr actually supports this use case. It will just combine labels and optional group declarations in the runs on output slice. This should be suitable for cover use cases for custom images targeting these labels.


raffis:
Any update on this?

cplee (approved):

PR #2063, opened 2023-10-24
feat: support config env expansion

raffis:
## Current situation
Using an .actrc config it is not possible to utilize envs as there is no env expansion since its not invoked from a shell.

I'd like to have a config like this which is currently not working:
```
--detect-event
-s GITUB_TOKEN=$GITHUB_TOKEN
-s GH_ACTIONS=$GITHUB_TOKEN
--container-options --privileged --mount type=bind,source=/home/$USER/.docker-act,target=/home/runner/.docker
```

## Proposal
Expand any envs while appending the caller args.


wolfogre (approved):
LGTM. I don't see any reason why users would want to keep the original string `$XXX` in args. If they do, they can use `$$XXX`.

PR #2088, opened 2023-11-11
Support array expressions in runs-on

### LINKED ISSUE #2080: `runs-on` does not accept array expressions
### Bug report info

```plain text
act version:            0.2.53
GOARCH:                 arm64
NumCPU:                 12
Sockets found:
	/var/run/docker.sock
	$HOME/.docker/run/docker.sock
Config files:
	/Users/jonj/.actrc:
		-P ubuntu-latest=node:16-buster-slim
		-P ubuntu-22.04=node:16-bullseye-slim
		-P ubuntu-20.04=node:16-buster-slim
		-P ubuntu-18.04=node:16-buster-slim
Build info:
	Go version:            go1.21.3
	Main version:
	Main path:
	Main checksum:
	Build settings:
		-buildmode:           exe
		-compiler:            gc
		-ldflags:             -X main.version=0.2.53
		CGO_ENABLED:          1
		CGO_CFLAGS:
		CGO_CPPFLAGS:
		CGO_CXXFLAGS:
		CGO_LDFLAGS:
		GOARCH:               arm64
		GOOS:                 darwin
Docker Engine:
	Engine version:        24.0.6
	Engine runtime:        runc
	Cgroup version:        2
	OS:                    Docker Desktop
	OS type:               linux
	OS version:
	OS arch:               aarch64
	OS kernel:             6.4.16-linuxkit
	OS CPU:                12
	OS memory:             7844 MB
	Security options:
		name=seccomp,profile=unconfined
		name=cgroupns
```

### Command used with act

```sh
act -vW .github/workflows/sample.yml
```

### Describe issue

If `runs-on` receives an array expression, act incorrectly evaluates it to `%!t(string=Array)`. act correctly handles string expressions, as well as static string and array yaml values.

I.e. these work:
* `runs-on: ubuntu-latest`
* `runs-on: ${{ fromJSON('"ubuntu-latest"') }}`
* `runs-on: [ubuntu-latest]`

But this does not:
* `runs-on: ${{ fromJSON('["ubuntu-latest"]') }}`

The bug isn't specific to `fromJSON`, as the same issue occurs if the expression is an array value from a matrix, e.g.

```yaml
runs-on: ${{ matrix.runner }}
strategy:
  matrix:
    runner: [[ubuntu-latest], [self-hosted, my-cool-label]]
```

These scenarios all work as expected in actual GitHub Actions workflow runs.

### Link to GitHub repository

https://github.com/jenseng/dynamic-uses/actions/runs/6801153354/job/18491354468

### Workflow content

```yml
name: test
on: push
jobs:
  test:
    # runs-on: ubuntu-latest
    # runs-on: ${{ fromJSON('"ubuntu-latest"') }}
    # runs-on: [ubuntu-latest]
    runs-on: ${{ fromJSON('["ubuntu-latest"]') }}
    steps:
      - run: echo "Hello, world!"
```

### Relevant log output

```sh
WARN  ⚠ You are using Apple M-series chip and you have not specified container architecture, you might encounter issues while running act. If so, try running it with '--container-architecture linux/amd64'. ⚠
DEBU[0000] Loading environment from /Users/jonj/projects/dynamic-uses/.env
DEBU[0000] Loading action inputs from /Users/jonj/projects/dynamic-uses/.input
DEBU[0000] Loading secrets from /Users/jonj/projects/dynamic-uses/.secrets
DEBU[0000] Loading vars from /Users/jonj/projects/dynamic-uses/.vars
DEBU[0000] Evaluated matrix inclusions: map[]
DEBU[0000] Loading workflow '/Users/jonj/projects/dynamic-uses/.github/workflows/sample.yml'
DEBU[0000] Reading workflow '/Users/jonj/projects/dynamic-uses/.github/workflows/sample.yml'
DEBU[0000] Preparing plan with all jobs
DEBU[0000] Using the only detected workflow event: push
DEBU[0000] Planning jobs for event: push
DEBU[0000] Conditional GET for notices etag=c694094f-9a86-4866-bf11-876c821d9fcb
DEBU[0000] gc: 2023-11-08 09:31:06.262245 -0700 MST m=+0.030477292  module=artifactcache
DEBU[0000] Plan Stages: [0x1400011ee10]
DEBU[0000] Stages Runs: [test]
DEBU[0000] Job.Name: test
DEBU[0000] Job.RawNeeds: {0 0    <nil> []    0 0}
DEBU[0000] Job.RawRunsOn: {8 0 !!str ${{ fromJSON('["ubuntu-latest"]') }}  <nil> []    8 14}
DEBU[0000] Job.Env: {0 0    <nil> []    0 0}
DEBU[0000] Job.If: {0 0  success()  <nil> []    0 0}
DEBU[0000] Job.Steps: echo "Hello, world!"
DEBU[0000] Job.TimeoutMinutes:
DEBU[0000] Job.Services: map[]
DEBU[0000] Job.Strategy: <nil>
DEBU[0000] Job.RawContainer: {0 0    <nil> []    0 0}
DEBU[0000] Job.Defaults.Run.Shell:
DEBU[0000] Job.Defaults.Run.WorkingDirectory:
DEBU[0000] Job.Outputs: map[]
DEBU[0000] Job.Uses:
DEBU[0000] Job.With: map[]
DEBU[0000] Job.Result:
DEBU[0000] Empty Strategy, matrixes=[map[]]
DEBU[0000] Job Matrices: [map[]]
DEBU[0000] Runner Matrices: map[]
DEBU[0000] Final matrix after applying user inclusions '[map[]]'
DEBU[0000] Loading revision from git directory
DEBU[0000] Found revision: 4af4b6a3356b02639095850fe2dce350906070ba
DEBU[0000] HEAD points to '4af4b6a3356b02639095850fe2dce350906070ba'
DEBU[0000] using github ref: refs/heads/debug-act-bug3
DEBU[0000] Found revision: 4af4b6a3356b02639095850fe2dce350906070ba
DEBU[0000] Detected CPUs: 12
[test/test] [DEBUG] evaluating expression 'success()'
[test/test] [DEBUG] expression 'success()' evaluated to 'true'
[test/test] [DEBUG] expression '${{ fromJSON('["ubuntu-latest"]') }}' rewritten to 'format('{0}', fromJSON('["ubuntu-latest"]'))'
[test/test] [DEBUG] evaluating expression 'format('{0}', fromJSON('["ubuntu-latest"]'))'
[test/test] [DEBUG] expression 'format('{0}', fromJSON('["ubuntu-latest"]'))' evaluated to '%!t(string=Array)'
[test/test] [DEBUG] expression '${{ fromJSON('["ubuntu-latest"]') }}' rewritten to 'format('{0}', fromJSON('["ubuntu-latest"]'))'
[test/test] [DEBUG] evaluating expression 'format('{0}', fromJSON('["ubuntu-latest"]'))'
[test/test] [DEBUG] expression 'format('{0}', fromJSON('["ubuntu-latest"]'))' evaluated to '%!t(string=Array)'
[test/test] 🚧  Skipping unsupported platform -- Try running with `-P Array=...`
DEBU[0000] Saving notices etag=c694094f-9a86-4866-bf11-876c821d9fcb
DEBU[0000] No new notices
```

### Additional information

_No response_

jenseng:
Ensure array expressions in `runs-on` are handled correctly. Also fix `-P` to work case-insensitively, since labels are not case sensitive.

There is still one outstanding issue with `runs-on` and expressions, but that is not new to this PR: a matrix with a templated runs-on (e.g. `runs-on: ${{ matrix.label }}`). That feels like a bigger fix/refactor, so I was thinking of filing a new issue issue/PR for that, but if you'd prefer I can incorporate it into this one.

Fixes #2080

### DIFF
--- a/pkg/model/workflow.go
+++ b/pkg/model/workflow.go
@@ -272,20 +272,23 @@ func (j *Job) Needs() []string {
 	return nil
 }
 
-// RunsOn string or list for Job
-func (j *Job) RunsOn() (runsOnString string, runsOnList []string, err error) {
+// RunsOn list for Job
+func (j *Job) RunsOn() []string {
 	switch j.RawRunsOn.Kind {
 	case yaml.ScalarNode:
-		if decodeNode(j.RawRunsOn, &runsOnString) && runsOnString != "" {
-			return
+		var val string
+		if !decodeNode(j.RawRunsOn, &val) {
+			return nil
 		}
+		return []string{val}
 	case yaml.SequenceNode:
-		if decodeNode(j.RawRunsOn, &runsOnList) && len(runsOnList) > 0 {
-			return
+		var val []string
+		if !decodeNode(j.RawRunsOn, &val) {
+			return nil
 		}
+		return val
 	}
-	err = fmt.Errorf("'runs-on' key not defined or invalid")
-	return
+	return nil
 }
 
 func environment(yml yaml.Node) map[string]string {
--- a/pkg/model/workflow.go
+++ b/pkg/model/workflow.go
@@ -272,23 +272,20 @@ func (j *Job) Needs() []string {
 	return nil
 }
 
-// RunsOn list for Job
-func (j *Job) RunsOn() []string {
+// RunsOn string or list for Job
+func (j *Job) RunsOn() (runsOnString string, runsOnList []string, err error) {
 	switch j.RawRunsOn.Kind {
 	case yaml.ScalarNode:
-		var val string
-		if !decodeNode(j.RawRunsOn, &val) {
-			return nil
+		if decodeNode(j.RawRunsOn, &runsOnString) && runsOnString != "" {
+			return
 		}
-		return []string{val}
 	case yaml.SequenceNode:
-		var val []string
-		if !decodeNode(j.RawRunsOn, &val) {
-			return nil
+		if decodeNode(j.RawRunsOn, &runsOnList) && len(runsOnList) > 0 {
+			return
 		}
-		return val
 	}
-	return nil
+	err = fmt.Errorf("'runs-on' key not defined or invalid")
+	return
 }
 
 func environment(yml yaml.Node) map[string]string {
--- a/pkg/runner/run_context.go
+++ b/pkg/runner/run_context.go
@@ -626,14 +626,7 @@ func (rc *RunContext) containerImage(ctx context.Context) string {
 }
 
 func (rc *RunContext) runsOnImage(ctx context.Context) string {
-	job := rc.Run.Job()
-
-	if job.RunsOn() == nil {
-		common.Logger(ctx).Errorf("'runs-on' key not defined in %s", rc.String())
-	}
-
-	for _, runnerLabel := range job.RunsOn() {
-		platformName := rc.ExprEval.Interpolate(ctx, runnerLabel)
+	for _, platformName := range rc.runsOnPlatformNames(ctx) {
 		image := rc.Config.Platforms[strings.ToLower(platformName)]
 		if image != "" {
 			return image
@@ -641,6 +634,44 @@ func (rc *RunContext) runsOnImage(ctx context.Context) string {
 	}
 
 	return ""
+}
+
+func (rc *RunContext) runsOnPlatformNames(ctx context.Context) []string {
+	job := rc.Run.Job()
+
+	runsOnString, runsOnList, err := job.RunsOn()
+	if err != nil {
+		common.Logger(ctx).Errorf("'runs-on' key not defined or invalid in %s", rc.String())
+	}
+
+	platformNames := []string{}
+	if runsOnString != "" {
+		var runsOn interface{}
+		if !strings.Contains(runsOnString, "${{") || !strings.Contains(runsOnString, "}}") {
+			runsOn = runsOnString
+		} else {
+			var runsOnErr error
+			runsOn, runsOnErr = rc.ExprEval.evaluate(ctx, runsOnString, exprparser.DefaultStatusCheckNone)
+			if runsOnErr != nil {
+				common.Logger(ctx).Errorf("Unable to interpolate expression '%s': %s", runsOnString, runsOnErr)
+			}
+		}
+
+		if platformNameList, ok := runsOn.([]interface{}); ok {
+			for _, platformName := range platformNameList {
+				platformNames = append(platformNames, fmt.Sprintf("%v", platformName))
+			}
+		} else {
+			platformNames = append(platformNames, fmt.Sprintf("%v", runsOn))
+		}
+	} else {
+		for _, runnerLabel := range runsOnList {
+			platformName := rc.ExprEval.Interpolate(ctx, runnerLabel)
+			platformNames = append(platformNames, platformName)
+		}
+	}
+
+	return platformNames
 }
 
 func (rc *RunContext) platformImage(ctx context.Context) string {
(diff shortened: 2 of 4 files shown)


pkg/runner/run_context.go
```diff
@@ -643,6 +636,44 @@ func (rc *RunContext) runsOnImage(ctx context.Context) string {
...
+		var runsOn interface{}
+		if !strings.Contains(runsOnString, "${{") || !strings.Contains(runsOnString, "}}") {
+			runsOn = runsOnString
+		} else {
+			var runsOnErr error
+			runsOn, runsOnErr = rc.ExprEval.evaluate(ctx, runsOnString, exprparser.DefaultStatusCheckNone)
+			if runsOnErr != nil {
+				common.Logger(ctx).Errorf("Unable to interpolate expression '%s': %s", runsOnString, runsOnErr)
+			}
+		}
```
jenseng, line 658: there's some overlap here with ExpressionEvaluator#Interpolate, but it seemed like sharing that would be a bigger refactor and not actually reduce any complexity. open to suggestions on better places to put this

pkg/runner/run_context.go
```diff
@@ -684,13 +715,8 @@ func (rc *RunContext) isEnabled(ctx context.Context) (bool, error) {
...
 	if img == "" {
-		if job.RunsOn() == nil {
-			l.Errorf("'runs-on' key not defined in %s", rc.String())
-		}
-
-		for _, runnerLabel := range job.RunsOn() {
-			platformName := rc.ExprEval.Interpolate(ctx, runnerLabel)
-			l.Infof("\U0001F6A7  Skipping unsupported platform -- Try running with `-P %+v=...`", platformName)
+		for _, platformName := range rc.runsOnPlatformNames(ctx) {
+			l.Infof("\U0001F6A7  Skipping unsupported platform -- Try running with `-P %+v=...`", strings.ToLower(platformName))
```
jenseng, line 719: added `ToLower` here to make the error messaging consistent with the existing (but I think undocumented?) behavior, i.e. any -P option needs to be lower case, since we only ever look up lower-cased keys in `rc.Config.Platforms`

pkg/model/workflow.go
```diff
@@ -272,23 +272,20 @@ func (j *Job) Needs() []string {
 	return nil
 }
 
-// RunsOn list for Job
-func (j *Job) RunsOn() []string {
+// RunsOn string or list for Job
```
jenseng, line 275: I considered leaving this method alone and always returning a `[]string`, and then in `runsOnPlatformNames` potentially spreading an expression array into `platformNames`, but that struck me as being less correct, since we wouldn't be able to differentiate between these two scenarios:

```yaml
# this is fine
runs-on: ${{ fromJSON('["my-label"]') }}
# this is not fine and will not work on GitHub
runs-on: [${{ fromJSON('["my-label"]') }}]
```

If we're not concerned about that and would prefer a slightly simpler implementation, I can tweak it

jenseng: Thanks for the yaml tip and the note about array merging, I'll go ahead simplify my change 🙏 

> I cannot even get my critical bugfix for windows merged or even reviewed with any comments, while a panic fix was merged

Oooof 😞

jenseng: Done ... though as you pointed out, matrix with templated runs-on is problematic (though that's the case before this change too), I'll file a separate issue for that, as I think the fix there will be a little more involved

pkg/runner/run_context.go
jenseng: Went ahead and just fixed the case sensitivity bug here since it was trivial... we'll see if the maintainer wants that in a separate change though

cplee (approved):

PR #2123, opened 2023-12-12
WorkflowDispatchConfig supports multiple yaml node kinds

pangliang:
I was working on PR https://github.com/go-gitea/gitea/pull/28163 . and found that the `workflow.WorkflowDispatchConfig()` method cannot be parsed correctly when the yaml node is not a mapping type

```
on: workflow_dispatch

on: [push, workflow_dispatch]

on:
    - push
    - workflow_dispatch

on:
    push:
    pull_request:
    workflow_dispatch:
        inputs:
            logLevel:
                description: 'Log level'
                required: true
                default: 'warning'
                type: choice
                options:
                - info
                - warning
                - debug
```

### DIFF
--- a/pkg/model/workflow.go
+++ b/pkg/model/workflow.go
@@ -5,6 +5,7 @@ import (
 	"io"
 	"reflect"
 	"regexp"
+	"slices"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


wolfogre:
Sorry, are you sure your cases are supported by GitHub? Have you tested them?

image

pangliang:
> SequenceNode

Yes, github has a special check and does not support nested SequenceNode; but the parsing method supports nesting to be more robust; what do you think?

wolfogre:
TBH, I don't think it's more robust, it will cause inconsistent behavior. And I don't see any reason why it should support `- [pull_request, workflow_dispatch]`.

pangliang:
> TBH, I don't think it's more robust, it will cause inconsistent behavior. And I don't see any reason why it should support `- [pull_request, workflow_dispatch]`.

I checked the [yaml spec](https://yaml.org/spec/1.2.2/),  chapter `Example 2.5 Sequence of Sequences`, The yaml language does have this structure. 

So I think this parse method is flow yaml spec, which is better than following the special restrictions of github; especially when this  act will be widely used by other gitxxx.

It's just my suggestion, but if you think it's better to follow github, I'll just delete the nested processing part.

wolfogre:
Please let me clarify, I'm not talking about the yaml specification. Of course, `- [pull_request, workflow_dispatch]` is supported by yaml, or else you wouldn't be able to parse it with `gopkg.in/yaml.v3`. I think we are discussing the [workflow syntax](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions).

pangliang:
Okay, here I flow github and add support for the following types:

```
on: workflow_dispatch

on: [push, workflow_dispatch]

on:
    - push
    - workflow_dispatch

on:
    push:
    pull_request:
    workflow_dispatch:
        inputs:
            logLevel:
                description: 'Log level'
                required: true
                default: 'warning'
                type: choice
                options:
                - info
                - warning
                - debug
```

pangliang:
There is another change. When workflow_dispatch is not defined, return nil

```
on:
    push:
```
Such cases pre returned non-nil

pkg/model/workflow.go
```diff
@@ -79,22 +80,42 @@ type WorkflowDispatch struct {
...
-		return nil
-	}
+	switch w.RawOn.Kind {
+	case yaml.ScalarNode:
+		var val string
+		err := w.RawOn.Decode(&val)
+		if err != nil {
+			log.Fatal(err)
+			return nil
+		}
```
wolfogre, line 90: Would you mind changing it to `decodeNode` here? It will be more friendly with Gitea since `log.Fatal(err)` will cause Gitea to shutdown. See also #1705

pkg/model/workflow.go
```diff
@@ -79,22 +80,42 @@ type WorkflowDispatch struct {
...
+		if val == "workflow_dispatch" {
+			return &WorkflowDispatch{}
+		}
+	case yaml.SequenceNode:
+		var val []string
+		err := w.RawOn.Decode(&val)
+		if err != nil {
+			log.Fatal(err)
+			return nil
+		}
```
wolfogre, line 100: Ditto.

pkg/model/workflow.go
pangliang: done

pkg/model/workflow.go
pangliang: done

wolfogre (approved):

wolfogre (approved):

cplee (approved):

pangliang:
I accidentally clicked the 'Re-request review' icon in the 'Reviewers' list of a pull-request page

cplee (approved):

PR #2128, opened 2023-12-16
feat: support offline mode

TKaxv-7S:
In some regions, connecting to github.com may not be as smooth.

So every time we using `uses:` method, it is not a good way to `fetch` and `pull` the current action content, and I think the action tag will also be rarely updated.

We can add a configurable option in `OfflineMode` to control it, setting the default value to `false`, so this will not affect other users.

PS: This feature is mainly intended for use by [Gitea act_runner](https://gitea.com/gitea/act_runner). If possible, I will also create a pull request over there.


wolfogre:
Can you provide a detailed explanation of how offline mode works with act runner?

BTW, if you cannot access github.com, you can handle it with `use: https://custom_domain/some/actions`, like `https://gitea.com/actions/checkout`, it's supported by act runner.

TKaxv-7S:
> Can you provide a detailed explanation of how offline mode works with act runner?
> 
> BTW, if you cannot access github.com, you can handle it with `use: https://custom_domain/some/actions`, like `https://gitea.com/actions/checkout`, it's supported by act runner.

Yes, I created a new pull request at [act_runner#454](https://gitea.com/gitea/act_runner/pulls/454)
This code is very limited.

TKaxv-7S:
> No progress here? You would need two approvals for getting this merged.
> 
> FYI My approval for this depends on the following
> 
> > In nektos act is forcepull enabled by default, I would expect that offline mode would also stop pulling existing images e.g. turning the default force-pull off.
> 
> Alternatively I think it might be reasonable that everywhere the forcepull is checked a extra check to OfflineMode is done to behave like forcepull is disabled if OfflineMode is active.
> 
> e.g.
> 
> ```go
> if forcepull {
> }
> if !forcepull {
> }
> ```
> 
> becomes
> 
> ```go
> if !OfflineMode && forcepull {
> }
> if OfflineMode || !forcepull {
> }
> ```
> 
> _I'm not requesting changes in the hard way, so that two other maintainer can still vote to get this merged as it is_
> 
> _I couldn't find the related issue about running act offline, but I believe it exists_

I agree with this suggestion. I have modified the code and [committed](https://github.com/nektos/act/pull/2128/commits/106275b1c0a96b1f7dc5489ebbf4400eec0de0b7) it here

TKaxv-7S:
@ChristopherHX Sorry for misunderstanding your meaning earlier. Now I have modified the code and [committed](https://github.com/nektos/act/compare/106275b1c0a96b1f7dc5489ebbf4400eec0de0b7..fe280974cf087bf1da74e1a9c10c5c60ba9f8a10) it here.

TKaxv-7S:
> Works perfectly fine for me on windows, thank you.
> 
> Can rerun act with offlinemode using it's cache without getting any error due to turning off internet.
> 
> `.\t\.github\workflows\t.yaml`
> 
> ```yaml
> on: push
> jobs:
>   _:
>     runs-on: t
>     steps:
>     - uses: actions/download-artifact@v4 # This actions obviously fails, but never due to no internet
> ```
> 
> via cli
> 
> ```
> .\act.exe -C .\t\ -P t=-self-hosted --action-offline-mode
> ```
> 
> Once the mode is enabled the following is the new behavior
> 
> * not yet cloned actions will be cloned in this mode, assumes internet is working
> * already cloned actions won't be updated (breaks updates of rolling tags / branches, but that's a user decision)
> * already pulled images won't be updated
> * works offline, with disabled internet
>   
>   * no rewriting of workflow files needed
> 
> One pending review, from maintainers, required...

Thank you for your detailed description of this feature, which makes the originally vague function clearer and easier to understand

cplee (approved):

PR #2140, opened 2023-12-28
fix: write default config in XDG config dir to avoid cluttering the HOME directory by default

mrnossiom:


cplee (approved):

mrnossiom:
With the logs being so large, I don't quite get what failed. Even looking at raw logs. Can you guide me ?

mrnossiom:
Regression fixed in #2195

PR #2155, opened 2024-01-18
feat: Test writing remote Actions locally

ChristopherHX:
# How to test
Usage: `./act --local-repository "https://github.com/test/test@v0=$PWD" -W test.yml -P self-hosted=-self-hosted -v`

Now also possible `./act --local-repository "test/test@v0=$PWD" -W test.yml -P self-hosted=-self-hosted -v`

```yaml
on: push
jobs:
  _:
    runs-on: self-hosted
    steps:
    - uses: test/test@v0
```

action.yml in cwd
```yaml
runs:
  using: composite
  steps:
  - run: echo Test
```

# Why?

Local actions cannot have pre steps and use a different syntax in the uses key


ethanwater (commented):
hey, you should probably squash any related commits. quality of life improvement so the devs don't have to navigate much.

cplee (approved):

PR #2160, opened 2024-01-19
fix: accept platforms name contained equal sigh

### LINKED ISSUE #2159: Error: Reject platforms name contained "="
### Bug report info

```plain text
act version:            0.2.57-4-gcd40f3f-dirty
GOOS:                   linux
GOARCH:                 amd64
NumCPU:                 16
Sockets found:
        /var/run/docker.sock
Config files:
Build info:
        Go version:            go1.21.6
        Main version:
        Main path:
        Main checksum:
        Build settings:
                -buildmode:           exe
                -compiler:            gc
                -ldflags:             -X main.version=0.2.57-4-gcd40f3f-dirty
                CGO_ENABLED:          1
                CGO_CFLAGS:
                CGO_CPPFLAGS:
                CGO_CXXFLAGS:
                CGO_LDFLAGS:
                GOARCH:               amd64
                GOOS:                 linux
                GOAMD64:              v1
Docker Engine:
        Engine version:        24.0.6
        Engine runtime:        runc
        Cgroup version:        1
        OS:                    Docker Desktop
        OS type:               linux
        OS version:
        OS arch:               x86_64
        OS kernel:             5.15.133.1-microsoft-standard-WSL2
        OS CPU:                16
        OS memory:             15374 MB
        Security options:
                name=seccomp,profile=unconfined
```

### Command used with act

```sh
act -P IMAGE=debian-12=nektos/act-environments-ubuntu:22.04
```

### Describe issue

If we attempt use custom runner label `IMAGE=debian-12` at self-hosted-runner, act failed and puts this error log.   
I guess that `-P` parameter parser is unsupported value containd multi "=" string.   
> [test workflow/demo] 🚧  Skipping unsupported platform -- Try running with `-P IMAGE=debian-12=...

### Link to GitHub repository

_No response_

### Workflow content

```yml
name: GitHubActions on GitHub Enterprise Server

on:
  workflow_dispatch:

jobs:
  demo:
    runs-on: [ IMAGE=debian-12 ]
    steps:
      - name:
        run: |
          echo "konnichi wa"
```

### Relevant log output

```sh
act -v
DEBU[0000] Conditional GET for notices etag=492cc8e5-303b-4af6-ab7f-7eddf714ee2f 
DEBU[0000] Loading environment from /work/.env
DEBU[0000] Loading action inputs from /work/.input
DEBU[0000] Loading secrets from /work/.secrets
DEBU[0000] Loading vars from /work/.vars
DEBU[0000] Evaluated matrix inclusions: map[]
DEBU[0000] Loading workflows from '/work/.github/workflows'
DEBU[0000] Loading workflows recursively
DEBU[0000] Found workflow 'main.yml' in '/work/.github/workflows/main.yml' 
DEBU[0000] Reading workflow '/work/.github/workflows/main.yml'
DEBU[0000] Preparing plan with all jobs
DEBU[0000] Using the only detected workflow event: workflow_dispatch
DEBU[0000] Planning jobs for event: workflow_dispatch
? Please choose the default image you want to use with act:
  - Large size image: ca. 17GB download + 53.1GB storage, you will need 75GB of free disk space, snapshots of GitHub Hosted Runners without snap and pulled docker images
  - Medium size image: ~500MB, includes only necessary tools to bootstrap actions and aims to be compatible with most actions
  - Micro size image: <200MB, contains only NodeJS required to bootstrap actions, doesn't work with all actions

Default image and other options can be changed manually in ~/.actrc (please refer to https://github.com/nektos/act#configuration for additional information about file structure)  [Use arrows to move, type to filter, ? for more help]
  Large
DEBU[0000] Saving notices etag=492cc8e5-303b-4af6-ab7f-7eddf714ee2f 
DEBU[0000] No new notices
```

### Additional information

_No response_

totechite:
Fix #2159
I hope to help act user using GitHub EnterPrise Server.

### DIFF
--- a/cmd/platforms.go
+++ b/cmd/platforms.go
@@ -14,9 +14,12 @@ func (i *Input) newPlatforms() map[string]string {
 
 	for _, p := range i.platforms {
 		pParts := strings.Split(p, "=")
-		if len(pParts) == 2 {
-			platforms[strings.ToLower(pParts[0])] = pParts[1]
-		}
+		// pop operation
+		pValue := pParts[len(pParts)-1]
+		pParts = pParts[:len(pParts)-1]
+
+		pName := strings.Join(pParts, "=")
+		platforms[strings.ToLower(pName)] = pValue
 	}
 	return platforms
 }


wolfogre (changes requested):
I have read #2159 carefully, however, I didn't see a strong reason to support `=` in the platform name. Could you please explain why it will help users using GitHub Enterprise Server? Why is `=` required? TBH, I think `IMAGE=debian-12=nektos/act-environments-ubuntu:22.04` is quite confusing.

totechite:
@wolfogre 
Thank you for checking 
As you may know, GHE users can't pick the GH Managed runner service so, we necessarily use self-hosted-runner when using GH Actions.

In my case, I use self-hosted-runners that are labeled the name `IMAGE=debian-12` at my organization. 
however, I can't use act in my workspace using GHE because of this issue. And I don't hope to modify the runner-label for using this tool.

runner-label 's specification  accepts `=` so if act support self-hosted-runner platform, may need this change. 

thanks.

wolfogre:
Sorry, I'm afraid I need to reserve my opinion. I don't see a strong reason to include `=` in the label name in your case.

For most command tools, `a=b=c` will be treated as `a: b=c`, like `docker run --env a=b=c --label a=b=c`. And act has many parameters with key-value format. I hope they could keep the same behavior to avoid confusion.

```bash
$ act -h | grep stringArray
      --container-cap-add stringArray                     kernel capabilities to add to the workflow containers (e.g. --container-cap-add SYS_PTRACE)
      --container-cap-drop stringArray                    kernel capabilities to remove from the workflow containers (e.g. --container-cap-drop SYS_PTRACE)
      --env stringArray                                   env to make available to actions with optional value (e.g. --env myenv=foo or --env myenv)
      --input stringArray                                 action input to make available to actions (e.g. --input myinput=foo)
      --matrix stringArray                                specify which matrix configuration to include (e.g. --matrix java:13
  -P, --platform stringArray                              custom image to use per platform (e.g. -P ubuntu-18.04=nektos/act-environments-ubuntu:18.04)
      --replace-ghe-action-with-github-com stringArray    If you are using GitHub Enterprise Server and allow specified actions from GitHub (github.com), you can set actions on this. (e.g. --replace-ghe-action-with-github-com =github/super-linter)
  -s, --secret stringArray                                secret to make available to actions with optional value (e.g. -s mysecret=foo or -s mysecret)
      --var stringArray                                   variable to make available to actions with optional value (e.g. --var myvar=foo or --var myvar)
```

totechite:
I agree with your thoughts, but how should I interpret the message here, which is also mentioned in #2159 ? 
It seems to accept `a=b: c`.
> [test workflow/demo] 🚧 Skipping unsupported platform -- Try running with `-P IMAGE=debian-12=...

I would like to avoid changing the runner-label name to use act if possible.
This is a so useful tool, but when using GH Actions, I don't hope should not be forced to use label names that take into account the specifications of this tool .

totechite:
How about modifying the content to accept the format of grouping the key string with any symbols. Would it look ugly?  `'a=b'=c`

totechite:
@wolfogre Please re-review.

cmd/platforms.go
```diff
@@ -14,9 +14,12 @@ func (i *Input) newPlatforms() map[string]string {
...
 		pParts := strings.Split(p, "=")
-		if len(pParts) == 2 {
-			platforms[strings.ToLower(pParts[0])] = pParts[1]
-		}
+		// pop operation
+		pValue := pParts[len(pParts)-1]
+		pParts = pParts[:len(pParts)-1]
+
+		pName := strings.Join(pParts, "=")
+		platforms[strings.ToLower(pName)] = pValue
```
allexiusw, line 22: Wondering if we can have tests for this? Not sure if there is a rule where you should add tests for each new feature.

allexiusw (commented):
Can we have some unittests for this feature?

wolfogre:
> @wolfogre Please re-review.

I take a neutral stance on this. Please ignore my change request and let other maintainers decide.

PR #2179, opened 2024-01-30
fix: EACCESS error regarding /opt/hostedtoolcache

AndesKrrrrrrrrrrr:
The `RUNNER_TOOL_CACHE` environment variable points to the hard-coded `/opt/hostedtoolcache`, which is not user-accessible.

### DIFF
--- a/pkg/common/env.go
+++ b/pkg/common/env.go
@@ -4,12 +4,12 @@ import (
 	"os"
 )
 
-var DefaultEnvironment = map[string][]string{
+var defaultEnvironment = map[string][]string{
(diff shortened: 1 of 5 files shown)


AndesKrrrrrrrrrrr:
Can elaborate more on the host vs. container `RUNNER_TOOL_CACHE`?

> Does act --env RUNNER_TOOL_CACHE=path not work for you?
It does, but that means I have to set it manually on every run. Here I'm trying to address it with sane defaults, like we do with `socketLocation()`. Also, I find it crazy to default (hard-coded) to a system-level directory.

AndesKrrrrrrrrrrr:
Closing, since this seems to be not an issue anymore according to your comment :smile:

PR #2180, opened 2024-01-30
chore: Socket logic consolidation

AndesKrrrrrrrrrrr:
Depends on #2179

Attempt at moving all logic regarding the docker socket location to a single place, to prevent duplication and human error when needing to edit in multiple places.


AndesKrrrrrrrrrrr:
I've lost context and interest. My problems have been fixed in #2181and #1494

PR #2185, opened 2024-01-31
fix: sync DOCKER_HOST with cli parameter

ChristopherHX:
- **new** mount unix socket from `DOCKER_HOST` if it uses that protocol to `/var/run/docker.sock`
  - **old** behavior was mount `/var/run/docker.sock` if no cli arg changed it
- **new** mount the auto detected unix socket if it uses the unix socket protocol to `/var/run/docker.sock` (grey list of known non default sockets)
  - **old** behavior was mount `/var/run/docker.sock` if no cli arg changed it


AndesKrrrrrrrrrrr:
This does, indeed, seem to work as advertised. A nice and simple change :heart: 

I still don't like the logic block; it's nested with plenty of double-negatives.

LGTM :+1:

AndesKrrrrrrrrrrr:
For future reference, the other PR is #2181

Issue #2135, opened 2023-12-19
Research about actions/upload-artifact@v4 and actions/download-artifact@v4


ChristopherHX:
Decompiled proto for artifacts from toolkit only typescript sources provided by GitHub.

[Source](https://github.com/actions/toolkit/blob/68f22927e727a60caff909aaaec1ab7267b39f75/packages/artifact/src/generated/results/api/v1/artifact.ts)

```proto
syntax = "proto3";

import "google/protobuf/timestamp.proto";
import "google/protobuf/wrappers.proto";

package github.actions.results.api.v1;

message CreateArtifactRequest {
    string workflow_run_backend_id = 1;
    string workflow_job_run_backend_id = 2;
    string name = 3;
    google.protobuf.Timestamp expires_at = 4;
    int32 version = 5;
}

message CreateArtifactResponse {
    bool ok = 1;
    string signed_upload_url = 2;
}

message FinalizeArtifactRequest {
    string workflow_run_backend_id = 1;
    string workflow_job_run_backend_id = 2;
    string name = 3;
    int64 size = 4;
    google.protobuf.StringValue hash = 5;
}

message FinalizeArtifactResponse {
  bool ok = 1;
  int64 artifact_id = 2;
}

message ListArtifactsRequest {
    string workflow_run_backend_id = 1;
    string workflow_job_run_backend_id = 2;
    google.protobuf.StringValue name_filter = 3;
    google.protobuf.Int64Value id_filter = 4;
}

message ListArtifactsResponse {
    repeated ListArtifactsResponse_MonolithArtifact artifacts = 1;
}

message ListArtifactsResponse_MonolithArtifact {
    string workflow_run_backend_id = 1;
    string workflow_job_run_backend_id = 2;
    int64 database_id = 3;
    string name = 4;
    int64 size = 5;
    google.protobuf.Timestamp created_at = 6;
}

message GetSignedArtifactURLRequest {
    string workflow_run_backend_id = 1;
    string workflow_job_run_backend_id = 2;
    string name = 3;
}

message GetSignedArtifactURLResponse {
    string signed_url = 1;
}

message DeleteArtifactRequest {
    string workflow_run_backend_id = 1;
    string workflow_job_run_backend_id = 2;
    string name = 3;
}

message DeleteArtifactResponse {
    bool ok = 1;
    int64 artifact_id = 2;
}
```

_Notice, this was decompiled by bing chatbot and only required include fixes_
[...]

PR #2224, opened 2024-02-19
Actions Artifacts v4 backend

ChristopherHX:
A pretty insecure backport with stubs and removed checks

I'm the original author of most code in artifacts_v4.go and the dumped proto code

jwt code is from the gitea project

Use at your own risk

If you want tests, someone else need to take this over.

Tests for the full Gitea Implementation can be found here: https://github.com/go-gitea/gitea/pull/28965

Fixes https://github.com/nektos/act/issues/2135

**dumped proto file has been deleted, because I'm not going to make changes to that file to satisfy the linter**


TWiStErRob:
Awesome! I can confirm this works (I grabbed the executable artifact from https://github.com/nektos/act/actions/runs/7959602137).
Will need to adjust some scripts because of the way artifact upload works, but looks as expected:

| Before | After |
| - | - |
| image | image image |

cplee (approved):

PR #2257, opened 2024-03-22
docs: use HTTPS for cloning instead of SSH

AbhinavMir:
Cloning is a read-only command, and using SSH is a bit of an overkill, and gives me an error when I am inside a server, with unconfigured git (which I imagine a lot of people do).


AbhinavMir:
NVM I forgot people had private repos.

PR #2265, opened 2024-03-28
Support overwriting caches

wolfogre:
In the old implementation, it assumes that there will be no duplicate uploads of caches with the same key.

But in fact, some actions like `actions/setup-go` will do something like:

1. Download the cache with the generated key.
2. Add more contents to the cache dir by downloading and building.
3. Upload the cache dir with the same key to replace the old cache.

It will result in a failure (though it will actually be ignored) with the following log:
```
/root/go/pkg/mod
/root/.cache/go-build
[command]/bin/tar --posix -cf cache.tgz --exclude cache.tgz -P -C /workspace/gitea/act_runner --files-from manifest.txt -z
::warning::Failed to save: {"error":"already exist"}
```

To support it, this PR introduces some changes.

- Uploading caches with same key (and the version which indicates the env) will be allowed.
- Return the latest completed one when retrieving a cache with a key.
- The old caches with the same key will be deleted regularly, keep the latest only.

I have tested the code with a custom act_runner. It works well and is compatible with old data.

### DIFF
--- a/pkg/artifactcache/handler.go
+++ b/pkg/artifactcache/handler.go
@@ -206,22 +206,12 @@ func (h *Handler) reserve(w http.ResponseWriter, r *http.Request, _ httprouter.P
 	api.Key = strings.ToLower(api.Key)
 
 	cache := api.ToCache()
-	cache.FillKeyVersionHash()
 	db, err := h.openDB()
 	if err != nil {
 		h.responseJSON(w, r, 500, err)
 		return
 	}
 	defer db.Close()
-	if err := db.FindOne(cache, bolthold.Where("KeyVersionHash").Eq(cache.KeyVersionHash)); err != nil {
-		if !errors.Is(err, bolthold.ErrNotFound) {
-			h.responseJSON(w, r, 500, err)
-			return
-		}
-	} else {
-		h.responseJSON(w, r, 400, fmt.Errorf("already exist"))
-		return
-	}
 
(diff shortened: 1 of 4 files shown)


wolfogre:
Weird, I don't think the failed CI is related now.

image

wolfogre:
> > FAIL: TestRunEvent/networking
> 
> Should we remove that flaky test? I mean it caused a lot of reruns in the past

@ChristopherHX Thank you for pointing out the problem. I was lost in the huge logs.

How about #2266.

I have tested the commit in this PR, and it works, but I will revert it since it is unrelated.

wolfogre:
> > Upload the cache dir with the same key to replace the old cache
> 
> Yes this is probably true in act, since the run_number, run_id, run_attempt doesn't update automatically.
> 
> I'm not yet shure that this is the same Key on GitHub Actions Servers, but they could have silently implemented the override cache feature request.
> 

@ChristopherHX I am sure GitHub Actions support multiple caches with the same key.

Let me show you the [cache manage page](https://github.com/go-gitea/gitea/actions/caches) of Gitea

image

wolfogre:
> > @ChristopherHX I am sure GitHub Actions support multiple caches with the same key
> 
> Ah yes that are cache scopes that are not implemented here.
> 
> They are unique per repo + ref only default branch scope

Actually I have already implemented the cache scope, maybe port it later after this PR.

PR #2284, opened 2024-04-09
Remove local action cache if remote has changed

wolfogre:
Remove the local action cache if the remote has changed.

This typically does not happen until the user sets a different `GITHUB_SERVER_URL`.


wolfogre:
> @wolfogre this pull request has failed checks 🛠

The failed CI seems unrelated.

image

PR #2288, opened 2024-04-10
Fix codecov in workflow

wolfogre:


wolfogre:
> afaik this only works for us with write access, I assume pr's from forks still see an error

I don't think so, see the discussion.

wolfogre:
Although I have certain write permissions, this PR still comes from a forked PR, so secrets will not be injected.

PR #2296, opened 2024-04-16
Add codespell support (config, workflow to detect/not fix) and make it fix few typos

yarikoptic:
More about codespell: https://github.com/codespell-project/codespell .

I personally introduced it to dozens if not hundreds of projects already and so far only positive feedback.

CI workflow has 'permissions' set only to 'read' so also should be safe.

### DIFF
--- a/.codespellrc
+++ b/.codespellrc
@@ -1,0 +1,6 @@
+[codespell]
+# Ref: https://github.com/codespell-project/codespell#using-a-config-file
+skip = .git*,go.sum,package-lock.json,*.min.*,.codespellrc,testdata
+check-hidden = true
+ignore-regex = .*Te\{0\}st.*
+# ignore-words-list =
--- a/.codespellrc
+++ b/.codespellrc
@@ -1,6 +1,6 @@
 [codespell]
 # Ref: https://github.com/codespell-project/codespell#using-a-config-file
-skip = .git*,go.sum,package-lock.json,*.min.*,.codespellrc,testdata
+skip = .git*,go.sum,package-lock.json,*.min.*,.codespellrc,testdata,./pkg/runner/hashfiles/index.js
 check-hidden = true
 ignore-regex = .*Te\{0\}st.*
 # ignore-words-list =
--- a/.github/workflows/codespell.yml
+++ b/.github/workflows/codespell.yml
@@ -1,0 +1,23 @@
+# Codespell configuration is within .codespellrc
+---
+name: Codespell
+
+on:
+  push:
+    branches: [master]
+  pull_request:
+    branches: [master]
+
+permissions:
+  contents: read
+
+jobs:
+  codespell:
+    name: Check for spelling errors
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Codespell
+        uses: codespell-project/actions-codespell@v2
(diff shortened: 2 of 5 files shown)


wolfogre:
Hi Yaroslav, I remember I had refused your PR like this to Gitea because it touched too many files (50+) involving docs, js, go files and template files, making it almost impossible to review. Sorry again if it made you unhappy.

However, for this PR, I don't see any reason not to go, all changes are all reasonable and reviewable. 👍

pkg/runner/hashfiles/index.js
wolfogre, line 1: I am sure if we should touch `index.js` since I remember it's copied from `actions/runner`.

@ChristopherHX I think you should take a look.

wolfogre (dismissed):

yarikoptic:
Thank you @wolfogre ! FWIW -- we can exclude any file you like from being checked/fixed by codespell.

my 1c: Number of files with typos is not an indicator of deficiency in the goal of having typo free code base ;-)  typos are usually easy to spot and relatively easy to review IMHO since it is quite often a typo or two in a file.  GitHub also provides means to mark reviewed files so they collapse.  So it becomes a matter of a few quality minutes of time to review and then live with (much more so) typo free codebase . But indeed a number of files quite often scary people away, but I can find some brave projects which made the dive and so far I have not heard much of complaints.

pkg/runner/hashfiles/index.js
yarikoptic: cool, excluded

<details>
<summary>here is the force pushed diff</summary> 

```shell
❯ git diff gh-yarikoptic/enh-codespell..
diff --git a/.codespellrc b/.codespellrc
index fe7ef1c..4857dfe 100644
--- a/.codespellrc
+++ b/.codespellrc
@@ -1,6 +1,6 @@
 [codespell]
 # Ref: https://github.com/codespell-project/codespell#using-a-config-file
-skip = .git*,go.sum,package-lock.json,*.min.*,.codespellrc,testdata
+skip = .git*,go.sum,package-lock.json,*.min.*,.codespellrc,testdata,./pkg/runner/hashfiles/index.js
 check-hidden = true
 ignore-regex = .*Te\{0\}st.*
 # ignore-words-list =
diff --git a/pkg/runner/hashfiles/index.js b/pkg/runner/hashfiles/index.js
index 37e21db..f2bed37 100644
--- a/pkg/runner/hashfiles/index.js
+++ b/pkg/runner/hashfiles/index.js
@@ -942,10 +942,10 @@ class Summary {
         return this.addRaw(element).addEOL();
     }
     /**
-     * Adds a collapsible HTML details element to the summary buffer
+     * Adds a collapsable HTML details element to the summary buffer
      *
      * @param {string} label text for the closed state
-     * @param {string} content collapsible content
+     * @param {string} content collapsable content
      *
      * @returns {Summary} summary instance
      */
@@ -1844,7 +1844,7 @@ class Pattern {
             // false for `/foo` but returns true for `/foo/`. Append a trailing slash to handle that quirk.
             if (!itemPath.endsWith(path.sep)) {
                 // Note, this is safe because the constructor ensures the pattern has an absolute root.
-                // For example, formats like C: and C:foo on Windows are resolved to an absolute root.
+                // For example, formats like C: and C:foo on Windows are resolved to an aboslute root.
                 itemPath = `${itemPath}${path.sep}`;
             }
         }
```
</details>

wolfogre (approved):

PR #2301, opened 2024-04-21
chore: use errors.New to replace fmt.Errorf with no parameters will much better

ChengenH:
use errors.New to replace fmt.Errorf with no parameters will much better


wolfogre:
No, I don't see it will be "much better" when `fmt.Errorf` with no parameters is the same as `errors.New`.

image

Code is changing as long as the project is being maintained, and it will be annoyed once the maintainers wants to add new stuff to `errors.New("abc")`, they have to import `fmt` and change the function to `fmt.Errorf`, while `fmt.Errorf("abc")` -> `fmt.Errorf("abc %v: %w", v, err)` is much easier.

PR #2312, opened 2024-05-01
feat: worklow: add `permissions` field

Frankkkkk:
The `permissions` field can either be a top level key, or a per-job one. It supports a map of scope<>permission relations.
For example, to allow the jobs to write to the the packages registry, the following must be set:
```yaml
permissions:
  packages: write
```

this fixes that.
Right now, this field is not used/useful when running the actions locally, but can be used to handle permissions on forges that implement the same action format, such as Forgejo/Gitea.


Frankkkkk:
Good catch, indeed! What do you propose?

PR #2352, opened 2024-06-05
feat: generate manual page automatically with cobra/doc

ittner:
Add a command line switch "--man-page" to print an automatically generated manual page to stdout.


ittner:
> ...I'm still undecided about this feature

Hi @ChristopherHX !

A bit of rationale for this feature: I'm looking into packaging act for Debian and Ubuntu and manpages are required by both distro policies. I can imagine other distros having similar requirements.

Of course, manpages can be written manually (at the risk of leaving them outdated as the CLI changes) or I can keep a local patch use it to generate the pages when needed, but this code seems more useful if it is shared among potential users.  Hugo and others uses a similar approach for generating man-pages, for example.

I though about having it optionally write the output to a file but I fear this can introduce more complexity that needed for potential users, as they will need to remember or look into another options instead of just redirecting with the shell... but perhaps my fear is undeserved and the output to a file is welcome.

ittner:
Branch rebased to fix new conflict on go.sum/mod.

wolfogre (approved):

PR #2372, opened 2024-06-21
fix: make node tool non volatile

ChristopherHX:
Currently downgrading node via setup-node can break later actions


deltreey:
Love it!  Thank you for doing this!

cplee (approved):

PR #2373, opened 2024-06-21
fix: changed location of actrc in info message

just-dev-creator:
Hi there,
when trying to switch the docker image used for one of my projects, I noticed that the .actrc file is not in the path stated during setup (~/.actrc), but rather in ~/.config/act/actrc. I changed the info text accordingly. 
I hope this isn't some special thing when using act via the AUR or something like that, but at least for me it took some minutes to find the correct config. 
If this issue is specific to me, please just close this PR - I certainly don't have enough knowledge of Go to find the code part that creates the config and investigate further.


just-dev-creator:
Perfect, I will try to copy the functionality of the bug-report command to this info message.

PR #2416, opened 2024-08-04
feat: Validate GitHub Actions schema

### LINKED ISSUE #2414: act should validate schema in a way that is close to what github does
### Act version

act version 0.2.64

### Feature description

act should run the workflow yaml through a schema validation pass and catch schema-level issues early.

## My Example

I made an error in my workflow and had:

```yml
jobs:
  my-job:
    contents: write
```

I was missing a parent `permissions.contents` node. 

**Actual behavior** act ran the workflow whereas github immediately showed an error:

```The workflow is not valid. .github/workflows/example.yml (Line: 12, Col: 5): Unexpected value 'contents'```

**Desired behavior** should be able to get similar error from `act` without running the workflow.

ChristopherHX:
Resolves #2414 

**BREAKING** previously accepted workflows are now invalid

@wolfogre you would need to add context gitea everywhere github is in the context array in your fork if this would be merged

I'm not shure if Gitea Actions has labeled some behavior allowed without schema as features.

Moreover Singleworlflows are more likely to be rejected after this change.

TBD add tests that serialized workflows from model can be parsed again, e.g. the Container structure used to contain additional fields that are invalid per schema

As draft to see which test is invalid as well


wolfogre:
Thanks for mentioning me. I'll keep an eye on this PR. Don't worry, although there could be some challenges for the Gitea's fork, I believe we can find a way to handle them. So please feel free to do what you think is right. 👍

weyert:
This looks really interesting! I have had cases were I was happy writing Github Actions/workflows with `act` and then Github Actions CI on GitHub.com complained about things. Would be really nice to have early warnings about such errors.

wolfogre (approved):

PR #2434, opened 2024-08-24
feat: tart macOS vm's as job container

ChristopherHX:
adds the tart:// protocol to platform mapping

e.g. `-P macos-14=tart://ghcr.io/cirruslabs/macos-sonoma-base:latest` if you have a mac.

I have written the code in a payed dedicated host machine that is now deallocated, expenses satisfied by sponsoring of my other act based project while I don't see anything here, as I don't own a mac.

`add-path` is probably broken, but I ignore the fact as of now.

### DIFF
--- a/pkg/runner/runner_test.go
+++ b/pkg/runner/runner_test.go
@@ -375,7 +375,7 @@ func TestTartNotSupportedOnNonDarwin(t *testing.T) {
 
 		tables = append(tables, []TestJobFileInfo{
 			// Shells
-			{workdir, "shells/sh", "push", "tart not supported", platforms, secrets},
+			{workdir, "basic", "push", "tart not supported", platforms, secrets},
 		}...)
 	}
 


YOU54F:
Just starting to test this out and it looks good on first pass. testing on a mac m1 mbp

if the gh action step fails, the container is left in a running state. 

if the container fails to start (say 2 vm limit is reached), the container is left in a running state.

I've been using ssh, with tart / parallels / qemu to run jobs with `-P ${PLATFORM}=-self-hosted` on windows/macos/linux vms for the last year (https://gist.github.com/YOU54F/3ac099e54e48a31a69ac2d671aa878f6 / https://gist.github.com/YOU54F/789c53ae8380a1f90325952615635f8a )

I had a play with your runner.server & github-act-runner to build gh actions in my homelab with a set of vm's on a few diff machines. Cheers for your work

pkg/runner/expression.go
```diff
@@ -512,7 +512,7 @@ func getEvaluatorInputs(ctx context.Context, rc *RunContext, step step, ghc *mod
 			for k, v := range config.Inputs {
 				value := nestedMapLookup(ghc.Event, "inputs", k)
 				if value == nil {
-					v.Default.Decode(&value)
+					_ = v.Default.Decode(&value)
```
allexiusw, line 515: Any special reason to use _? Instead of keeping `v.Default.Decode(&value)` we are not using `_` anyways? Just curious...

pkg/runner/run_context_darwin.go
```diff
@@ -0,0 +1,111 @@
...
+				rawLogger.Debugf("%s", s)
+			}
+			return true
+		})
+		cacheDir := rc.ActionCacheDir()
+		randBytes := make([]byte, 8)
+		_, _ = rand.Read(randBytes)
+		miscpath := filepath.Join(cacheDir, hex.EncodeToString(randBytes))
+		actPath := filepath.Join(miscpath, "act")
+		if err := os.MkdirAll(actPath, 0o777); err != nil {
```
allexiusw, line 35: Should ` 0o777` be added as a constant?

pkg/runner/run_context_other.go
```diff
@@ -0,0 +1,16 @@
...
+import (
+	"context"
+	"fmt"
+
+	"github.com/nektos/act/pkg/common"
+)
+
+func (rc *RunContext) startTartEnvironment() common.Executor {
+	return func(_ context.Context) error {
+		return fmt.Errorf("You need macOS for tart")
```
allexiusw, line 14:
```suggestion
		return fmt.Errorf("You need macOS for start")
```
Is it a typo?

allexiusw: Figured out it isn't a typo.

allexiusw (commented):
I'm seeing not test coverage, do we require them in a PR in order to get merged in this repo?

pkg/runner/expression.go
allexiusw: Ah gotcha! yeah with that convention the `_` has a lot of sense semantically. Thanks a lot for explaining this to my.

PR #2494, opened 2024-10-19
Remove else statements from if/return blocks.

stephenwithav:
Also, sort runner.Config, model.Githubcontext fields to be clearer.


stephenwithav:
@ChristopherHX Would a PR to your fork earn your approval?  What changes could be made to this PR?

wolfogre (changes requested):
> sort runner.Config, model.Githubcontext fields

Please don't do this. Perhaps the current order does need to be adjusted indeed, but sorting the fields in lexicographical order is a bad idea. It will scatter the related fields everywhere, making the code extremely difficult to read.

stephenwithav:
> Please don't do this. Perhaps the current order does need to be adjusted indeed, but sorting the fields in lexicographical order is a bad idea. It will scatter the related fields everywhere, making the code extremely difficult to read.

Would grouping the fields and sorting the grouped fields be an acceptable compromise?

wolfogre:
> > Please don't do this. Perhaps the current order does need to be adjusted indeed, but sorting the fields in lexicographical order is a bad idea. It will scatter the related fields everywhere, making the code extremely difficult to read.
> 
> Would grouping the fields and sorting the grouped fields be an acceptable compromise?

I am wondering whether it makes the code more readable for developers. IMO, lexicographical order does not help.

For example:

```go
type URL struct {
	Scheme      string
	Opaque      string    // encoded opaque data
	User        *Userinfo // username and password information
	Host        string    // host or host:port (see Hostname and Port methods)
	Path        string    // path (relative paths may omit leading slash)
	RawPath     string    // encoded path hint (see EscapedPath method)
	OmitHost    bool      // do not emit empty host (authority)
	ForceQuery  bool      // append a query ('?') even if RawQuery is empty
	RawQuery    string    // encoded query values, without '?'
	Fragment    string    // fragment for references, without '#'
	RawFragment string    // encoded fragment hint (see EscapedFragment method)
}

// vs

type URL struct {
	ForceQuery  bool      // append a query ('?') even if RawQuery is empty
	Fragment    string    // fragment for references, without '#'
	Host        string    // host or host:port (see Hostname and Port methods)
	OmitHost    bool      // do not emit empty host (authority)
	Opaque      string    // encoded opaque data
	Path        string    // path (relative paths may omit leading slash)
	RawFragment string    // encoded fragment hint (see EscapedFragment method)
	RawPath     string    // encoded path hint (see EscapedPath method)
	RawQuery    string    // encoded query values, without '?'
	Scheme      string
	User        *Userinfo // username and password information
}
```

stephenwithav:
> I am wondering whether it makes the code more readable for developers. IMO, lexicographical order does not help.

The `net.URL` example is interesting.  I would counter by saying that `net.URL` is glanceable whereas `model.GithubContext` and `runner.Config` are not.  It would be nice if those could be deconstructed into smaller structs and then composed into the original struct.

stephenwithav:
> I don't currently think...
> 
> > deconstructed into smaller structs
> 
> ...is a good idea, since this somewhat breaks api or intellisense of top level fields.
> 
> I sometimes find myself to rewrite struct initialization when doing this, eventually due to linter while it actually should work regardless.

Interesting.  I tested it just now and never realized LSP wouldn't autocomplete embedded structs.  Thanks!

```go
package main

type Name struct {
	First string
	Last  string
}

type EmbeddedName struct {
	Name
}

func main() {
	embeddedName := EmbeddedName{
		First: "Steven",
		Last:  "Edwards",
	}

	name := Name{
		First: "Steven",
		Last:  "Edwards",
	}
}
```

stephenwithav:
Will this be accepted if I remove the sorted fields?

stephenwithav:
> from my perspective this would increase a chance of a merge by 400%, but let's hear wolfogres opinion
> 
> _In the past 2021/22 such kind of PR would have been rejected, because it doesn't really improve act by itself_

Understood.  My goal here was just to make the code more readable while I get used to the code base.

wolfogre:
> from my perspective this would increase a chance of a merge by 400%, but let's hear wolfogres opinion

Yeah, 400%. 😄

stephenwithav:
The revert is complete along with a cleanup of findGitSlug if you're interested.

pkg/common/git/git.go
```diff
@@ -198,21 +198,34 @@ func findGitRemoteURL(_ context.Context, file, remoteName string) (string, error
...
-		if matches := gheHTTPRegex.FindStringSubmatch(url); matches != nil {
-			return "GitHubEnterprise", fmt.Sprintf("%s/%s", matches[1], matches[2]), nil
-		} else if matches := gheSSHRegex.FindStringSubmatch(url); matches != nil {
+	}
+
+	if githubInstance != "github.com" {
+		if matches := matchesRegex(url,
+			regexp.MustCompile(fmt.Sprintf(`^https?://%s/(.+)/(.+?)(?:.git)?$`, githubInstance)),
+			regexp.MustCompile(fmt.Sprintf(`%s[:/](.+)/(.+?)(?:.git)?$`, githubInstance)),
+		); matches != nil {
```
allexiusw, line 228:
```suggestion
		if matches := matchesRegex(url,
			regexp.MustCompile(fmt.Sprintf(`^https?://%s/(.+)/(.+?)(?:.git)?$`, githubInstance)),
			regexp.MustCompile(fmt.Sprintf(`%s[:/](.+)/(.+?)(?:.git)?$`, githubInstance)),
		);
		matches != nil {
```

allexiusw, line 228:
```suggestion
		if matches := matchesRegex(url,
			regexp.MustCompile(fmt.Sprintf(`^https?://%s/(.+)/(.+?)(?:.git)?$`, githubInstance)),
			regexp.MustCompile(fmt.Sprintf(`%s[:/](.+)/(.+?)(?:.git)?$`, githubInstance)),
		);
		matches != nil {
```

pkg/common/git/git.go
```diff
@@ -198,21 +198,34 @@ func findGitRemoteURL(_ context.Context, file, remoteName string) (string, error
 	return remote.Config().URLs[0], nil
 }
 
+type findStringSubmatcher interface {
+	FindStringSubmatch(string) []string
+}
+
+func matchesRegex(url string, matchers ...findStringSubmatcher) []string {
```
allexiusw, line 205: Can you add unit tests for this function? That would be great.

PR #2510, opened 2024-11-01
fix: return an error when the argument count is wrong

earl-warren:


earl-warren:
> @earl-warren Is this coming from a fork of act? like forgeo~~/gitea~~ those lack behind updates
> 
> The act/schema package aborts already in this case as of August 2024.

Yes, sorry about that.

PR #2546, opened 2024-11-26
Add NoProfile flag to PowerShell/Pwsh Call

### LINKED ISSUE #2545: PowerShell steps load profile during execution
## Expected Behavior
Running a PowerShell script or command should run in it's own process without a profile being loaded.

## Current Behavior
Currently the invocation runs `powershell.exe -command . {0}` where the commands are replaced in.

## Possible Solution
Append `-NoProfile` to the powershell/pwsh calls.`

## Steps to Reproduce
1. On the instance running the command create a profile.
   - See https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_profiles?view=powershell-7.4#how-to-create-a-profile
   - Add something like `throw 'An error in your profile'`
2. Create a github workflow that runs powershell.

```yaml filename="helloworld.yaml"
on:
  push:
jobs:
  test:
    runs-on: windows-latest
  steps:
    - run: |
      Write-Host "Hello world"
 ```
 
3. Run act on the workflow while using `-self-hosted`
   - `act --platform windows-latest=-self-hosted --workflows "./.github/workflows/helloworld.yaml"`
5. See error

## Context
We are currently attach with `-P windows=self-host` on a windows machine.

### PR DESCRIPTION
**Pull request by HeyItsGilbert**
This is a known best practice when launching build commands. While this is not necessary for a container, it does issues when the platform is `-self-hosted`.

Fixes #2545

### DIFF
--- a/pkg/model/workflow.go
+++ b/pkg/model/workflow.go
@@ -625,7 +625,7 @@ func (s *Step) ShellCommand() string {
 			shellCommand = "bash --noprofile --norc -e -o pipefail {0}"
 		}
 	case "pwsh":
-		shellCommand = "pwsh -command . '{0}'"
+		shellCommand = "pwsh -NoProfile -command . '{0}'"
 	case "python":
 		shellCommand = "python {0}"
 	case "sh":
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by tablackburn** [comment]
The best workaround I've found to this so far is to conditionally run profile logic in interactive sessions only, like this:

```powershell
#region Set the PSReadLine prediction source to 'History' if the host is interactive
if ([System.Environment]::UserInteractive -and
    (Get-PSReadLineOption).PredictionSource -ne 'History' -and
    $env:ACT -ne $True) {
    Set-PSReadLineOption -PredictionSource 'History'
}
#endregion
```

PR #2552, opened 2024-12-01
[BREAKING] refactor logger

ChristopherHX:
**Set up job and Complete job are now steps in json logs**

**BREAKING this can break my own embedded use cases**

@SanjulaGanepola I have no plan for this PR to land before 6 pending PR's including breaking changes, since this has huge impact for github-act-runner or even act_runner of gitea.


SanjulaGanepola (commented):
I tested the changes with the extension and they work. I see we now have `stepResult` for setup and complete job steps. Do you think this PR can contain the other 2 items mentioned in the linked issue:
- [ ] `stepResult` for pre stage
- [ ] Message for skipped jobs/steps

SanjulaGanepola:
> > * `stepResult` for pre stage
> 
> Maybe debug this?, technically you should have this in the log I mean
> 
> https://github.com/nektos/act/blob/9c7f103bb39f11430f58c1a68d5629a42e6d56a1/pkg/runner/step.go#L64
> 
> should be called for all stages and has the code for stepResult.
> I noticed some log messages are not scoped in via this function in pre, like downloading the action aka (pre pre step, yes this get's weird, and it is)

@ChristopherHX I had a try with the following workflow:
```yaml
name: Sample

on:
  push:
    branches: ["main"]

jobs:
  cli:
    name: CLI

    runs-on: ubuntu-latest

    steps:
      - name: Install Node
        uses: actions/setup-node@v4
        with:
          node-version: 18.x
          registry-url: 'https://registry.npmjs.org'
```
This has both a `pre` and `post` stage. I have attached the full json log below. There is a `stepResult` only for the step itself and the `post` stage. `pre` stage for some reason does not have it. In act, it would be best to always have a `stepResult` for every stage to ensure we can properly get back the status of it rather than having to assume it was successful.

[output.log](https://github.com/user-attachments/files/18057419/output.log)

SanjulaGanepola:
> No this nodejs action has no pre step, the logger might make you believe this is the case.

> Downloading the action is needed to check if it has a pre step, those log messages use the same log fields as the pre step log messages.

At the moment, is there a way to properly determine when there is a real pre step?

I feel that it would make more sense for there to be a different log type to avoid this issue. What do you think? And can this be implemented easily?

SanjulaGanepola:
> moving downloading actions this into the setup job step makes more sense for me to align with actions/runner

I agree, this makes the most sense. `pre` step logs should strictly be for if the step truly does have a `pre` step.

SanjulaGanepola:
> > moving downloading actions this into the setup job step makes more sense for me to align with actions/runner
>
> I agree, this makes the most sense. pre step logs should strictly be for if the step truly does have a pre step.

@ChristopherHX Other than what we discussed above which still needs to be done^, curious what the status of this PR is.

SanjulaGanepola:
@ChristopherHX Yes, please invite me to actions-oss.

@cplee Could we get this change merged and released?

SanjulaGanepola (approved):

SanjulaGanepola:
@cplee This is another one of the PRs I mentioned I was hoping you could review and merge in.

PR #2590, opened 2024-12-25
Fix typo in `--action-offline-mode` option description

sobolevn:
It was like this:

```
      --action-offline-mode                               If action contents exists, it will not be fetch and pull again. If turn on this,will turn off force pull
```


cplee (approved):

sobolevn:
This failure does not look related:
Снимок экрана 2024-12-26 в 01 27 34

PR #2610, opened 2025-01-11
Log docker pull failure to the raw_output logger so it reaches gitea

### LINKED ISSUE #2609: Act step using docker silently fails when the image access is forbidden
### Bug report info

```plain text
Error: unknown flag: --bug-report

but:


./act_runner-0.2.11-linux-amd64 --version
act_runner version v0.2.11
```
```


### Command used with act

```sh
./act_runner daemon --config /etc/act_runner/config.yaml
```

### Describe issue

When docker is correctly logged in to a registry, but the user does not have permission to access the container it is trying to pull, `gitea act_runner` silently fails the step without providing any feedback on the problem.

Upon a lot of debugging, I reached the error message:

```
Failed to pull docker image registry.gitlab.com/path_to_the_image_I_want/image_name:hamster: Error response from daemon: Head "https://registry.gitlab.com/v2/path_to_the_image_I_want/image_name/manifests/hamster": denied: access forbidden
```

The problem is not that access is forbidden, as expected, but that failure is completely silent.

Apologies if this arrives from `gitea`, but after reviewing the code I see that this issue happens inside `act` itself, as that's the one that is handling the step.

### Link to GitHub repository

_No response_

### Workflow content

```yml
jobs:
  monolith:
    runs-on: ubuntu-latest
    steps:
    - name: Login to GitLab Registry
      uses: docker/login-action@v3
      with:
        registry: registry.gitlab.com
        username: ${{ secrets.GITLAB_LOGIN }}
        password: ${{ secrets.GITLAB_TOKEN }}
    - uses: actions/checkout@v4
    - name: Do something
      uses: docker://registry.gitlab.com/path_to_the_image_I_want/image_name:hamster
      with:
        args: echo "We never get to see this"
```

### Relevant log output

```sh
Jan 10 20:24:26 dev-01 act_runner[747]: [testing/monolith] [DEBUG]   🐳  docker pull registry.gitlab.com/path_to_the_image_I_want/image_name:hamster
Jan 10 20:24:27 dev-01 act_runner[747]: [testing/workflow]   ❌  Failure - Do something
Jan 10 20:24:27 dev-01 act_runner[747]: [testing/workflow] [DEBUG] expression '${{ github.workspace }}/.cache' rewritten to 'format('{0}/.cache', github.workspace)'
```

### Additional information

_No response_

pcarranza:
Fixes #2609

### DIFF
--- a/pkg/container/docker_pull.go
+++ b/pkg/container/docker_pull.go
@@ -70,6 +70,11 @@ func NewDockerPullExecutor(input NewDockerPullExecutorInput) common.Executor {
 			}
 			return err
 		}
+
+		// Log the error to the task output, otherwise it fails silently and makes everyone sad
+		// rawOutput is how to reach either gitea runner output, or the terminal output itself
+		common.Logger(ctx).WithField("raw_output", true).Errorf("Failed to pull docker image %s: %s", imageRef, err)
+
 		return nil
 	}
 }


Scutua (approved):

pcarranza:
I think that it's best to close this one in favor of https://github.com/nektos/act/pull/2631

As it was correctly pointed out by @ChristopherHX this should be fixed in Gitea - the only change I introduced was increase the logging level when debug is enabled by setting up SetReportCaller and pushing it to the job logger so I can understand where the logging should be happening.

In the end, the fact that this is not percolated in gitea seems to be a specific failure case of act_runner and not act.

Apologies for the noise.

Issue #2648, opened 2025-02-07
Matrix include error


silky1313:
### Bug report info

```plain text
act version:            0.2.71
GOOS:                   linux
GOARCH:                 amd64
NumCPU:                 8
Docker host:            DOCKER_HOST environment variable is not set
Sockets found:
        /var/run/docker.sock
Config files:           
        /home/silky/.config/act/actrc:
                -P ubuntu-latest=catthehacker/ubuntu:act-latest
                -P ubuntu-22.04=catthehacker/ubuntu:act-22.04
                -P ubuntu-20.04=catthehacker/ubuntu:act-20.04
                -P ubuntu-18.04=catthehacker/ubuntu:act-18.04
Build info:
        Go version:            go1.23.5
        Module path:           github.com/nektos/act
        Main version:          (devel)
        Main path:             github.com/nektos/act
        Main checksum:         
        Build settings:
                -buildmode:           exe
                -compiler:            gc
                CGO_ENABLED:          0
                GOARCH:               amd64
                GOOS:                 linux
                GOAMD64:              v1
                vcs:                  git
                vcs.revision:         a9e31cde304fb989e59068c6b1cb2f04a5fa5931
                vcs.time:             2025-01-13T02:27:17Z
                vcs.modified:         true
Docker Engine:
        Engine version:        27.4.1
        Engine runtime:        runc
        Cgroup version:        2
        Cgroup driver:         systemd
        Storage driver:        overlay2
        Registry URI:          https://index.docker.io/v1/
        OS:                    Ubuntu 22.04.5 LTS
        OS type:               linux
        OS version:            22.04
        OS arch:               x86_64
        OS kernel:             6.8.0-52-generic
        OS CPU:                8
        OS memory:             15946 MB
        Security options:
                name=apparmor
                name=seccomp,profile=builtin
                name=cgroupns
```

### Command used with act

```sh
act --action-offline-mode --verbose push
```

### Describe issue

# My own example
When I run the following workflow
```yaml
name: test matrix
on:
  push:

jobs:
  echo-color:
    runs-on: ubuntu-latest
    strategy:
      matrix:
        os: [macos-latest, windows-latest]
        version: [12, 14]
        include:
          - color: green
    steps:
      - name: echo color
        run: |
          echo 'os is ${{ matrix.os }}, version is ${{ matrix.version }}, color is ${{matrix.color}}'
```

The following is a relevant log of act
```
[...]

PR #2649, opened 2025-02-07
fix: job's GetMatrixes cannot get the correct matrix

silky1313:
#2648 
This Pull requests is mainly to solve this issue

### DIFF
--- a/pkg/model/workflow.go
+++ b/pkg/model/workflow.go
@@ -384,42 +384,6 @@ func (j *Job) GetMatrixes() ([]map[string]interface{}, error) {
 		j.Strategy.MaxParallel = j.Strategy.GetMaxParallel()
 
 		if m := j.Matrix(); m != nil {
-			includes := make([]map[string]interface{}, 0)
-			extraIncludes := make([]map[string]interface{}, 0)
-			for _, v := range m["include"] {
-				switch t := v.(type) {
-				case []interface{}:
-					for _, i := range t {
-						i := i.(map[string]interface{})
-						extraInclude := true
-						for k := range i {
-							if _, ok := m[k]; ok {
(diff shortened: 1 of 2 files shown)


silky1313:
@ChristopherHX ， Thank you for reviewing the PR I submitted. I think the problem is that the matrix is ​​empty when merging the include.

silky1313:
```
workflow_test.go:344: 
          	Error Trace:	/home/runner/work/act/act/pkg/model/workflow_test.go:344
          	Error:      	Not equal: 
          	            	expected: []map[string]interface {}{map[string]interface {}{"animal":"cat", "color":"green", "fruit":"apple", "shape":"circle"}, map[string]interface {}{"animal":"cat", "color":"green", "fruit":"pear"}, map[string]interface {}{"animal":"dog", "color":"green", "fruit":"apple", "shape":"circle"}, map[string]interface {}{"animal":"dog", "color":"green", "fruit":"pear"}, map[string]interface {}{"fruit":"banana"}, map[string]interface {}{"animal":"cat", "fruit":"banana"}}
          	            	actual  : []map[string]interface {}{map[string]interface {}{"animal":"cat", "color":"green", "fruit":"apple", "shape":"circle"}, map[string]interface {}{"animal":"dog", "color":"green", "fruit":"apple", "shape":"circle"}, map[string]interface {}{"animal":"cat", "color":"green", "fruit":"pear"}, map[string]interface {}{"animal":"dog", "color":"green", "fruit":"pear"}, map[string]interface {}{"fruit":"banana"}, map[string]interface {}{"animal":"cat", "fruit":"banana"}}
```
The pipeline failed to run, which seems to be affected by the order. Do we need to consider the impact of the order? If not, can we just add a sort before the final return?

PR #2658, opened 2025-02-10
Index entries in action cache directory by SHA

ChemiCalChems:
Previously they were indexed by ref name, which could be a SHA or a tag/branch name, but if a branch name was used and that branch was updated, the new reusable workflow action wouldn't be pulled.

This meant the user would have to delete the associated cache entry to be able to use the new version of the reusable workflow. If the user really wants stability, they should specify the reusable workflow by
SHA instead of by branch name/tag.

Ready to merge on my side.

### DIFF
--- a/pkg/common/git/git.go
+++ b/pkg/common/git/git.go
@@ -280,7 +280,7 @@ func CloneIfRequired(ctx context.Context, refName plumbing.ReferenceName, input 
 	return r, nil
 }
 
-func gitOptions(token string) (fetchOptions git.FetchOptions, pullOptions git.PullOptions) {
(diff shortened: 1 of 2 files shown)


ChemiCalChems:
@ChristopherHX I agree that using the new action cache would be better, but since as you said for the time being a full transition won't be happening, I believe it's important to support this. I initially faced this behavior when using Gitea's action runner which is used for CI/CD, not just local running of actions, which is forked from this, and there it was a PITA to remove the cache.

I also noticed I wasn't using the run context auth token to perform the listing operation, hence the PR update.

PR #2673, opened 2025-02-21
chore: Group DependableBot Updates monthly

ChristopherHX:
* run only 12 times per year (the cron job for releases might need adjustments)
* group as much as possible
* reduce merge conflicts by running less often

I am going to evaluate this change in my fork, what do you think?

This should free my inbox a lot.


cplee (approved):

panekj:
I'm thinking maybe we could replace dependabot with renovate or add renovate and use this PR. Renovate can keep dependency dashboard (an open issue) and keeps it updated.

panekj:
Example: https://github.com/renovatebot/renovate/issues/2958

PR #2680, opened 2025-02-22
removes arbitrary whites before unicode char/emojis

exastone:
there's inconsistent usage of whitespace before and after emoji usage in log messages making the output excessively messy and messages misaligned.

this a bandaid fix as I could prob refactor the logger to use a standard convention so whitespace isn't hardcoded but for the time being this would be a welcome partial fix

Before fix output example:
act-whitespace-example-before

After fix output example:
act-whitespace-example-after

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -413,7 +413,7 @@ func newRunCommand(ctx context.Context, input *Input) func(*cobra.Command, []str
 				DisableQuote:     true,
 				DisableTimestamp: true,
 			})
-			l.Warnf(" \U000026A0 You are using Apple M-series chip and you have not specified container architecture, you might encounter issues while running act. If so, try running it with '--container-architecture linux/amd64'. \U000026A0 \n")
+			l.Warnf("\U000026a0 You are using Apple M-series chip and you have not specified container architecture, you might encounter issues while running act. If so, try running it with '--container-architecture linux/amd64'.\U000026a0 \n")
 		}
 
 		log.Debugf("Loading environment from %s", input.Envfile())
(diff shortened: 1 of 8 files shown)


cmd/root.go
```diff
@@ -413,7 +413,7 @@ func newRunCommand(ctx context.Context, input *Input) func(*cobra.Command, []str
 				DisableQuote:     true,
 				DisableTimestamp: true,
 			})
-			l.Warnf(" \U000026A0 You are using Apple M-series chip and you have not specified container architecture, you might encounter issues while running act. If so, try running it with '--container-architecture linux/amd64'. \U000026A0 \n")
```
panekj, line 416: Don't change this, it's intentional

exastone: sure thing, fixed now

exastone:
whether to leave them in or remove them can we agree that it's currently inconsistent throughout the code base and should have something done. at the very least can we get this fix in so the beginning of messages are atleast *more* aligned then they currently are?

panekj:
I'm fine with merging this if someone is bothered by the space count. For reference this is the PR that was supposed to fix logging: https://github.com/nektos/act/pull/928

exastone:
> changed

i also thought maybe there was some form of intentionality to it but i haven't been able to decipher anything, seems random. Maybe just my poor reading comprehension but it's more distracting than anything imo

exastone:
@ChristopherHX this should be fixed now, looks like it was just a consequence of 3x space being used in the test string.

For future ref; what's the stance on using either a ~/.config/act/config.toml (or whatever file format) or looking for a environment variable (e.g. ACT_PLAIN_OUTPUT=true) to disable emojis? This would dictate what a purposed (better?) solution would look like for a refactor of logger to handle emojis

PR #2687, opened 2025-02-27
Clarify --no-skip-checkout flag

jsoref:
A peer didn't understand the behavior/way to get the other behavior. I think this clarifies things...


Habbie (approved):

Habbie:
> A peer didn't understand the behavior/way to get the other behavior. I think this clarifies things...

that was me! the new text is clearer to me

cplee (approved):

PR #2695, opened 2025-03-17
feat: add --completion flag

PeterCardenas:
adds a --completion flag to generate shell completion scripts builtin to
cobra. this includes completion for the --completion flag itself.

also allows for unused variables to be prefixed with an underscore instead of _only_ allowed to be `_` and disallows files as completion


PeterCardenas:
lol how is this medium size

PR #2699, opened 2025-03-17
Use localhost as default binding address

Dr-Electron:
Not sure why an the outbound ip approach was used but to me it makes more sense to use localhost.

I just had way to many issue trying to get `--artifact-server-path` using and in the end the issue was resolved by using `--artifact-server-addr localhost`.


Dr-Electron:
> Does some sort of firewall block this approach for you?

Could have been the issue. But I just tried to look into it again and currently I can't reproduce my issue with a simple upload/download action. Not sure what exactly I did back then. Will close this for now.

If it ever happens again I will maybe create another PR to mention in the docs as a potential troubleshoot step. Wdyt?

PR #2718, opened 2025-03-28
feat: Add support for ACT_CONFIG_HOME env var

MatejKafka:
Implement support for a custom config directory through a new `ACT_CONFIG_HOME` env var. If set, it overrides the default config file locations from XDG and the home dir, but still uses `.actrc` from the current directory.

With this change, it should be possible to bundle `act` as a fully portable program by setting `ACT_CONFIG_HOME` and `XDG_CACHE_HOME` to a local directory.


Ziip-dev:
This would be super cool in combination with devbox / nix env to enable more isolated development environments!

PR #2721, opened 2025-04-01
chore(ci): Only promote if possible

jsoref:
Without this, when people update `master` in forks with actions enabled, they get failures...

* [Before](https://github.com/check-spelling-sandbox/act/actions/runs/14186237035)
* [After](https://github.com/check-spelling-sandbox/act/actions/runs/14196148882/job/39771630398)


jsoref:
I've gone both ways in the past. I have no particular preference. My goal is smoothest adoption, and as you note this requires no additional work by the master repo owner.

Since most act forks are going to be public and _most_ forks won't have so many other repositories burning minutes, I think the 1 action minute isn't particularly expensive. That said, I'd be more than happy to use the vars approach if that's acceptable.

cplee (approved):

.github/workflows/promote.yml
```diff
@@ -11,12 +11,19 @@ jobs:
     environment: promote
     steps:
       - uses: actions/checkout@v4
+        id: checkout
+        env:
+          has_promote_token: ${{ secrets.PROMOTE_TOKEN && '1' || '' }}
```
yeikel, line 16: This was a nice trick. I learned something new. Thanks!

Issue #2722, opened 2025-04-05
Act should pass proxy build args when building Docker actions


john-tipper:
### Act version

act version 0.2.76

### Feature description

Act does not pass any build args to Docker when it builds Docker-based actions. As a result, if Act is run in an enterprise environment where access to the internet is through a proxy, then these actions all break and are unusable, i.e. all third-party docker actions are unusable.

Act should allow the user to optionally pass in proxy variables (HTTP_PROXY, HTTPS_PROXY and NO_PROXY) to the build of Docker actions. This should also include the lowercased versions too.

john-tipper:
PR submitted, where `--pass-proxy-vars-to-docker-build` will pass the proxy variables from the Act environment into the Docker build command.

john-tipper:
Hi @Wenzel,

I believe the reason your client configuration file is being ignored is because Act is using the Docker API, not the Docker client.

There are several configuration steps that need to be resolved in order for Act to be able to run behind a proxy:

1. Configuration of the proxy for building Docker-based GitHub Actions. This allows most normal off-the-shelf third party actions to still work, where they will make calls like `apt update` etc and could not otherwise reach the internet. This is the use case for my PR, where we explicitly pass any proxy variables into the build step of the container action as a build-arg.
2. Configuration of the Docker daemon so that it can pull containers for building actions. This is done by starting the Docker daemon when suitable HTTP(S)_PROXY and NO_PROXY environmental variables have been set. Running `docker system info` will show if these values have been picked up. Here's an example of doing this in CodeBuild: https://github.com/aws/aws-codebuild-docker-images/issues/355
3. Configure any CA certificate bundles into container actions if there are services inside the enterprise environment that use a private CA. This can be done by supplying a `--container-actions -v /path/to/ca.pem:/container/path/to/ca.pem:rc` parameter, where the container path will vary according to the flavour of container OS being run. Supplying this multiple times (to cover Ubuntu, Centos/RH, Alpine) should probably cover most bases.

Wenzel:
Hi @john-tipper,

thanks to your detailed reply, i learned about `--container-options` and was able to mount my `.docker/config.json`:

`act --container-options '--mount type=bind,source=/home/mtarral/.docker/config.json,target=/root/.docker/config.json'`

The issue was that one of my job was invoking Docker directly, and client is checking for that file to inject the proxy env var into the containers (build or run commands)

PR #2723, opened 2025-04-05
Enable passing of proxy variables to Docker build

john-tipper:
Implements #2722.

This PR adds a flag called `--pass-proxy-vars-to-docker-build` which allows the user to pass proxy variables through to the build of Docker-based actions. These variables are HTTP_PROXY, HTTPS_PROXY and NO_PROXY, plus the lowercase versions too because not all applications respect all cases.

The variables that are set are taken from the environment of Act itself. If the URL includes a password then this is redacted when logging the command passed to Docker.

This PR allows Act to be used within an enterprise environment where access to the internet is via an authenticating proxy, otherwise Docker-based actions are not able to be used if they need to access the internet as part of their build process.

### DIFF
--- a/cmd/input.go
+++ b/cmd/input.go
@@ -62,6 +62,7 @@ type Input struct {
 	useNewActionCache                  bool
 	localRepository                    []string
 	listOptions                        bool
+	passProxyVarsToDockerBuild         bool
 }
 
 func (i *Input) resolve(path string) string {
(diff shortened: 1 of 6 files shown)


john-tipper:
@mergifyio requeue

panekj:
I think it would be better to make it an option that allows passing any custom envvars rather than just proxy

john-tipper:
> I think it would be better to make it an option that allows passing any custom envvars rather than just proxy

I don't want to pass key=value pairs in case there are sensitive values in there which would get logged when the docker command gets printed, so I'd like to just pass envar keys. Would you be ok to have this as a flag that has a default value of `HTTP_PROXY,HTTPS_PROXY,NO_PROXY,http_proxy,https_proxy,no_proxy` and where the code then iterates through those keys and fetches the values from the environment? Users could supply their own values if they wanted - might be useful if needing to change things like `AWS_CA_BUNDLE` for example.

If we went down this route, what would the flag be called? `--pass-proxy-vars-to-docker-build` no longer makes sense IMO. Something like `--pass-envvars-to-docker-build`, perhaps?

john-tipper:
@panekj Please take a look at the revised PR, you can now do as you wish.


- No `--pass-envvars-to-docker-build` flag results in no variables being passed to docker build.
- `--pass-envvars-to-docker-build` by itself causes `HTTP_PROXY,HTTPS_PROXY,NO_PROXY,http_proxy,https_proxy,no_proxy` to be passed to docker build
- `--pass-envvars-to-docker-build=foo,bar` causes `foo` and `bar` to be retrieved from the env and passed to docker build

userid0x0:
Hi,
sorry for jumping in here - I just want to mention I have the same proxy issue on my side.

PR #2730, opened 2025-04-12
Use complete go version number

basicer:
Doesn't work on my version of ubuntu without this.


basicer:
This was for go version

go version go1.22.2 linux/amd64

PR #2734, opened 2025-04-16
docs: removed the workflow badge

operagxsasha:


panekj:
I prefer if projects are not (ab)used in any kind of crypto scams/drops

qoomon:
@panekj could explain why this change introduces a scam. I just don't understand

panekj:
Check user profile for their activity, it is mostly tiny typo/link updates. This is done to claim "drops" from some this-month-new-cryptocoin or to bypass the requirement for manual workflow run approval for the first time contributors.

operagxsasha:
@panekj
You have the right to think so. But you see, I haven't earned a single penny in a year, even if that were my main goal, I wouldn't be able to devote so much time to it. Regarding the same type of commits, I am not a developer, unfortunately my current level allows me to make such corrections and most developers thank me for this work and it is nice. And due to the scam, I lost money on buying altcoins, nft, play2earn, installing nodes, spinning drops and in my opinion this is a bigger scam. Thank you

Issue #2585, opened 2024-12-24
Pull access denied, repository does not exist or may require 'docker login'


cybersokari:
### Bug report info

```plain text
act version:            0.2.70
GOOS:                   darwin
GOARCH:                 arm64
NumCPU:                 8
Docker host:            unix:///var/run/docker.sock
Sockets found:
        /var/run/docker.sock
        $HOME/.docker/run/docker.sock
Config files:           
        /Users/sokari/Library/Application Support/act/actrc:
                -P ubuntu-latest=catthehacker/ubuntu:act-latest
                -P ubuntu-22.04=catthehacker/ubuntu:act-22.04
                -P ubuntu-20.04=catthehacker/ubuntu:act-20.04
                -P ubuntu-18.04=catthehacker/ubuntu:act-18.04
Build info:
        Go version:            go1.23.3
        Module path:           command-line-arguments
        Main version:          
        Main path:             
        Main checksum:         
        Build settings:
                -buildmode:           exe
                -compiler:            gc
                -ldflags:             -X main.version=0.2.70
                DefaultGODEBUG:       asynctimerchan=1,gotypesalias=0,httplaxcontentlength=1,httpmuxgo121=1,httpservecontentkeepheaders=1,tls10server=1,tls3des=1,tlskyber=0,tlsrsakex=1,tlsunsafeekm=1,winreadlinkvolume=0,winsymlink=0,x509keypairleaf=0,x509negativeserial=1
                CGO_ENABLED:          1
                CGO_CFLAGS:           
                CGO_CPPFLAGS:         
                CGO_CXXFLAGS:         
                CGO_LDFLAGS:          
                GOARCH:               arm64
                GOOS:                 darwin
                GOARM64:              v8.0
Docker Engine:
        Engine version:        27.4.0
        Engine runtime:        runc
        Cgroup version:        2
        Cgroup driver:         cgroupfs
        Storage driver:        overlay2
        Registry URI:          https://index.docker.io/v1/
        OS:                    Docker Desktop
        OS type:               linux
        OS version:            
        OS arch:               aarch64
        OS kernel:             6.10.14-linuxkit
        OS CPU:                8
        OS memory:             7837 MB
        Security options:
                name=seccomp,profile=unconfined
                name=cgroupns
```

### Command used with act

```sh
docker build -t act-github-actions-dockeraction:latest -f dockerfiles/action.Dockerfile .

act push --job test-action-local --secret-file ./config/act.secrets \
[...]

Issue #1633, opened 2023-02-18
Using -P platform with local image


dannystaple:
### Act version

0.2.42

### Feature description

I would like to be able to use a local image for the `-P` flag. 
For example:
```
act --eventpath myevent.json \
  -P ubuntu-latest=my-local-latest
```

The intent would be to take the medium image and add other parts I might need, for example the python components if the intent is to test pipelines around the python ecosystem.

Currently running this, it will force pull from a remote on dockerhub, however, this may simply be a local image that I do not intend to put onto dockerhub.

dannystaple:
Worked a treat - just updating the docs to reflect that.

joanrodriguez:
Hey, the flag doesn't work for me if I never uploaded the image to Docker Hub....

Steps to reproduce:
- Create Docker image locally and give repository name (foobar/docker-github-actions-runner)
- Run this:
  ```
  act --container-architecture linux/arm64/v8
  --pull=false
  -P ubuntu-latest=foobar/docker-github-actions-runner
  -P self-hosted=foobar/docker-github-actions-runner
  ```
Returns: `Error: Error response from daemon: pull access denied for spingwun/docker-github-actions-runner, repository does not exist or may require 'docker login': denied: requested access to the resource is denied`

Do I make a mistake with the flags or the feature doesn't exist?

dannystaple:
@joanrodriguez I take it the line continuation characters are not posted just for aesthetics or that this was all on one line? 
Can you show the docker image in docker list. It isn't looking for a local arm64 image when you have a x86 image is it/

joanrodriguez:
Hey guys, I had some conflicting stuff in my `.~/.actrc`. After emptying it and removing the architecture tag, I'm able to use a local docker image :)

thomashutcheson-msm:
> Hey, the flag doesn't work for me if I never uploaded the image to Docker Hub....
> 
> Steps to reproduce:
> 
> * Create Docker image locally and give repository name (foobar/docker-github-actions-runner)
> * Run this:
>   ```
>   act --container-architecture linux/arm64/v8
>   --pull=false
>   -P ubuntu-latest=foobar/docker-github-actions-runner
>   -P self-hosted=foobar/docker-github-actions-runner
>   ```
> 
> Returns: `Error: Error response from daemon: pull access denied for foobar/docker-github-actions-runner, repository does not exist or may require 'docker login': denied: requested access to the resource is denied`
> 
[...]

PR #3225, opened 2025-05-24
added info log when container image platform mismatched

rmfleet:
Relates to issue https://github.com/nektos/act/issues/2585
Relates to issue https://github.com/nektos/act/issues/1633

where the container image platform architecture doesn't match and a local image is being used, leading to the code then attempting to fetch remotely.

This can be confusing as no error is provided about the mismatched architecture and if no remote image exists, the run fails.

Added info logging to indicate that the platforms do not match as a hint.


pkg/container/docker_images.go
```diff
@@ -8,8 +8,23 @@ import (
...
+
+	if imagePlatform == platform {
+		return true
+	}
+
+	logger := common.Logger(ctx)
+	logger.Infof("image found but platform does not match: %s (image) != %s (platform)\n", imagePlatform, platform)
+	return false
+}
+
```
panekj, line 27: I don't think this needs to be a separate function

rmfleet: Perhaps not by necessity, however it's a discreet abstraction with a singular function following CleanCode methodology. Reusable if necessary and reduces complexity in the caller function ImageExistsLocally; making readability more comprehensive.

Is there an issue with abstracting it?

panekj: The abstraction is unnecessary because it's not used anywhere outside this file (or even function), it's a premature refactoring. I also disagree that it reduces complexity, now instead of single functions I have to look at 2 separate functions to see whole logic. This isn't a massive function where it's very easy to get yourself lost so reducing complexity is not a concern.

panekj (approved):

PR #5294, opened 2025-06-12
fix: matrix expansion logic

ChristopherHX:
* add another test

* remove //nolint:gocyclo

TODO check if the test in https://github.com/nektos/act/blob/162671cb083609e531939db62cf6b176d6d5a376/pkg/model/workflow_test.go could be merged here


ChristopherHX:
The new test case of https://github.com/nektos/act/pull/2649 can be pulled after this changed and it succeeds.

ChristopherHX:
Summary, I forget to remove old code that messed up my newer code. Most likely due to an invalid test case that is changed in both this and linked older PR.

panekj (approved):

PR #5326, opened 2025-06-13
feat: allow workflow description

ChristopherHX:
Otherwise act would reject such workflows right away


ChristopherHX:
In my other code splittet version I just replaced the file with a newer copy of https://github.com/actions/languageservices/blob/main/workflow-parser/src/workflow-v1.0.json

cplee (approved):

PR #5891, opened 2025-08-01
chore: mergify does no longer support queue_conditions != merge_condi…

ChristopherHX:
…tions for require status checks to be up to date

> Configuration not compatible with a branch protection setting
The branch protection setting Require branches to be up to date before merging is not compatible with max_parallel_checks>1, queue_conditions != merge_conditions and must be unset.

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -49,7 +49,7 @@ pull_request_rules:
       queue:
 queue_rules:
   - name: default
-    queue_conditions:
+    queue_conditions: &queue_conditions
       - '#changes-requested-reviews-by=0'
       - or:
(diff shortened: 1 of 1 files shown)


ChristopherHX:
Actually idk if this resolves the problem

https://github.com/nektos/act/runs/47170098045

The mergify config editor can successfully parse this change and shows that merge and queue conditions are the same

ref <https://github.com/nektos/act/pull/5889>

ChristopherHX:
Hi @cplee,

Manual Merge required, I can understand from a security point of view why mergify is not using the changed config to merge this PR.

PR #5900, opened 2025-08-13
fix broken links

zonescape:
Fix broken links


Andrej730:
https://help.github.com/en/actions/ seems to work just as an alias for https://help.github.com/en/actions/
Does it break in some case?

PR #5912, opened 2025-08-21
Limit servers to IPv4 to make them work with WSL

stickeegreg:


ChristopherHX:
:eyes: 

This is the first time I ever read that listening on both ipv6 and ipv4 is a problem source...

I cannot verify this until September.

stickeegreg:
I know it sounds very odd, and I've not found any documentation as to why it should be that case. All I can say is that it worked for me :/
PS I didn't test the cache server, only the artifacts server.

ChristopherHX:
@stickeegreg This is for scenario using act binary within wsl?

As far I know, the act binary for windows has no such issues?

My idea would be testing if **running inside wsl** then use ipv4 only, otherwise avoid regressions for everyone else.

PR #5928, opened 2025-09-13
Update LICENSE to 2025

laraibg786:


panekj:
This is incorrect. Licence should not have updated year.

PR #5941, opened 2025-09-26
refactor: replace sort.Slice with slices.Sort for natural ordering

asttool:
There is a [new function](https://pkg.go.dev/slices#Sort) added in the go1.21 standard library, which can make the code more concise and easy to read.

### DIFF
--- a/pkg/model/planner.go
+++ b/pkg/model/planner.go
@@ -8,7 +8,7 @@ import (
 	"os"
 	"path/filepath"
 	"regexp"
-	"sort"
+	"slices"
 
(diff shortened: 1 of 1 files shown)


asttool:
@panekj @cplee  Hi, Could you please review this PR at your convenience? Thank you very much.

panekj:
This is just a single function, it doesn't change anything and the code readability stays the same IMO

asttool:
> This is just a single function, it doesn't change anything and the code readability stays the same IMO

Thanks for your reply. I have closed it.

PR #5948, opened 2025-10-10
docs(man): redact IP-like defaults for address flags in generated man page - Fixes #5894

### LINKED ISSUE #5894: Do not embed outbound ip address in man page
### Bug report info

```plain text
act version:            0.2.80
GOOS:                   linux
GOARCH:                 amd64
NumCPU:                 16
Sockets found:
        $XDG_RUNTIME_DIR/podman/podman.sock
Config files:
Build info:
        Go version:            go1.24.5
        Module path:           github.com/nektos/act
        Main version:          (devel)
        Main path:             github.com/nektos/act
        Main checksum:
        Build settings:
                -buildmode:           pie
                -compiler:            gc
                -trimpath:            true
                CGO_ENABLED:          1
                GOARCH:               amd64
                GOOS:                 linux
                GOAMD64:              v1

Error: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
```

### Command used with act

```sh
act --man-page
```

### Describe issue

The generated man page includes the outbound ip address of the server which the act generated the man page of. This makes the package non-reproducible for Arch Linux as rebuilding the package on a different server will record a different IP Address making the package non-reproducible.

From a privacy perspective it also doesn't seem worth it to record the users IP Address in a man page, replacing it with an example ip address should resolve both issues.

```
│ ├── usr/share/man/man1/act.1.gz
│ │ ├── act.1
│ │ │ @@ -26,15 +26,15 @@
│ │ │  	If action contents exists, it will not be fetch and pull again. If turn on this, will turn off force pull
│ │ │  
│ │ │  .PP
│ │ │  \fB-a\fP, \fB--actor\fP="nektos/act"
│ │ │  	user that triggered the event
│ │ │  
│ │ │  .PP
│ │ │ -\fB--artifact-server-addr\fP="162.55.28.166"
│ │ │ +\fB--artifact-server-addr\fP="147.75.84.133"
│ │ │  	Defines the address to which the artifact server binds.
│ │ │  
│ │ │  .PP
│ │ │  \fB--artifact-server-path\fP=""
│ │ │  	Defines the path where the artifact server stores uploads and retrieves downloads from. If not specified the artifact server will not start.
│ │ │  
│ │ │  .PP
│ │ │ @@ -46,15 +46,15 @@
│ │ │  	bind working directory to container, rather than copy
│ │ │  
│ │ │  .PP
│ │ │  \fB--bug-report\fP[=false]
│ │ │  	Display system information for bug report
│ │ │  
│ │ │  .PP
│ │ │ -\fB--cache-server-addr\fP="162.55.28.166"
│ │ │ +\fB--cache-server-addr\fP="147.75.84.133"
│ │ │  	Defines the address to which the cache server binds.
│ │ │  
│ │ │  .PP
│ │ │  \fB--cache-server-external-url\fP=""
│ │ │  	Defines the external URL for if the cache server is behind a proxy. e.g.: https://act-cache-server.example.com. Be careful that there is no trailing slash.
│ │ │  
│ │ │  .PP
```

### Link to GitHub repository

_No response_

### Workflow content

```yml
None
```

### Relevant log output

```sh
None
```

### Additional information

_No response_

AbhinavSivanandhan:
Problem:
`act --man-page` embeds the local outbound IP in defaults for address flags (e.g., --artifact-server-addr, --cache-server-addr). That breaks reproducible builds and leaks a private detail.

Approach:
Only for documentation generation, replace printed defaults that:
- are for address-like flags (*-addr, *-address), and
- look like an IPv4 (optionally with :port)
with a deterministic placeholder: [auto-detected IP].

Runtime behavior is unchanged. When `act` actually runs, it still uses the dynamic outbound IP.

Implementation:
- Added a small sanitizer called from `generateManPage()` that walks the root command (and subcommands) and rewrites `DefValue` for matching flags.
- Unit test `TestSanitizeFlagDefaultsForDocs` validates behavior across persistent/local flags and subcommands.

Testing
- `go test ./cmd -run Sanitize -v` passes
- Manual: `./act --man-page` shows `[auto-detected IP]` in place of a real IP for address flags.

Fixes #5894

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -253,6 +253,42 @@ func bugReport(ctx context.Context, version string) error {
 	return nil
 }
 
+// ipPortRe is used only by the man-page generator to sanitize printed defaults. It matches IPv4 (optionally with :port)
+var ipPortRe = regexp.MustCompile(`\b(?:\d{1,3}\.){3}\d{1,3}(?::\d+)?\b`)
+
+// Keep scope tight: only redact flags that look like address defaults.
+func looksLikeAddrFlag(name string) bool {
+	return strings.HasSuffix(name, "-addr") || strings.HasSuffix(name, "-address")
+}
+
+func sanitizeFlagSet(fs *pflag.FlagSet) {
+	// For manpage generation only: show a neutral, deterministic doc default.
+	const docAddrPlaceholder = "[auto-detected IP]"
+	if fs == nil {
+		return
+	}
+	fs.VisitAll(func(f *pflag.Flag) {
+		if !looksLikeAddrFlag(f.Name) {
+			return
+		}
+		if ipPortRe.MatchString(f.DefValue) {
+			f.DefValue = docAddrPlaceholder
+		}
+	})
+}
+
+// Sanitize this command and its subcommands for doc generation only for --man-page. This changes printed defaults, not runtime behavior.
+func sanitizeFlagDefaultsForDocs(cmd *cobra.Command) {
+	if cmd == nil {
+		return
+	}
+	sanitizeFlagSet(cmd.PersistentFlags())
+	sanitizeFlagSet(cmd.Flags())
+	for _, child := range cmd.Commands() {
+		sanitizeFlagDefaultsForDocs(child)
+	}
+}
+
 func generateManPage(cmd *cobra.Command) error {
 	header := &doc.GenManHeader{
 		Title:   "act",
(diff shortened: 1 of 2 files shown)


sarvekshayr:
This PR has been marked as stale due to 21 days of inactivity. Please comment or remove the stale label to keep it open. Otherwise, it will be automatically closed in 7 days.

PR #5944, opened 2025-10-01
chore(mergify) Add merge queue configuration to .mergify.yml

ChristopherHX:
* So many breaking changes
* All credits to AI
* 0% Tested
* Mergify refuses to merge anything including this PR

PR #5951, opened 2025-10-19
Add 24.04 images, override -latest with -24.04

### LINKED ISSUE #2329: Add Ubuntu 24.04
### Act version

act version 0.2.62

### Feature description

Add `ubuntu24.04` image github actions now support it https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md

keks:
Unfortunately the repo with the images hasn't updated their -latest tag in the last year, so we explicitly set it here.

Closes #2329

### DIFF
--- a/IMAGES.md
+++ b/IMAGES.md
@@ -11,6 +11,7 @@
 | Image                                 | Size                                                       |
 | ------------------------------------- | ---------------------------------------------------------- |
 | [`node:16-bullseye`][hub/_/node]      | ![`bullseye-size`][hub/_/node/16-bullseye/size]            |
+| [`node:16-bookwork-slim`][hub/_/node] | ![`micro-bookwork-size`][hub/_/node/16-bookwork-slim/size] |
 | [`node:16-bullseye-slim`][hub/_/node] | ![`micro-bullseye-size`][hub/_/node/16-bullseye-slim/size] |
 | [`node:16-buster`][hub/_/node]        | ![`buster-size`][hub/_/node/16-buster/size]                |
 | [`node:16-buster-slim`][hub/_/node]   | ![`micro-buster-size`][hub/_/node/16-buster-slim/size]     |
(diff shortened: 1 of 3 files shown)


ChristopherHX (commented):
Look (almost) good to me, see comments

This is not going to update the actrc files of any existing user,  asked when empty then written once

---

FYI, if you see no merge in near future
- I am unable to merge anything right now without owner override of cplee
- Maintainer merge is broken, due to mergify breaking changes that I tried to fix in three weeks ago <https://github.com/nektos/act/pull/5944>, still unable to merge dependency updates that had automerge permission

Andrej730:
Don't want this go out of scope, but maybe there should be a warning in case someone has their `latest` in `actrc` pointing to `22.04`? So users will be pointed towards resetting their `actrc` (either by removing it completely or removing `-P` lines from it) and updating `latest` in their config. 
Just came across this issue trying to switch to this fork.

ChristopherHX (dismissed):
Looks good to me

keks:
Wow you were faster than me. I added the warning that @Andrej730 asked for, but I am not sure it's where you'd like it to be.

panekj:
The warning is wrong, as it will always trigger when using latest image

keks:
> The warning is wrong, as it will always trigger when using latest image

I think you were right that there was an issue, but it was a different issue.

The issue this PR solves is that catthehacker/ubuntu seems more or less unmaintained and `ubuntu:act-latest` and `ubuntu:full-latest` have been out of date for years. That's why we explicitly set it to `ubuntu:act-24.04` and `ubuntu:full-24.04`. If the config still contains a setting that points `ubuntu-latest` to `ubuntu:act-latest` or `ubuntu:full-latest`, then we know it is out of date.

However, so far the warning only came for one of the image variants. I added checks for all of the default images. I also got rid of the HasPrefix - I thought this is needed because the string might contain a newline that would mess up a simple equality check, but now I see we are stripping arg of whitespace after taking it from the scanner.

ChristopherHX:
> The issue this PR solves is that catthehacker/ubuntu seems more or less unmaintained

Well we could fix this, we both have write access to the mentioned repository,

BTW I am a bit sceptical about everything you changed after my approval.

keks:
Happy to revert all of the warning stuff, just wanted to make Andrej happy. I suppose the cleaner fix would be to update the container tag, but I can't do that in a PR.

Andrej730:
The warning is not critical or anything, just wanted to avoid people stumbling upon the same issue and creating issues on why `latest` is still not pointing to `24.04`.

panekj:
`-latest` will point to `24.04` once this run completes: https://github.com/catthehacker/docker_images/actions/runs/19011364007/job/54292773341

ChristopherHX:
bumped full-latest as well in the same repo

Andrej730:
This can be closed?

ChristopherHX:
Should be updated to not alert on latest tag and preserve them

> override -latest with -24.04 #

only this should be reverted

franzliedke:
I believe this hack should also be adjusted, to have the correct `ImageOS` environment variable:

https://github.com/nektos/act/blob/1030ad20bf3da1856f8603b758114d5f1d1cbf3f/pkg/runner/run_context.go#L1066

ChristopherHX:
Why does act even define that one env variable? Shouldn't this only do the image itself? (the "small" node images are no ubuntu images, so defining it only confuses applications)

angelikatyborska:
Hi there! If I understand correctly, the ubuntu 22 github runner will be deprecated in a month: https://github.com/actions/runner-images/issues/14254

If that happens, we will be all forced to upgrade to a newer runner and lose the ability to use act to test our workflows.

PR #5961, opened 2025-10-30
Fix platform parsing to support multiple equals signs

TheRobotCarlson:
Use strings.LastIndex instead of strings.Split to correctly parse platform strings containing multiple equals signs (e.g., runner=fast-disk-network=catthehacker/ubuntu:act-22.04).

This allows the platform key to contain equals signs, splitting only on the last occurrence to separate the key from the value.


crohr:
Would be nice to get this added, thanks for the PR!

PR #5973, opened 2025-11-22
fix possible leakage of secrets in debug logging

drornir:
Found this while browsing the source code.

I just thought it was odd to to a regex replace on the string 'true' or 'false' so it caught my attention.

I know it's not an important fix but I was here so why not.

reference to the commit that originally changed this:
https://github.com/nektos/act/commit/4391a10d5a013afb3a4cbc1640eb745132593d7a#diff-4fe45e900ed33a3395bb42c8d8a85c1afb94987b2368bfe31d62e9e1ebd69060R124
image

### DIFF
--- a/pkg/runner/expression.go
+++ b/pkg/runner/expression.go
@@ -234,11 +234,11 @@ type expressionEvaluator struct {
 
 func (ee expressionEvaluator) evaluate(ctx context.Context, in string, defaultStatusCheck exprparser.DefaultStatusCheck) (interface{}, error) {
 	logger := common.Logger(ctx)
-	logger.Debugf("evaluating expression '%s'", in)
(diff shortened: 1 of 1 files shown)


ChristopherHX (changes requested):
this does not look right, the printableInput might prevent leakage that you hardcoded in yaml.

However if the workflow content contains a secret than it is no longer secret

e.g.
```yaml
- run: |
    echo ::add-mask::${{ vars.SOMETHING_YOU_MIGHT_WANTED_TO_HIDE }}

ChristopherHX (approved):
_I am unsure when merges become possible again: blocked by https://github.com/nektos/act/pull/5944_

drornir:
@ChristopherHX bumping this.
No pressure whatsoever, the whole point of this PR is to help you

PR #6029, opened 2026-03-19
fix: serialize concurrent GoGitActionCache access per repo

### LINKED ISSUE #6028: GoGitActionCache race condition with concurrent matrix jobs
### Describe the bug

When multiple matrix jobs within a reusable workflow concurrently fetch the same action via `GoGitActionCache`, a race condition causes intermittent failures:

```
failed to read 'action.yml' from action 'Assume deploy ASG Role' with path '' of step: file does not exist
```

or:

```
GoGitActionCache failed to fetch https://github.com/... : repository not found: Repository not found.
```

### Root cause

`GoGitActionCache.Fetch()` operates on a shared bare git repo at `gitPath` (derived from `cacheDir`). When two goroutines call `Fetch()` for the same repo concurrently:

1. Both call `git.PlainInit(gitPath, true)` — one succeeds, one gets `ErrRepositoryAlreadyExists`
2. Both create anonymous remotes and call `FetchContext()` concurrently on the same bare repo
3. The deferred `DeleteBranch(branchName)` cleanup in one goroutine can remove objects that another goroutine's `GetTarArchive()` is reading

This manifests as:
- `action.yml` not found (objects not yet written when read)
- Repository not found (bare repo in inconsistent state during concurrent init/fetch)

### To reproduce

1. Create a workflow with a reusable workflow that uses `fromJSON()` matrix with 2+ entries
2. Each matrix job must reference the same external action (e.g., `aws-actions/configure-aws-credentials`)
3. Run with act using containerized runners (GoGitActionCache is used)
4. The failure is intermittent — depends on goroutine scheduling

### Expected behavior

Concurrent matrix jobs should be able to safely share the action cache.

### Suggested fix

Add a per-repo `sync.Mutex` (stored in a `sync.Map` on `GoGitActionCache`) to serialize `Fetch()` and `GetTarArchive()` calls for the same `gitPath`. Different repos remain fully parallel.

### act version

v0.2.84

jun-mercury:
## Summary

When multiple matrix jobs within a reusable workflow concurrently fetch the same action, they share a bare git repository via `GoGitActionCache`. Concurrent `PlainInit`/`FetchContext`/`GetTarArchive` calls on the same repo cause intermittent failures:

- `failed to read 'action.yml'` — objects not yet written when another goroutine reads
- `repository not found` — bare repo in inconsistent state during concurrent init/fetch

## Fix

Add a per-repo `sync.Mutex` (stored in a `sync.Map` on `GoGitActionCache`) to serialize `Fetch` and `GetTarArchive` calls for the same `gitPath`. Different repos remain fully parallel.

Changes:
- `GoGitActionCache`: add `locks sync.Map` field and `repoMu()` helper
- `Fetch()` / `GetTarArchive()`: acquire per-repo mutex before git operations
- Receivers changed from value to pointer (`*GoGitActionCache`) since `sync.Map` must not be copied
- `GoGitActionCacheOfflineMode.Parent`: changed to `*GoGitActionCache` accordingly
- `cmd/root.go` and test files: updated to use pointer instantiation

## Reproduction

1. Create a workflow calling a reusable workflow with a `fromJSON()` matrix (2+ entries)
2. Each matrix job references the same external action (e.g., `aws-actions/configure-aws-credentials`)
3. Run with act using containerized runners
4. Failure is intermittent — depends on goroutine scheduling

Fixes #6028

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -649,7 +649,7 @@ func newRunCommand(ctx context.Context, input *Input) func(*cobra.Command, []str
 		if input.useNewActionCache || len(input.localRepository) > 0 {
 			if input.actionOfflineMode {
 				config.ActionCache = &runner.GoGitActionCacheOfflineMode{
-					Parent: runner.GoGitActionCache{
+					Parent: &runner.GoGitActionCache{
 						Path: config.ActionCacheDir,
 					},
 				}
--- a/pkg/runner/action_cache.go
+++ b/pkg/runner/action_cache.go
@@ -11,6 +11,7 @@ import (
 	"io/fs"
 	"path"
 	"strings"
+	"sync"
 	"time"
 
 	git "github.com/go-git/go-git/v5"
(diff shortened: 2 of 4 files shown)


themavik (commented):
nit: sync.Map never removes mutex entries, so a long-lived process that hits many distinct cacheDir-derived git paths could grow that map without bound.

PR #6062, opened 2026-04-03
Feature/unique container names

drivera73:
When running two instances of act against the same Docker environment, the container naming code produces consistent, repeatable names every time - even if the user isn't interested in reusing the build containers. As a result, these act instances will interfere with each other and this will result in odd errors and failures. The solution is to introduce a random string to the container name such that all containers created and managed by a specific instance of act will be distinct from those managed by other instances.

In addition, if the --reuse option is in play, this behavior is disabled and the old, repeatable naming approach (albeit with a slight tweak to the name :) ) will come into play again.

An explanation of the use case and the issues can be found [here](https://github.com/nektos/act/discussions/6058).


drivera-armedia:
Hi ...is there any hope of getting this merged into the mainline?

PR #6085, opened 2026-05-01
fix: resolve Dockerfile context from action.yml location for remote actions

### LINKED ISSUE #739: Issue: Dockerfile context relative path
I experience this with such action below

```
      - name: yq - my yq step
        uses: mikefarah/yq@v4.9.6
        with:
          cmd: yq --help
```

it appears to use https://github.com/mikefarah/yq/blob/master/Dockerfile even though it points to 
https://github.com/mikefarah/yq/blob/master/github-action/Dockerfile via https://github.com/mikefarah/yq/blob/master/action.yml#L15.

Probably somethings should be done in this line below  to fix this issue. https://github.com/nektos/act/blob/a9fe0383478554705100dc4416b235b2265f718c/pkg/runner/step_context.go#L472

## System information
    - Operating System: macOS
    - Architecture: arm64 (64-bit)
    - Apple M1: yes
    - Docker version: 20.10.7
    - Docker image used in `act`: catthehacker/ubuntu:act-latest
    - `act` version: act version 0.2.23

## Expected behaviour

use the right context path (`github.com/mikefarah/yq/github-action(/Dockerfile)`) to build the image
## Actual behaviour

use the wrong context path (`github.com/mikefarah/yq(/Dockerfile`) to build the image

## Workflow and/or repository

<details>
  <summary>workflow</summary>

```yaml
name: example workflow
on: [push]
jobs:
  example-action:
    runs-on: ubuntu-latest
    steps:
      - name: yq - my yq step
        uses: mikefarah/yq@v4.9.6
        with:
          cmd: yq --help
```

</details>

## `act` output

<details>
  <summary>Log</summary>

```none
...
[Repository xxx/myEvent] Starting container: 9403a846d8fa99c60c066cc1b480e0b2039f6d47149122953583cc4abaaed313
| Error: unknown command "yq e --help" for "yq"
| Run 'yq --help' for usage.
```

</details>

Raj-glitch-max:
When a remote action declares runs.using: docker with a relative runs.image (e.g. image: Dockerfile), the build context must be resolved relative to the directory containing action.yml, not from the repository root.

Previously readActionImpl dropped the actionPath after reading metadata, so execAsDocker always used the uses: subpath to resolve runs.image. If action.yml lives in a subdirectory but uses: has no explicit path component, the Dockerfile was resolved from the wrong location.

Changes:
- Add ActionPath string (yaml:"-") to model.Action to carry the resolved subpath through the runner without affecting YAML parsing
- Set action.ActionPath in all return paths of readActionImpl
- When action.yml is not found at the repo root, scan immediate subdirectories of actionDir/actionPath and read from the unique match; return an error if zero or multiple matches are found
- Use action.ActionPath as dockerSubpath at all three execAsDocker call sites when running remote actions (main, pre, post)

Local actions and the ActionCache tar-stream path are unaffected. Added TestActionReaderDiscoverSubdir to cover the new discovery path.

## Test output
![TestActionReaderDiscoverSubdir passing]
Screenshot from 2026-05-01 17-07-36
Screenshot from 2026-05-01 17-07-36

Fixes #739

### DIFF
--- a/pkg/model/action.go
+++ b/pkg/model/action.go
@@ -93,6 +93,7 @@ type Action struct {
 	Inputs      map[string]Input  `yaml:"inputs"`
 	Outputs     map[string]Output `yaml:"outputs"`
 	Runs        ActionRuns        `yaml:"runs"`
+	ActionPath  string            `yaml:"-"`
 	Branding    struct {
 		Color string `yaml:"color"`
 		Icon  string `yaml:"icon"`
--- a/pkg/runner/action.go
+++ b/pkg/runner/action.go
@@ -67,6 +67,7 @@ func readActionImpl(ctx context.Context, step *model.Step, actionDir string, act
 						Using: "docker",
 						Image: "Dockerfile",
 					},
+					ActionPath: actionPath,
 				}
 				logger.Debugf("Using synthetic action %v for Dockerfile", action)
 				return action, nil
@@ -99,6 +100,7 @@ func readActionImpl(ctx context.Context, step *model.Step, actionDir string, act
 							Using: "node12",
 							Main:  "trampoline.js",
 						},
+						ActionPath: actionPath,
 					}
 					logger.Debugf("Using synthetic action %v", action)
 					return action, nil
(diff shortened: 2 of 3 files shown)


Raj-glitch-max:
Addressed all Copilot review comments:
ActionCache (Comment #1): Added explicit error with explanation when os.Stat fails — clarifies that tar-backed ActionCache readers are incompatible with filesystem discovery
Multiple match error (Comment #2): Now returns fmt.Errorf("ambiguous action: found action.yml in multiple subdirectories...") with candidate paths
ReadDir error propagation (Comment #3): readErr now wrapped with context and returned instead of silently falling through
Error context (Comment #4): openErr/readErr now include step.String() + path info for consistent debugging
Test coverage (Comment #5): Added TestActionReaderDiscoverSubdirMultipleMatches and TestActionReaderDiscoverSubdirActionCacheUnsupported

PR #6096, opened 2026-05-19
feat: support concurrency.queue in workflow schema

### LINKED ISSUE #6095: Workflows using concurrency.queue fail schema validation ("Unknown Property queue")
### Bug report info

```plain text
act version:            0.2.88
GOARCH:                 arm64
NumCPU:                 12
Docker host:            unix:///Users/.../docker.sock
Sockets found:
Config files:
        /Users/.../Library/Application Support/act/actrc
        /Users/.../.actrc
Build info:
        Go version:            go1.24+
        Main path:             github.com/nektos/act
        Main version:          0.2.88
```

### Command used with act

```sh
act -l
```

### Describe issue

`act` rejects valid workflows that use `concurrency.queue`, which GitHub officially documents under the workflow-syntax reference. Any workflow with `queue: single` or `queue: max` under `concurrency` fails schema validation at parse time, so the listing/run never starts.

Relevant references:

- GitHub docs entry: https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#concurrency
- Docs commit that added `queue`: https://github.com/github/docs/commit/336b7f546d9443dab4e1fa4f0f470e45448c7abc
- SchemaStore already encodes the enum (`single | max`, default `single`) in `github-workflow.json` under `#/definitions/concurrency`: https://raw.githubusercontent.com/SchemaStore/schemastore/master/src/schemas/json/github-workflow.json
- Cross-reference in `actionlint`: https://github.com/rhysd/actionlint/issues/657

Per the docs, `queue` accepts `single` (default — at most one pending run) or `max` (up to 100 pending runs). The docs also state that `queue: max` combined with `cancel-in-progress: true` is a workflow validation error, but that mutual-exclusion is documented in prose only and not encoded in SchemaStore.

### Link to GitHub repository

_No response_

### Workflow content

```yml
name: queue-test
on: pull_request
concurrency:
  group: g
  queue: max
jobs:
  noop:
    runs-on: ubuntu-latest
    steps:
      - run: echo ok
```

### Relevant log output

```sh
Error: workflow is not valid. 'queue.yml': Line: 4 Column 3: Failed to match string: Line: 4 Column 3: Expected a scalar got mapping
Line: 4 Column 3: Failed to match concurrency-mapping: Line: 6 Column 3: Unknown Property queue
Actions YAML Schema Validation Error detected:
For more information, see: https://nektosact.com/usage/schema.html
```

### Additional information

Tracked in `pkg/schema/workflow_schema.json` - the `concurrency-mapping` definition lists only `group` and `cancel-in-progress`.

xenjke:
## Summary

Workflows using the documented `concurrency.queue` property are currently rejected by `act`'s schema with `Unknown Property queue`. GitHub added `queue` to the concurrency mapping ([docs](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#concurrency), added in [github/docs@336b7f5](https://github.com/github/docs/commit/336b7f546d9443dab4e1fa4f0f470e45448c7abc)); it accepts `single` (default) or `max`. This PR mirrors SchemaStore's existing encoding in act's hand-maintained schema so those workflows validate.

Fixes #6095.

## Changes

- `pkg/schema/workflow_schema.json` — add a top-level `concurrency-queue` definition (`allowed-values: ["single", "max"]`, following the precedent of `branch-protection-rule-activity-type` at line 188) and reference it from `concurrency-mapping.properties.queue`.
- `pkg/schema/schema_test.go` — add `TestConcurrencyQueue` covering `single`, `max`, job-level placement, and an invalid value.

### Out of scope

The docs state that `queue: max` combined with `cancel-in-progress: true` is a workflow validation error. act's schema engine (`checkMapping` in `pkg/schema/schema.go`) validates each property in isolation, so cross-property constraints cannot be expressed in the current schema vocabulary. SchemaStore makes the same trade-off — `queue` is encoded as an enum but the mutual exclusion is left to prose. Can be revisited as a follow-up.

## Test plan

- [x] `go test ./pkg/schema/...` — new test passes (`TestConcurrencyQueue` + existing tests)
- [x] `golangci-lint run ./...` — no issues
- [x] `go fmt ./...` / `go mod tidy` — no diff
- [x] Positive e2e: `act -l` against a workflow with `concurrency.queue: max` now lists jobs instead of erroring
- [x] Negative e2e: `concurrency.queue: bogus` errors with `Expected one of single,max got bogus`

## References

- GitHub docs: https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#concurrency
- Docs change that introduced `queue`: https://github.com/github/docs/commit/336b7f546d9443dab4e1fa4f0f470e45448c7abc
- SchemaStore precedent: https://raw.githubusercontent.com/SchemaStore/schemastore/master/src/schemas/json/github-workflow.json (`#/definitions/concurrency`)
- Related cross-reference in actionlint: https://github.com/rhysd/actionlint/issues/657, https://github.com/rhysd/actionlint/pull/654


McNultyyy:
Apologies — I opened #6152 covering the same `concurrency.queue` schema gap without spotting this PR first. You filed #6095 and had a fix up a minute later; mine came months afterwards and I should have checked the open PRs, not just the issues. This one has precedence.

Having read the diff, I think yours is the better-scoped fix for #6095: schema-only, with table-driven cases in `pkg/schema/schema_test.go` covering workflow level, job level, and the invalid-value rejection. That is exactly what the issue asks for and nothing more.

I have retitled #6152 so it no longer claims to close #6095 and it now points here. The only thing it carries beyond your change is the `pkg/model` side — a `Concurrency` struct with `Group`/`CancelInProgress`/`Queue` and the parsing for it — which is only actually needed by the runtime work I am doing separately, so it belongs with that rather than with this fix.

Happy to do whichever you prefer: close #6152 entirely and move the model bits into my runtime PR, or leave it as a follow-up that rebases on top of yours once this lands. If it is useful I can also open a PR against your branch with the extra `pkg/model` tests.

xenjke:
Thanks for taking a look @McNultyyy. This is the smaller schema-only fix, would it make sense to merge it first and then rebase #6152 on top for the `pkg/model` follow-up? Up to @cplee 🙏

McNultyyy:
Sounds like a plan. 

My larger-scale idea is to get the work in https://github.com/nektos/act/pull/6147 reviewed and merged. However I appreciate that it is a mammoth change and will need to be broken down and reviewed in smaller manageable pieces. 

Keen to hear both of your ideas on this and where best to have this conversation / discussion , as to not pollute this thread.

sampathintouch:
@cplee — independent verification of this change, in case it helps it move.

I hit #6095 while evaluating `act` for a GitHub Enterprise Server monorepo (75 workflows, 112 jobs, an 18-workflow reusable-workflow fan-out). Before finding this PR I wrote the same fix from scratch and landed on an identical implementation — same `queue` property, same `concurrency-queue` definition, same `allowed-values: ["single", "max"]`. Two independent attempts converging suggests this is the idiomatic shape for the schema dialect rather than one option among several.

I built this branch (`9896075`) and tested it against the real repository:

- Both affected workflows parse; stock 0.2.89 rejects both.
- `queue: single` and `queue: max` accepted at workflow level and job level.
- `queue: bogus` still rejected — the enum is enforced, so this doesn't loosen validation.
- The full dispatcher resolves a job through the reusable-workflow chain that previously failed outright.
- `go test ./pkg/schema/... ./pkg/model/...` pass; `go vet` and `gofmt` clean.

One detail that may not be obvious from the issue: the blast radius is larger than a single ignored key. Because an unknown property makes the whole job block fail to match, one `queue: max` cascades into misleading `Unknown Property runs-on` / `outputs` / `steps` errors and takes the entire file down. Worse, any workflow that `uses:` the affected file fails too — so a single key on one reusable workflow can make an entire pipeline unrunnable locally, which is how it presented for us.

Since `queue` is documented GitHub syntax ([May 2026 changelog](https://github.blog/changelog/2026-05-07-github-actions-concurrency-groups-now-allow-larger-queues/)) and already encoded in SchemaStore, this is a straightforward correctness gap rather than a new feature. It's currently a hard adoption blocker for us; happy to help with whatever this needs to land.

PR #6115, opened 2026-06-14
fix(artifacts): support upload-artifact@v7 / download-artifact@v8

### LINKED ISSUE #6114: Artifact server rejects upload-artifact@v7 / download-artifact@v8 (mime_type + sig padding)
### Summary

`act`'s v4 artifact server (`pkg/artifacts/artifacts_v4.go`) cannot complete an upload from `actions/upload-artifact@v7` (and the matching `download-artifact@v8`). The artifact handoff between jobs fails, which breaks any workflow that uses the current major versions of the official artifact actions locally.

There are **two independent** failures:

#### 1. `unknown field "mime_type"` (upload fails at `CreateArtifact`)

`parseProtbufBody` decodes request bodies with a strict `protojson.Unmarshal`, which errors on any field not present in act's vendored protobuf. `upload-artifact@v7` adds a `mime_type` field to the artifact requests:

```
level=error msg="Error decode request body: proto: (line 1:184): unknown field \"mime_type\""
```

#### 2. `Error unauthorized` (upload fails on the blob `UploadArtifact` PUT)

After fixing (1), the upload then fails the signature check. `verifySignature` decodes the `sig` query parameter with `base64.URLEncoding`, which requires `=` padding. The Azure storage SDK that `upload-artifact@v7` uses for blob upload **strips the `=` padding** when it re-serializes the signed URL to append `comp=block&blockid=...`. The strict decode fails silently (error is discarded), yielding an empty signature, so the HMAC comparison rejects the request:

```
level=error msg="Error unauthorized"
```

`expires`, `artifactName`, and `taskID` all match — only the padding-stripped `sig` causes the mismatch.

### Reproduction

```yaml
# .github/workflows/roundtrip.yml
name: roundtrip
on: push
jobs:
  up:
    runs-on: ubuntu-latest
    steps:
      - run: mkdir -p out && echo hi > out/data.txt
      - uses: actions/upload-artifact@v7
        with: { name: blob, path: out/ }
  down:
    runs-on: ubuntu-latest
    needs: up
    steps:
      - uses: actions/download-artifact@v8
        with: { name: blob }
      - run: cat data.txt
```

```
act push --artifact-server-path /tmp/artifacts
```

`upload-artifact@v4` / `download-artifact@v4` work; `@v7` / `@v8` fail as above.

### Environment

- act `0.2.89`
- `actions/upload-artifact@v7`, `actions/download-artifact@v8`

### Fix

Both are narrow, backward-compatible relaxations of the request parsing:

1. Decode protobuf bodies with `protojson.UnmarshalOptions{DiscardUnknown: true}` — act doesn't consume `mime_type`, so ignoring unknown optional fields is safe and forward-compatible.
2. Decode the `sig` with `base64.RawURLEncoding` after trimming any `=`, accepting both the padded (v4-era) and unpadded (v7+) forms.

PR incoming. Verified end-to-end with both v4↔v4 and v7↔v8 round-trips.

royteeuwen:
Fixes #6114.

The v4 artifact server rejected uploads from `actions/upload-artifact@v7` (and the matching `download-artifact@v8`) for two independent reasons. Both are narrow, backward-compatible relaxations of request parsing.

### 1. `unknown field "mime_type"`

`parseProtbufBody` used a strict `protojson.Unmarshal`, which fails on any field absent from act's vendored protobuf. `upload-artifact@v7` adds a `mime_type` field, producing `Error decode request body: unknown field "mime_type"` at `CreateArtifact`.

Decoding with `protojson.UnmarshalOptions{DiscardUnknown: true}` ignores fields act doesn't consume and keeps the server forward-compatible with future optional additions.

### 2. `Error unauthorized`

`verifySignature` decoded the `sig` query parameter with `base64.URLEncoding`, which requires `=` padding. The Azure storage SDK used by `upload-artifact@v7` for blob upload strips the padding when it re-serializes the signed URL to append `comp`/`blockid`, so the strict decode failed silently and the HMAC check rejected the upload.

Decoding with `base64.RawURLEncoding` after trimming any `=` accepts both the padded (v4-era) and unpadded (v7+) forms.

### Compatibility

Both changes only *relax* parsing — they never reject anything the previous code accepted, so v4/v5/v6 clients are unaffected. (`upload-artifact@v1`–`v3` use a separate, non-protobuf code path that this change does not touch.)

### Testing

- Added `TestCreateArtifactV4IgnoresUnknownFields` and `TestUploadArtifactV4AcceptsUnpaddedSignature`.
- `go test ./pkg/artifacts/` passes.
- Verified end-to-end with real workflows: `upload-artifact@v4` → `download-artifact@v4`, and `upload-artifact@v7` → `download-artifact@v8`, both round-trip successfully (including a ~160 MB artifact).

### DIFF
--- a/pkg/artifacts/artifacts_v4.go
+++ b/pkg/artifacts/artifacts_v4.go
@@ -219,7 +219,13 @@ func (r artifactV4Routes) verifySignature(ctx *ArtifactContext, endp string) (in
 	sig := ctx.Req.URL.Query().Get("sig")
 	expires := ctx.Req.URL.Query().Get("expires")
 	artifactName := ctx.Req.URL.Query().Get("artifactName")
-	dsig, _ := base64.URLEncoding.DecodeString(sig)
+	// Newer @actions/upload-artifact releases (v7+) upload blob content with
+	// the Azure storage SDK, which strips the `=` base64 padding from the `sig`
+	// query value when it re-serializes the signed URL to append comp/blockid.
+	// v4-era clients keep the padding. RawURLEncoding + TrimRight decodes both
+	// the padded and unpadded forms, so the signature check works across every
+	// supported artifact-action version.
+	dsig, _ := base64.RawURLEncoding.DecodeString(strings.TrimRight(sig, "="))
 	taskID, _ := strconv.ParseInt(rawTaskID, 10, 64)
 
 	expecedsig := r.buildSignature(endp, expires, artifactName, taskID)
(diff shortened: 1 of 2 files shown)


aadrian:
any chance merging this finally and cutting a new release?

PR #6132, opened 2026-07-17
fix(artifactcache): return after storage.Write error on upload

### LINKED ISSUE #6131: [Bug]: artifact cache upload missing return after storage.Write error (double response + useCache)
### Describe the bug

In the artifact cache upload handler, when `storage.Write` fails the code responds with 500 but **does not return**, then continues as if the upload succeeded:

```go
if err := h.storage.Write(cache.ID, start, r.Body); err != nil {
    h.responseJSON(w, r, 500, err)
}
h.useCache(id)
h.responseJSON(w, r, 200)
```

On Write failure this:

1. Emits a 500 JSON error body
2. Still calls `h.useCache(id)` (DB touch as if success)
3. Calls `responseJSON(200)` again → `http: superfluous response.WriteHeader` and a second body fragment

Every other error branch in this handler returns after `responseJSON`.

### Debug information

- Checkout @ `4f411281417e88660bea1c1a1749aa71ae0bd60f`
- File: `pkg/artifactcache/handler.go` (upload / Write path)

### Repro steps

Force `storage.Write` to fail (e.g. cache root is not a usable directory), then PATCH the cache upload endpoint with `Content-Range` + body.

Observed pattern:

- status 500
- response body like `{"error":"..."}{}`
- log: `superfluous response.WriteHeader` from `responseJSON`

### Expected behavior

On Write error: respond once with 500 and return; do not call `useCache` or emit 200.

### Actual behavior

Double response + `useCache` after failure.

### Suggested direction

```go
if err := h.storage.Write(cache.ID, start, r.Body); err != nil {
    h.responseJSON(w, r, 500, err)
    return
}
```

I'd be happy to open a PR with a regression test.

baiyuxi930826:
## Summary

Fixes #6131: when `storage.Write` fails during artifact cache upload, the handler now returns after the 500 response instead of continuing into `useCache` and a second 200 `responseJSON` (which produced a double body and `superfluous response.WriteHeader`).

## Test plan

- [x] `go test ./pkg/artifactcache/ -run TestHandler`
- [x] New regression case: force Write failure (storage root is a file) and assert single 500 JSON error body

### DIFF
--- a/pkg/artifactcache/handler.go
+++ b/pkg/artifactcache/handler.go
@@ -281,6 +281,7 @@ func (h *Handler) upload(w http.ResponseWriter, r *http.Request, params httprout
 	}
 	if err := h.storage.Write(cache.ID, start, r.Body); err != nil {
 		h.responseJSON(w, r, 500, err)
+		return
 	}
 	h.useCache(id)
 	h.responseJSON(w, r, 200)
--- a/pkg/artifactcache/handler_test.go
+++ b/pkg/artifactcache/handler_test.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"os"
 	"path/filepath"
 	"strings"
 	"testing"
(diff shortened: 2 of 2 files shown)


baiyuxi930826:
Friendly ping for review when you have a moment.

This is a small fail-closed fix for #6131: return after `storage.Write` failure on artifact cache upload so we don't double-write a 200 after a 500. Includes a regression test that forces Write failure (storage root is a file) and asserts a single 500 JSON body.

Local: `go test ./pkg/artifactcache/ -run TestHandler` passes on this branch.

louzt:
Verified locally against `act` master HEAD 4f41128 — the missing `return` after `responseJSON(w, r, 500, err)` correctly prevents the second `WriteHeader(200)` call. The new regression test `upload write failure returns 500 once` (bad rootDir path) is a clean way to assert the single-response contract. Happy to retest if you rebase or want a second reviewer.

PR #6136, opened 2026-07-25
Fix panic in format() on negative replacement index

semx:
## Description

`format()` panics with `runtime error: index out of range [-1]` when a placeholder uses a negative index, e.g. `format('{-1}', 'x')`.

When parsing a `{N}` placeholder, the `-` and digits are collected into `replacementIndex`, so `strconv.ParseInt("-1", 10, 32)` succeeds and returns `-1`. The only guards were a parse error and an index that is too large (`len(replaceValue) <= int(index)`), which is `false` for a negative index. Execution then reaches `replaceValue[index]` with a negative index and the runtime panics.

GitHub Actions instead returns a handled expression error for such input, the same way `act` already handles the out-of-range case `{2147483648}`.

## Reproduction

```go
NewInterpeter(&EvaluationEnvironment{}, Config{}).Evaluate("format('{-1}', 'x')", DefaultStatusCheckNone)
```

Before: `panic: runtime error: index out of range [-1]`
After: returns error `The following format string is invalid: '{-1}'`

## Fix

Reject a negative index alongside the existing parse-error guard. A test case `format('{-1}', 'x')` is added next to the existing `{2147483648}` case.

### DIFF
--- a/pkg/exprparser/functions.go
+++ b/pkg/exprparser/functions.go
@@ -93,7 +93,7 @@ func (impl *interperterImpl) format(str reflect.Value, replaceValue ...reflect.V
 
 			case '}':
 				index, err := strconv.ParseInt(replacementIndex, 10, 32)
-				if err != nil {
+				if err != nil || index < 0 {
 					return "", fmt.Errorf("The following format string is invalid: '%s'", input)
 				}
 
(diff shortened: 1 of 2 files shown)


louzt:
Verified locally against `act` master HEAD 4f41128 — the `index < 0` bounds check correctly handles `format('{-1}', 'x')`, returning the same controlled error as the existing `{2147483648}` overflow case. The new `format-invalid-negative-index` test case is a clean regression test. Happy to retest if you rebase or want a second reviewer.

PR #6139, opened 2026-07-28
fix(artifactcache): return after storage.Write error on upload (rebase of #6132)

louzt:
## Summary
Fix silent data loss in artifact upload: when `storage.Write` fails mid-upload, the handler currently continues processing instead of returning the error.

## What
- Return early after `storage.Write` error in artifact upload path
- Test verifies error propagation through handler response

## Rebase status
Rebased against current master (4f41128). Zero conflicts. Tests pass.


louzt:
Closing this PR — I verified the missing-`return` fix locally against `act` master HEAD 4f41128; the head commit is the same SHA as the still-active #6132 from @baiyuxi930826, so no added value in opening a duplicate. Thanks to the original author for the fix.

PR #6140, opened 2026-07-28
fix(exprparser): panic in format() on negative replacement index (rebase of #6136)

louzt:
## Summary
Fix panic in `format()` function when replacement index is negative (out of bounds access on underlying writer).

## What
- Bounds-check replacement index before write
- Test covers negative index case

## Rebase status
Rebased against current master (4f41128). Zero conflicts. Tests pass.


louzt:
Closing this PR — I verified the `index < 0` bounds check locally against `act` master HEAD 4f41128; the head commit is the same SHA as the still-active #6136 from @semx, so no added value in opening a duplicate. Thanks to the original author for the fix.

PR #6148, opened 2026-08-04
feat: run steps in parallel with background, wait, wait-all and cancel

McNultyyy:
Implements the concurrent step syntax GitHub [shipped on June 25, 2026](https://github.blog/changelog/2026-06-25-actions-steps-can-now-be-run-in-parallel/): `jobs.<job_id>.steps[*].background`, `wait`, `wait-all`, `cancel` and `parallel` ([reference](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idstepsparallel)). act's schema currently rejects these keys as unknown properties, so workflows using them cannot run locally at all.

## What this does

* **`background: true`** starts the step asynchronously and continues with the next step. At most 10 background steps run concurrently (matching GitHub's documented limit), additional ones queue until a slot frees up.
* **`wait:`** (single id or array) pauses the job until the referenced background steps complete; **`wait-all:`** waits for all active background steps. A failed background step fails the job at the first `wait`/`wait-all` that includes it — later steps before the wait still run — unless `continue-on-error` is set. Outputs of background steps become available after the wait. Both step kinds always run (even after a previous failure) and do not support `if`, per the documentation.
* **`cancel:`** gracefully terminates a running background step (context cancellation kills the process/exec); a deliberately cancelled step does not fail the job. Background steps still running when the job ends (e.g. services) are stopped the same way without affecting the job result.
* **`parallel:`** desugars to background steps plus an implicit `wait` for the group — exactly the shorthand described in the changelog. Not supported inside composite actions, matching GitHub.

Schema, model (`Step` type, parsing, new `StepType`s) and runner (a per-job background step registry) are covered.

## Thread safety

Steps now genuinely run concurrently inside one job, so the shared job state handling was hardened:

* Shared state mutations (StepResults, env files, extra path, intra-action state) are serialized: each step holds a lock during its setup and result-processing phase, while the actual command runs unlocked.
* The runner file commands (`GITHUB_OUTPUT`, `GITHUB_ENV`, ...) get a unique directory per step execution so concurrent steps no longer share the same files.
* Outputs and saved state are routed by explicit step id instead of the mutable `CurrentStep` field.
* The log writer swap in `HostEnvironment` and the docker container reference now happens under a mutex, and mask lists are guarded (they are appended by running steps and read on every formatted log line). This also fixes a pre-existing `ptyWriter.AutoStop` race.

The new integration tests pass with `-race` on Linux.

Known limitation: log lines of concurrently running steps can occasionally be attributed to the wrong step label in terminal output (GitHub renders separate log sections per background step; act interleaves). Legacy `::set-output::` workflow commands from concurrently running steps are routed by the step that owns the output stream.

## Testing

* `TestReadWorkflow_BackgroundSteps` — parsing and schema validation of all five keys.
* `TestExpandParallelStepGroups` — the `parallel` desugaring (ids, implicit wait).
* `TestRunBackgroundSteps` (host environment, bash + PowerShell testdata variants) — proves via file timestamps that a background step overlaps the following step, that parallel group members overlap each other and complete before the next step, that outputs are visible after `wait`, and that `cancel` stops a 60s step immediately.
* `TestRunBackgroundStepFailure` — failure surfaces at the `wait` (not before), and `continue-on-error` suppresses it.
* A container-based workflow added to the `TestRunEvent` table; also verified manually against Docker (background exec, cancel and wait-all all behave inside job containers).

`go build ./...`, `go vet`, and `golangci-lint run` (v2.11.4) are clean on the touched packages, and the new tests were run with the race detector enabled.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/pkg/container/docker_run.go
+++ b/pkg/container/docker_run.go
@@ -15,6 +15,7 @@ import (
 	"runtime"
 	"strconv"
 	"strings"
+	"sync"
 
 	"net/netip"
 
@@ -197,6 +198,9 @@ func (cr *containerReference) GetHealth(ctx context.Context) Health {
 }
 
 func (cr *containerReference) ReplaceLogWriter(stdout io.Writer, stderr io.Writer) (io.Writer, io.Writer) {
+	cr.logWriterMutex.Lock()
+	defer cr.logWriterMutex.Unlock()
+
 	out := cr.input.Stdout
 	err := cr.input.Stderr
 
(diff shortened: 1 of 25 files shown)


McNultyyy:
Folded into #6147, which now contains both the concurrency groups and the parallel steps implementation.

PR #6156, opened 2026-08-08
fix: serialize artifact cache upload and commit for the same entry

### LINKED ISSUE #6012: Race Condition in Artifact Cache Upload/Commit Operations
## Bug: Race Condition in Artifact Cache Upload/Commit Operations

**Summary:**  
A race condition exists between the `upload` and `commit` handlers in `pkg/artifactcache/handler.go` that can lead to data corruption or inconsistent cache state when multiple concurrent operations target the same cache entry.

**Affected Code:**  
`pkg/artifactcache/handler.go`, lines 259-282 (upload handler)

**Description:**

The `upload` handler performs a time-of-check to time-of-use (TOCTOU) vulnerability:

1. Opens DB, checks if `cache.Complete` is false
2. **Closes DB** (line 269)
3. Writes data to storage via `h.storage.Write()`

Meanwhile, the `commit` handler can:
1. Mark the same cache as `Complete = true`
2. Finalize the cache entry

**Race Condition Timeline:**

```
Thread A (upload)         |  Thread B (commit)
--------------------------|---------------------------
Open DB                   |
Check cache.Complete=false|
Close DB                  |
                          | Open DB
                          | Set cache.Complete=true
                          | storage.Commit(cache.ID)
                          | Close DB
storage.Write(cache.ID)   | ← Writes to "completed" cache!
```

**Impact:**

1. **Data Corruption**: Partial writes may be committed to a cache marked as complete
2. **Invariant Violation**: The system assumes completed caches are immutable, but concurrent uploads can modify them
3. **Inconsistent State**: Cache metadata shows "complete" while storage files are still being written

**Reproduction:**

The bug is most likely to manifest when:
- Multiple workflow steps upload chunks to the same cache simultaneously
- Network delays cause commit to execute during an in-progress upload
- High concurrency environments with many parallel actions

**Suggested Fix:**

Hold the DB connection open (or use proper locking) during the entire upload operation:

```go
// PATCH /_apis/artifactcache/caches/:id
func (h *Handler) upload(w http.ResponseWriter, r *http.Request, params httprouter.Params) {
    id, err := strconv.ParseUint(params.ByName("id"), 10, 64)
    if err != nil {
        h.responseJSON(w, r, 400, err)
        return
    }

    db, err := h.openDB()
    if err != nil {
        h.responseJSON(w, r, 500, err)
        return
    }
    defer db.Close()  // ← Keep open throughout

    cache := &Cache{}
    if err := db.Get(id, cache); err != nil {
        if errors.Is(err, bolthold.ErrNotFound) {
            h.responseJSON(w, r, 400, fmt.Errorf("cache %d: not reserved", id))
            return
        }
        h.responseJSON(w, r, 500, err)
        return
    }

    if cache.Complete {
        h.responseJSON(w, r, 400, fmt.Errorf("cache %v %q: already complete", cache.ID, cache.Key))
        return
    }

    // Re-check after write to catch concurrent commits
    start, _, err := parseContentRange(r.Header.Get("Content-Range"))
    if err != nil {
        h.responseJSON(w, r, 400, err)
        return
    }
    
    if err := h.storage.Write(cache.ID, start, r.Body); err != nil {
        h.responseJSON(w, r, 500, err)
        return
    }

    // Verify cache is still incomplete before finalizing
    if err := db.Get(id, cache); err != nil {
        h.responseJSON(w, r, 500, err)
        return
    }
    if cache.Complete {
        h.responseJSON(w, r, 409, fmt.Errorf("cache %v %q: was committed during upload", cache.ID, cache.Key))
        return
    }

    cache.UsedAt = time.Now().Unix()
    if err := db.Update(cache.ID, cache); err != nil {
        h.responseJSON(w, r, 500, err)
        return
    }

    h.responseJSON(w, r, 200)
}
```

Alternatively, introduce a mutex per cache ID or use BoltDB's transaction isolation more effectively.

**Environment:**
- Analyzed in latest main branch (commit as of 2026-02-26)
- Applies to all versions using the artifact cache handler

**Additional Notes:**

This is a classic TOCTOU race condition. While the window may be small, in high-throughput CI/CD environments with concurrent workflow executions, it can lead to subtle corruption that's difficult to debug.

The same pattern should be audited in the `commit` handler to ensure it doesn't have similar issues with concurrent operations.

hamodywe:
upload and commit both close their database handle before touching storage, so nothing prevented a commit from finalizing a cache -- and, on cleanup, deleting its temporary chunk directory -- while a concurrent upload for the same cache ID was still writing into that directory. Depending on timing this drops or corrupts the uploaded bytes, or causes the in-flight upload request to fail outright.

Add a per-cache-ID mutex (cacheLock) and hold it for the full check-then-write sequence in both handlers, so upload and commit for the same cache ID are now mutually exclusive while unrelated cache IDs remain fully concurrent.

TestHandler_UploadCommitRace uses a small test-only hook to force the exact interleaving deterministically rather than relying on winning a nanosecond-wide race by chance: it pauses upload right after its completeness check, starts a concurrent commit, confirms commit cannot finish while upload holds the lock, then verifies the committed artifact matches exactly what was uploaded. Reverting just the lock (keeping the hook) makes this test fail every time with "commit finished while upload still held the cache lock", often alongside a connection reset on the upload request itself -- confirming both that the race is real and that the test catches it.

fixes #6012

### DIFF
--- a/pkg/artifactcache/cachelock.go
+++ b/pkg/artifactcache/cachelock.go
@@ -1,0 +1,53 @@
+package artifactcache
+
+import "sync"
+
+// cacheLock hands out one mutex per cache ID, so callers operating on
+// different cache entries never block each other while callers operating on
+// the same entry are fully serialized.
+//
+// This exists to close the race between the upload and commit handlers
+// described in https://github.com/nektos/act/issues/6012: both close their
+// database handle before touching storage, so nothing previously prevented
+// a commit from finalizing (and, on cleanup, deleting) a cache's temporary
+// chunk directory while a concurrent upload was still writing into it.
+type cacheLock struct {
+	mu    sync.Mutex
+	locks map[uint64]*refCountedMutex
+}
+
+type refCountedMutex struct {
+	mu  sync.Mutex
+	ref int
+}
+
+func newCacheLock() *cacheLock {
+	return &cacheLock{locks: make(map[uint64]*refCountedMutex)}
+}
+
+// Lock blocks until the lock for id is held and returns a function that
+// releases it. The caller must invoke the returned function exactly once,
+// typically via defer, to avoid leaking the entry.
+func (c *cacheLock) Lock(id uint64) func() {
+	c.mu.Lock()
+	l, ok := c.locks[id]
+	if !ok {
+		l = &refCountedMutex{}
+		c.locks[id] = l
+	}
+	l.ref++
+	c.mu.Unlock()
+
+	l.mu.Lock()
+
+	return func() {
+		l.mu.Unlock()
+
+		c.mu.Lock()
+		l.ref--
+		if l.ref == 0 {
+			delete(c.locks, id)
+		}
+		c.mu.Unlock()
+	}
+}
--- a/pkg/artifactcache/cachelock_test.go
+++ b/pkg/artifactcache/cachelock_test.go
@@ -1,0 +1,107 @@
+package artifactcache
+
+import (
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestCacheLock_SameIDIsSerialized(t *testing.T) {
+	c := newCacheLock()
+
+	unlock := c.Lock(1)
+
+	acquired := make(chan struct{})
+	go func() {
+		unlock := c.Lock(1)
+		defer unlock()
+		close(acquired)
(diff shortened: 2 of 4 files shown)


hamodywe:
﻿Closing this to keep my open-PR list honest - no hard feelings, I know maintainer time is the scarce thing. The branch stays on my fork, so if this is ever useful just say the word and I'll reopen it.

PR #6174, opened 2026-08-30
fix(artifacts): support upload-artifact@v7 and download-artifact@v8

### LINKED ISSUE #6114: Artifact server rejects upload-artifact@v7 / download-artifact@v8 (mime_type + sig padding)

### LINKED ISSUE #6173: upload-artifact@v7 direct uploads do not round-trip with download-artifact@v8
### Bug report info

```plain text
act version: 0.2.89
```

### Command used with act

```sh
act push --artifact-server-path ./artifacts
```

### Describe issue

- `actions/upload-artifact@v7` supports uploading one file directly with `archive: false`.
- This is separate from the basic v7/v8 request compatibility tracked in #6114. Accepting the new `mime_type` field is not enough because act currently discards its value.
- `actions/download-artifact@v8` uses the blob response `Content-Type` to decide whether to extract it and `Content-Disposition` to recover the filename.
- Without that metadata, the direct download uses the fallback filename `artifact` instead of the uploaded filename.
- Direct artifacts should retain their MIME type and download with their original filename.

### Workflow content

```yaml
on:
  push:
jobs:
  artifacts:
    runs-on: ubuntu-latest
    steps:
      - run: echo Direct artifact > direct.txt

      - uses: actions/upload-artifact@v7
        with:
          path: direct.txt
          archive: false

      - uses: actions/download-artifact@v8
        with:
          name: direct.txt
          path: out

      - run: cmp direct.txt out/direct.txt
```

### Relevant log output

```sh
Downloading raw file (non-zip) to: .../out/artifact
cmp: out/direct.txt: No such file or directory
```

### Additional information

- With `archive: false`, upload-artifact uses the source basename as the artifact name.
- The artifact server needs to preserve the request MIME type and return appropriate `Content-Type` and `Content-Disposition` headers for the download.

chuckatkins:
Closes #6114
Closes #6173

- Accept optional fields sent by current artifact clients.
- Accept padded and unpadded artifact URL signatures.
- Preserve MIME type and filename metadata for direct uploads.
- Test real action pairs:

  - `upload-artifact@v4` -> `download-artifact@v4`
  - `upload-artifact@v4` -> `download-artifact@v5`
  - `upload-artifact@v5` -> `download-artifact@v6`
  - `upload-artifact@v6` -> `download-artifact@v7`
  - `upload-artifact@v7` -> `download-artifact@v8`

- Test a direct `archive: false` v7-to-v8 round trip.

Related work: #6039, #6099, and #6115 cover some or all of the request compatibility fixes. This PR adds the multiversion integration coverage and direct-artifact support.

Tested with `make format`, `make build`, package lint, and `go test ./pkg/artifacts -count=1`.


aadrian:
Any news on merging this?

PR #6179, opened 2026-09-04
feat(runner): inject an optional docker-in-docker service container

### LINKED ISSUE #2139: Automatically pre-start a Docker service inside the workflow containers
### Act version

act version 0.2.56

### Feature description

I am looking for an easy way (preferably without modifying my `workflow.yaml`) to automatically start a Docker daemon **inside** the running workflow container (i.e., within what is spawned by _`act`_). I am using *`-P ubuntu-22.04=catthehacker/ubuntu:act-22.04`*, both _`act`_ and the underlying image freshly pulled:

```
❯ docker images | grep catthehacker | choose 2 3..6
4b481b526128 5 days ago
```

The problem is that letting _`act`_ mount the host Docker socket and thus running a Docker container configured by the workflow on the host's Docker environment results in that container's side effect being put to a wrong location.

I have the following workflow, which I heavily simplified for the sake of exposition:

```yaml
steps:
  - uses: actions/checkout@v3  # Main repository "whisperity/Project".
  - uses: actions/checkout@v3
    with:
      repository: "organisation/Project-Resources"
      path: "Resources"

  - name: "Compile resources"
    # I do not have write rights to change what this Action does!
    uses: <a custom action which executes the 'run' script inside a pre-baked Docker environment>
    with:
      run: |
        apt-get install -yy --no-install-recommends <a bunch of dependencies...>
        cp -rv ./Resources /tmp/resource-project
        cd /tmp/resource-project
        make -j
        cp -rv /tmp/resource-project/Build ./Resources/

  - name: "Compile main project"
    run: |
      compiler RESOURCE_PATH=$(pwd)/Resources -j project
```

Here, I have a Docker container executed on a separately cloned repository which, amongst some other actions, creates some resources in a directory that **must** be available for later steps to work properly.

**This works without any problems on the "real" GitHub Actions SaaS runners.**

However, with _`act`_, it does not. Assume that I have my local repository in `<host>//home/whisperity/proj`. If I do `act` here, the Docker container created by _`act`_ will have its own `<docker-1>//home/whisperity/proj`, where the side project is also cloned: `<docker-1>//home/whisperity/proj/Resources`.

However, when the reusable inner action spawns another Docker container, `<docker-2>`, it will do so through the **host**'s socket. In there, `<docker-2>//home/whisperity/proj` (the working directory) is the same as `<host>//home/whisperity/proj`, but this inner action fails, as the checkout of the side repository did **not** happen to this location, failing the entire workflow. The _"Resources"_ directory is only present in `<docker-1>//`.

Specifying `--container-daemon-socket "-"` will just result in the _`docker`_ binary complaining there is no Docker service running within the container. Rightfully so.

My request is to have a way to automatically start a Docker daemon _inside the containers_ that are running through _`act`_. Preferably, without having to touch the workflow file! This whole idea might be related to services (#173), but having control over a Docker daemon not attached to the main host is sort of like a "meta service". I see two ways forward for this:

 - A dedicated option, and feature support, within _`act`_ and the client images, which will spawn Docker if the user requests it when running `act`. Or something like `--container-daemon-socket "spawn"`?
 - The option to specify a script that is executed every time a job starts running in a container (and vice versa, after all workflow actions had run?). Here, the user could manually script in any arbitrary action, not just the backgrounded starting of a Docker daemon.

vicamo:
## Summary

Adds an opt-in Docker-in-Docker (dind) service that runs alongside every job's container when the workflow does not declare its own `docker` service, so steps can run `docker build`, `docker compose`, testcontainers, etc. **without bind-mounting the host's Docker socket** into the job container.

Off by default — existing runs are byte-identical.

Closes #2139

## Usage

```sh
# simplest: just turn it on
act --docker-daemon-service

# override the dind image
act --docker-daemon-service --docker-daemon-service-image=docker:27-dind
```

Inside the job container:

```sh
ls -l /var/run/docker.sock          # exists, symlinked to the shared socket
docker version                      # talks to the injected dockerd
docker run --rm -v "$GITHUB_WORKSPACE:/w" alpine cat /w/marker
```

No `DOCKER_HOST` needed. Workflows that hardcode `/var/run/docker.sock` keep working.

## Why not just mount `/var/run/docker.sock` from the host?

That's what `act` does today by default, and it's what #2139 explicitly wants to avoid: any workflow step can then root the host, and paths handed to dockerd resolve on the host filesystem, not the workflow's. This change gives users a per-run daemon whose lifecycle is bounded by the job and whose filesystem view matches the job container.

## Design

Two commits:

### 1. `feat(runner): inject optional docker-in-docker service container`

- Two new CLI flags plumbed through `cmd.Input` into `runner.Config`:
  - `--docker-daemon-service` (bool, default off)
  - `--docker-daemon-service-image` (string, default `docker:dind`)
- **Socket-sharing, not TCP.** dind runs with `DOCKER_TLS_CERTDIR=""` and `dockerd --host=unix:///var/run/act-dind/docker.sock`. A per-job named volume `<jobcontainer>-dind-sock` is mounted at `/var/run/act-dind` in both containers. Mirrors GitHub-hosted runners, which expose a plain unix socket with no `DOCKER_HOST`. No forced bridge network (`--network host` is preserved), no TLS-cert bootstrap.
- **`/var/run/docker.sock` symlink.** After dind is up, a post-start step polls for the socket (bounded, exponential backoff, 2 min cap) and drops `/var/run/docker.sock -> /var/run/act-dind/docker.sock` inside the job container. GitHub-identical path; nothing else needs `DOCKER_HOST`.
- **Injection stays out of the workflow model.** The dind container is only appended to `rc.ServiceContainers` at runtime, never to `rc.Run.Job().Services`. `${{ job.services }}` therefore reflects only what the user wrote in YAML, preserving compatibility with expressions, reusable-workflow outputs, and downstream tooling. `useDockerDaemonService()` returns false when the workflow already declares a service literally named `docker`, so a user-defined service always wins.
- **Lifecycle reuses existing plumbing.** dind flows through `pullServicesImages` / `startServiceContainers` / `waitForServiceContainers` / `stopServiceContainers` unchanged. Cleanup removes the per-job socket volume too (skipped under `--reuse`).
- **Mutually exclusive with `--container-daemon-socket`**, extracted into `validateDockerDaemonServiceFlags` so it's unit-testable:
  - not explicitly set → auto-forced to `-` so the default host-socket bind is suppressed;
  - explicitly `-` → accepted;
  - anything else → hard error with the offending value quoted.

### 2. `feat(runner): mirror job binds/mounts onto the dind service container`

GitHub-hosted runners run dockerd on the same filesystem as the job. `docker run -v $GITHUB_WORKSPACE:/x`, testcontainers volume mounts, `docker cp` between job and helper containers, and actions that bind `/opt/hostedtoolcache` into sidecars all resolve identically on both sides. Passes the job container's binds/mounts through to the dind container:

- workdir bind (when `BindWorkdir` on) or anonymous workdir volume (when off) attached to dind at the same in-container path;
- named volumes (`act-toolcache`, `<name>-env`, user-declared `job.container.volumes`) shared with dind;
- the shared socket volume is already in the mount map.

Also fixes a latent cleanup-ordering bug this makes deterministic: we were removing named volumes before stopping service containers, so any service holding a workdir/toolcache volume (dind, or a user-declared service that mounts the workdir) failed with `volume is in use`. Now the pipeline stops the job container, stops all service containers and the ad-hoc network, and *then* removes the volumes.

## Compatibility

- Flag off (default): no code path changes, no new containers, no new volumes. Full test suite unchanged.
- Flag on + workflow declares `services.docker`: user's service wins, no injection.
- Flag on + no user `docker` service: dind is injected, `job.services` context is untouched.
- Embedders (Gitea/Forgejo `act_runner`) that consume `runner.Config` only see two new opt-in fields — additive, no semver break.

## Tests

Unit:
- `cmd.TestValidateDockerDaemonServiceFlags` — 9 subtests covering the full matrix of `dockerDaemonService × containerDaemonSocket × flag.Changed`, asserting both the error (with substring match on the quoted offending value) and the resulting socket value.

Integration (`pkg/runner.TestRunEventDockerDaemonService`, skipped under `-short`, needs a real Docker daemon that permits `--privileged`):

- `testdata/docker-daemon-service/push.yml` — verifies `/var/run/docker.sock` exists at the GitHub-hosted path, `docker version` reaches the injected daemon, and `docker run alpine` succeeds through it.
- `testdata/docker-daemon-service-shared-fs/push.yml` — writes a marker to `$GITHUB_WORKSPACE`, then starts an `alpine` container via the injected dockerd with `-v $GITHUB_WORKSPACE:/w` and reads the marker back, proving the shared-filesystem invariant.

Verified on Docker 29.7.2:

```
--- PASS: TestRunEventDockerDaemonService (19.49s)
    --- PASS: TestRunEventDockerDaemonService/docker-daemon-service (8.88s)
    --- PASS: TestRunEventDockerDaemonService/docker-daemon-service-shared-fs (10.61s)
ok  github.com/nektos/act/cmd         22.552s
ok  github.com/nektos/act/pkg/runner   5.835s
```

## Out of scope

- Rootless / Podman: dind requires `--privileged` on the outer daemon. Rootless setups should keep using `--container-daemon-socket=-` and run their own dind sidecar out-of-band.
- Per-job dind image pinning via workflow syntax — image is a CLI/config concern, not a workflow concern.

## Disclaimer

This change was developed with AI assistance (pair-programming with a large language model). I drove the design decisions, reviewed every diff, and ran the full unit and integration test suites locally on Docker 29.7.2 before pushing. Any bugs are mine, not the model's.

### DIFF
--- a/cmd/input.go
+++ b/cmd/input.go
@@ -35,6 +35,8 @@ type Input struct {
 	containerArchitecture              string
 	containerDaemonSocket              string
 	containerOptions                   string
+	dockerDaemonService                bool
+	dockerDaemonServiceImage           string
 	noWorkflowRecurse                  bool
 	useGitIgnore                       bool
 	githubInstance                     string
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -113,6 +113,8 @@ func createRootCommand(ctx context.Context, input *Input, version string) *cobra
 	rootCmd.PersistentFlags().StringVarP(&input.containerArchitecture, "container-architecture", "", "", "Architecture which should be used to run containers, e.g.: linux/amd64. If not specified, will use host default architecture. Requires Docker server API Version 1.41+. Ignored on earlier Docker server platforms.")
 	rootCmd.PersistentFlags().StringVarP(&input.containerDaemonSocket, "container-daemon-socket", "", "", "URI to Docker Engine socket (e.g.: unix://~/.docker/run/docker.sock or - to disable bind mounting the socket)")
 	rootCmd.PersistentFlags().StringVarP(&input.containerOptions, "container-options", "", "", "Custom docker container options for the job container without an options property in the job definition")
+	rootCmd.PersistentFlags().BoolVarP(&input.dockerDaemonService, "docker-daemon-service", "", false, "Start a Docker-in-Docker service container (network alias 'docker') alongside the job container so builds have a Docker daemon without mounting the host socket. Sets DOCKER_HOST/DOCKER_TLS_VERIFY/DOCKER_CERT_PATH in the job container and disables the host socket bind.")
+	rootCmd.PersistentFlags().StringVarP(&input.dockerDaemonServiceImage, "docker-daemon-service-image", "", "docker:dind", "Image to use for the docker-in-docker service container.")
 	rootCmd.PersistentFlags().StringVarP(&input.githubInstance, "github-instance", "", "github.com", "GitHub instance to use. Only use this when using GitHub Enterprise Server.")
 	rootCmd.PersistentFlags().StringVarP(&input.artifactServerPath, "artifact-server-path", "", "", "Defines the path where the artifact server stores uploads and retrieves downloads from. If not specified the artifact server will not start.")
 	rootCmd.PersistentFlags().StringVarP(&input.artifactServerAddr, "artifact-server-addr", "", common.GetOutboundIP().String(), "Defines the address to which the artifact server binds.")
@@ -406,6 +408,10 @@ func newRunCommand(ctx context.Context, input *Input) func(*cobra.Command, []str
 			return listOptions(cmd)
 		}
 
+		if err := validateDockerDaemonServiceFlags(input, cmd.Flag("container-daemon-socket").Changed); err != nil {
+			return err
+		}
+
 		if ret, err := container.GetSocketAndHost(input.containerDaemonSocket); err != nil {
 			log.Warnf("Couldn't get a valid docker connection: %+v", err)
 		} else {
@@ -630,6 +636,8 @@ func newRunCommand(ctx context.Context, input *Input) func(*cobra.Command, []str
 			ContainerArchitecture:              input.containerArchitecture,
 			ContainerDaemonSocket:              input.containerDaemonSocket,
 			ContainerOptions:                   input.containerOptions,
+			DockerDaemonService:                input.dockerDaemonService,
+			DockerDaemonServiceImage:           input.dockerDaemonServiceImage,
 			UseGitIgnore:                       input.useGitIgnore,
 			GitHubInstance:                     input.githubInstance,
 			ContainerCapAdd:                    input.containerCapAdd,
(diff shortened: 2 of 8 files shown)


vicamo:
Test container volumes as well.

vicamo:
Add `Co-Authored-By: Claude <noreply@anthropic.com>` and use work email as author email.

